PR #191, opened 2023-02-05
typo fixes

dnoenickx:


packages/2d/src/components/Layout.ts
```diff
@@ -292,7 +292,7 @@ export class Layout extends Node {
    * Represents the size of this node.
    *
    * @remarks
-   * A size is a two-dimensional vector, where `x` represents the width, and `y`
+   * A size is a two-dimensional vector, where `x` represents the `width`, and `y`
    * represents the `height`.
```
dnoenickx, line 296: not a typo, but making it consistent for width and height

packages/core/src/flow/chain.ts
```diff
@@ -25,7 +25,7 @@ decorate(chain, threadable());
  * functions. For example:
  * ```ts
  * yield* all(
- *   rect.corenerRadius(20, 3),
+ *   rect.cornerRadius(20, 3),
```
aarthificial, line 28: This is actually outdated. It should be simply:
```suggestion
 *   rect.radius(20, 3),
```

aarthificial (changes requested):

aarthificial:
The checks have failed but you don't need to worry about this. In the future be sure to name your commits like this:
```
docs: typo fixes
```
In order to meet the [Angular Commit Message Conventions](https://github.com/angular/angular/blob/main/CONTRIBUTING.md#commit).

PR #194, opened 2023-02-05
docs: add custom font

CIOSAI:
Thought this might be useful.

### DIFF
--- a/packages/docs/docs/advanced/custom-font.mdx
+++ b/packages/docs/docs/advanced/custom-font.mdx
@@ -1,0 +1,1 @@
+hello bruh
--- a/packages/docs/docs/advanced/custom-font.mdx
+++ b/packages/docs/docs/advanced/custom-font.mdx
@@ -2,14 +2,14 @@
 
 To use a font from hosters like Google Fonts. First make a css file under `src`
 
-```
+```diff
 root
 └─src
   ├─scenes/
   ├─motion-canvas.d.ts
   ├─project.meta
   ├─project.ts
-  └─global.css *
++ └─global.css
 ```
 
 Inside `global.css`, import the font using `@import url(your link)`.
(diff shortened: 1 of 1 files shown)


packages/docs/docs/advanced/custom-font.mdx
```diff
@@ -0,0 +1,71 @@
...
+
+```
+root
+└─src
+  ├─scenes/
+  ├─motion-canvas.d.ts
+  ├─project.meta
+  ├─project.ts
+  └─global.css *
+```
```
aarthificial, line 13: You can the `diff` language to better highlight the changes:
````suggestion
```diff
root
└─src
  ├─scenes/
  ├─motion-canvas.d.ts
  ├─project.meta
  ├─project.ts
+ └─global.css
```
````
image

File: packages/docs/docs/advanced/custom-font.mdx [hunk 2]
```diff
@@ -0,0 +1,71 @@
...
+
+Then, in `project.ts`, import the css file.
+
+```ts
+import {makeProject} from '@motion-canvas/core/lib';
+
+import example from './scenes/example?scene';
+
+import "./global.css" <- import the css
```
aarthificial, line 28: This should be a comment to make this a valid snippet, in case people will copy it.
```suggestion
import "./global.css"; // <- import the css
```

packages/docs/docs/advanced/custom-font.mdx
```diff
@@ -0,0 +1,71 @@
+# Custom font
+
+To use a font from hosters like Google Fonts. First make a css file under `src`
```
aarthificial, line 3: I'd recommend spliting this into two sections: "Loading fonts from the web" and "Loading fonts locally" by adding the corresponding markdown headers:
```suggestion
## Loading fonts from the web

To use a font from hosters like Google Fonts. First make a css file under `src`
```

aarthificial (changes requested):

CIOSAI:
thanks for tell me about diff

Issue #179, opened 2023-02-04
Text is invisible in Firefox


TomatechGames:
**Describe the bug**
By default, Firefox has `dom.textMetrics.fontBoundingBox` disabled, meaning that Text components dont render at all (this includes the HTML embeds of Motion Canvas animations)

**To Reproduce**
In a fresh install of Firefox with default settings, try to view a scene that uses a Text component

**Expected behavior**
The text is displayed

**Additional context**
A workaround for users is to manually enable `dom.textMetrics.fontBoundingBox` within Firefox's config by typing `about:config` in the address bar

LegoGuy32109:
Would love to have this mentioned in the docs somewhere when following quickstart or the signals paged https://motion-canvas.github.io/docs/signals

n3oney:
I've found another text related bug that only happens on FF, so not sure if I should open a new issue or mention it here. Text wrapping doesn't work, it always renders on one line
![image](https://user-images.githubusercontent.com/30625554/217674166-4e4abda4-a2e4-4ffa-b440-01cb9578c041.png)
*chrome on the left, firefox on the right*

(no matter if it's preformatted or auto-wrapping)

exdeejay:
@n3oney can you make a separate issue with your reply?

n3oney:
Sure.

PR #195, opened 2023-02-05
[Documentation] Info panel for Firefox Users

Tandashi:
Since there currently is some confusion for Firefox users, why the `Text` component is not rendered / visible an Info Panel could help aid with confusion.

I put them everywhere where the `Text` component has been used in the docs, since it made the most sense to me.

Related Issue: #179

### DIFF
--- a/packages/docs/docs/getting-started/hierarchy.mdx
+++ b/packages/docs/docs/getting-started/hierarchy.mdx
@@ -6,6 +6,7 @@ slug: /hierarchy
 import Mermaid from '@theme/Mermaid';
 import CodeBlock from '@theme/CodeBlock';
 import ApiSnippet from '@site/src/components/Api/ApiSnippet';
+import FirefoxTextWarning from '@site/src/warnings/FirefoxTextWarning';
 
 # Scene hierarchy
 
@@ -15,6 +16,8 @@ the Document Object Model used ot represent HTML and XML documents.
 
 Here's an example of a simple scene hierarchy together with its object
 representation:
+
+<FirefoxTextWarning />
 
 <div className="row margin-bottom--md">
   <div className="col col--6">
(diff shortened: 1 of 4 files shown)


yemibox51 (approved):

packages/docs/docs/getting-started/hierarchy.mdx
```diff
@@ -16,6 +16,14 @@ the Document Object Model used ot represent HTML and XML documents.
 Here's an example of a simple scene hierarchy together with its object
 representation:
 
+:::info
+
+If you are using Firefox make sure to check if `dom.textMetrics.fontBoundingBox`
```
aarthificial, line 21: I think we should turn this into a component:
```tsx
import Admonition from '@site/src/theme/Admonition';
import Link from '@docusaurus/Link';

export default function FirefoxTextWarning() {
  return (
    <Admonition type="caution">
      If you are using Firefox make sure to check if
      `dom.textMetrics.fontBoundingBox` is enabled else you might not see the{' '}
      <Link to="/api/2d/components/Text">
        <code>Text</code>
      </Link>
      . You can enable this by navigating to <code>about:config</code>.
    </Admonition>
  );
}
```

And then use it in markdown:
```tsx
import FirefoxTextWarning from '@site/src/warnings/FirefoxTextWarning';
// ...
<FirefoxTextWarning />
```

This will make it easier to find and remove all these notices in the future, once the bug is fixed.

aarthificial (changes requested):

Minenash:
I would probably also add some type of notice for any examples that use it. For example, the Signals page

packages/docs/docs/getting-started/hierarchy.mdx
Tandashi: Yeah, totally agree. Done it this way now.

Issue #196, opened 2023-02-05
npm init @motion-canvas doesn't work


FilledStacks:
**Description**
When I run `npm init @motion-canvas` it prints out the following. It's confusing for developers that are not web developers and are unfamiliar with the npm system and how to use it with packages that are not on npm. 

```shell
npm ERR! code E404
npm ERR! 404 Not Found - GET https://npm.pkg.github.com/@motion-canvas%2fcreate - npm package "create" does not exist under owner "motion-canvas"
npm ERR! 404 
npm ERR! 404  '@motion-canvas/create@latest' is not in this registry.
npm ERR! 404 
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/danemackier/.npm/_logs/2023-02-05T13_14_09_505Z-debug-0.log
```

I think there needs to be clearer instructions on how to get started from scratch. If you give me clear instructions I'll make a PR to update the docs as well.

FilledStacks:
After running `npm init @motion-canvas` in the github repo it's working and creating the project. 

Now when I run `npm install` in the created project I get 

```shell
npm ERR! code E404
npm ERR! 404 Not Found - GET https://npm.pkg.github.com/@motion-canvas%2f2d - npm package "2d" does not exist under owner "motion-canvas"
npm ERR! 404 
npm ERR! 404  '@motion-canvas/2d@^2.0.0' is not in this registry.
npm ERR! 404 
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/danemackier/.npm/_logs/2023-02-05T13_23_03_387Z-debug-0.log
```

dtamai:
@FilledStacks, do you have some custom npm configuration that causes it to use Github's registry (_npm.pkg.github.com_) instead of npm's one (_registry.npmjs.org_)?

You can check that the package exists in the npm registry [here](https://registry.npmjs.org/@motion-canvas%2f2d).

FilledStacks:
@dtamai I don't think so. I use to use motion canvas back when it was patreon only, that might have done something?

I'll google how to reset npm registry and then try again. Will report in a few minutes.

20milliliter:
Unless you plan to contribute, you do not need to download the repo.
[...]

PR #200, opened 2023-02-05
docs: clarification on scaffolding process

20milliliter:
Clarifies that to use Motion Canvas, cloning the repo is not required.

Issue #196

### DIFF
--- a/packages/docs/docs/getting-started/quickstart.mdx
+++ b/packages/docs/docs/getting-started/quickstart.mdx
@@ -34,6 +34,13 @@ Run the following command in order to scaffold a new Motion Canvas project:
 npm init @motion-canvas
 ```
 
+:::tip
+
+Unless you plan to contribute, you do not need to clone the github repo. 
+As long as you have node, this command is all you need.
+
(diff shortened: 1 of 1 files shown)


yemibox51 (approved):

VinithS (changes requested):
Seems redundant. How to being using is already defined in [quickstart guide](https://motion-canvas.github.io/docs/quickstart/).

And if you are a programmer, you probably know how to get started with contributing.

aarthificial:
Agree with @VinithS
Unless there's something in the docs that I forgot about, that would suggest you need to clone the repo, @20milliliter?

aarthificial:
I've noticed that this PR originates [from this issue](https://github.com/motion-canvas/motion-canvas/issues/196#issuecomment-1417869868) and I must disagree. The issue was caused by the fact that @FilledStacks used to participate in the early access of Motion Canvas and had their npm configured to fetch mc packages from GitHub. **They did not try to clone the repo** and I don't see how anyone would be confused about that.

Issue #186, opened 2023-02-04
Docs typo


nick-1666:
Text says 320, code says 240
![image](https://user-images.githubusercontent.com/67550680/216787793-eb0cea6f-25fa-445d-9748-e71c23056693.png)

aarthificial:
Fixed in #193

PR #193, opened 2023-02-05
Fix numerical typo

damageboy:

PR #203, opened 2023-02-05
docs: fix typo in quickstart.mdx

michaelharlow:
- Fixed typo in value assigned to **width** and **height** in [quickstart.mdx](/motion-canvas/motion-canvas/blob/main/packages/docs/docs/getting-started/quickstart.mdx)
- Resolves: #186

### Previously:
image

### Fixed:
image


aarthificial:
Thanks for the fix!
I'm closing it because it seems like someone have already fixed this in #193

PR #206, opened 2023-02-05
feat(components): axes and plots

Kobemeka:
Support for drawing plots and axes.

Initial commit for plotting a function in a given range and drawing axes.


Kobemeka:
#269

levirs565:
Why this PR closed?

Issue #171, opened 2023-02-04
Use time events for duration


aarthificial:
**Description**
Time events allow for editing *when* something should happen, but they could also define *how long* it should last.

**Proposed solution**
From a technical perspective, this is already possible.
The following example will make the tween last as long as the `event` event:
```ts
useScene().timeEvents.register('event')
const duration = useScene().timeEvents.get('event').offset;
yield* circle().scale(2, duration);
```
But it would be nice to have a compact helper function for this:
```ts
yield* circle().scale(2, useDuration('event'));
```

Tandashi:
**Note for others**: Working on this.

PR #226, opened 2023-02-06
feat: add `useDuration` helper

Tandashi:
## What has changed?

This Pull Request adds a new helper / utility function that allows for easier creation and usage of `TimeEvent`s for animations.

Furthermore, the documentation has been extended and now features a subsection ("Controlling animation duration with Time Events") under the `Time Events` chapter.

Closes: #171

---

## Example
Let's look at the following example where we want to animate the `scale` of a `circle`:
```tsx
import {Circle} from '@motion-canvas/2d/lib/components';
import {createRef, useScene} from '@motion-canvas/core/lib/utils';

  view.add(
  );

  useScene().timeEvents.register('event');
  const duration = useScene().timeEvents.get('event').offset;
  yield* circle().scale(2, duration);
});
```

In the example a `TimeEvent` called `event` is registered and its duration gets used to control
the `duration` of the circle's `scale` animation.

With the new `useDuration` function, this can be reduced to a single line of code as follows:
```tsx
yield* circle().scale(2, useDuration('event'));
```

### DIFF
--- a/packages/core/src/utils/useDuration.ts
+++ b/packages/core/src/utils/useDuration.ts
@@ -1,0 +1,31 @@
+import {useScene} from './useScene';
+
+/**
+ * Register a time event and get its duration in seconds.
+ *
+ * @remarks
+ * This can be used to better specify when an animation should start
+ * as well as how long this animation should take
+ *
+ * @example
+ * ```ts
+ * export default makeScene2D(function* (view) {
+ *   const circle = createRef<Circle>();
+ *
+ *   view.add(
+ *     <Circle ref={circle} width={320} height={320} fill={'lightseagreen'} />,
+ *   );
+ *
+ *   yield* circle().scale(2, useDuration('circleGrow'));
+ * });
+ * ```
+ *
+ * @param name - The name of the event.
+ *
+ * @returns The duration of the event in seconds.
+ */
+export function useDuration(name: string): number {
+  const scene = useScene();
+  scene.timeEvents.register(name);
+  return scene.timeEvents.get(name).offset;
+}
--- a/packages/docs/docs/getting-started/time-events.mdx
+++ b/packages/docs/docs/getting-started/time-events.mdx
@@ -48,3 +48,14 @@ it. This is useful when you want to extend or shorten a pause in your voiceover
 because correcting the first time event after the pause will also fix all events
 after it. You can hold `SHIFT` when editing an event to prevent this from
 happening.
+
+## Controlling animation duration
+
+Aside from specifying _when_ something should happen, Time Events can also be used to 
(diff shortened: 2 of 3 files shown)


packages/core/src/utils/useDuration.ts
```diff
@@ -0,0 +1,29 @@
+import {useScene} from './useScene';
+
+/**
+ * Register a time event and get its duration in seconds.
+ *
+ * @example
+ * This can be used to better specify when an animation should start
```
aarthificial, line 7: This should be under `@remarks`
Examples are collapsed by default when reading the docs

packages/docs/docs/getting-started/time-events.mdx
```diff
@@ -48,3 +48,36 @@ it. This is useful when you want to extend or shorten a pause in your voiceover
...
+
+```ts
+const duration = useScene().timeEvents.get('event').offset;
+```
+
+And finally we can use this `duration` to animate the circle:
+
+```ts
+yield* circle().scale(2, duration);
+```
```
aarthificial, line 75: This is an internal implementation detail, I don't think we should include it in the docs.
I'd simply say:
````markdown
Aside from specifying _when_ something should happen, time events can be used to 
control _how long_ something should last. You can use the 
[`useDuration`](/api/core/utils#useDuration) function to retrieve the duration of an event
and use it in your animation:

```ts
yield* circle().scale(2, useDuration('event'));
```
````

File: packages/docs/docs/getting-started/time-events.mdx [hunk 3]
```diff
@@ -48,3 +48,36 @@ it. This is useful when you want to extend or shorten a pause in your voiceover
 because correcting the first time event after the pause will also fix all events
 after it. You can hold `SHIFT` when editing an event to prevent this from
 happening.
+
+## Controlling animation duration with Time Events
```
aarthificial, line 52: The entire section is about time events so I think it could be just:
```suggestion
## Controlling animation duration
```

aarthificial (changes requested):

packages/core/src/utils/useDuration.ts
Tandashi: Do you mean to replace `@example` with `@remarks` or add a `@remarks` above `@example`?

aarthificial: Add remarks above example

Tandashi: Hope it's now how you meant it

aarthificial: Yup, it's okay

aarthificial:
Thanks for the PR!

PR #230, opened 2023-02-07
Stop code highlighting from jumping

### LINKED ISSUE #229: Issue with drawing CodeBlock selection
**Describe the bug**
Incorrect parts of the code are highlighted using `CodeBlock.selection`

**To Reproduce**
```ts
import {waitUntil} from '@motion-canvas/core/lib/flow';

import {
  CodeBlock,
  edit,
  word,
} from '@motion-canvas/2d/lib/components/CodeBlock';

    const code = createRef<CodeBlock>();
    view.add(
        <CodeBlock
            ref={code}
            language={'hlsl'}
            code={() => `float y = (sin(a)) * b;`}
        />
    );
    yield* code().selection(word(0, 10, 8), 1);
    yield* code().edit(1)`float y = ${edit('(sin(a))', '(cos(a))')} * b;`;
});
```

**Expected behavior**

1. `float y = (sin(a)) * b;` gets added
2. `(sin(a))` gets highlighted
3. `(sin(a))` gets replaced with `(cos(a))`, which gets highlighted

**Actual behavior**

1. `float y = (sin(a)) * b;` gets added
2. `sin(a)) * b;` gets highlighted
3. highlighting suddenly switches to `(sin(a))` and gets replaced with `(cos(a))`, which gets highlighted

- core: 2.0.0
- ui: 2.0.0

**Additional context**
Changing the language type causes the behavior to change, but does not resolve the issue. I suspect the issue is between `CodeBlock.draw` and `CodeBlock.selectionStrength`, but don't know enough typescript to be able to understand what went wrong.

hhenrichsen:
Closes #229.

### DIFF
--- a/packages/2d/src/components/CodeBlock.ts
+++ b/packages/2d/src/components/CodeBlock.ts
@@ -312,42 +312,19 @@ export class CodeBlock extends Shape {
         context.save();
         context.fillStyle = token.color ?? '#ff6470';
 
-        if (token.code.includes('()')) {
-          const alpha = context.globalAlpha;
-          for (let i = 0; i < token.code.length; i++) {
-            const char = token.code.charAt(i);
-            if (char === '\n') {
-              y++;
-              x = 0;
-              continue;
-            }
-            context.globalAlpha = alpha * getSelectionAlpha(x, y);
-            context.fillText(char, x * w, y * lh);
-            x++;
+        const alpha = context.globalAlpha;
+        for (let i = 0; i < token.code.length; i++) {
+          const char = token.code.charAt(i);
+          if (char === '\n') {
+            y++;
+            x = 0;
+            continue;
           }
-          context.restore();
-          continue;
-        }
-
-        //*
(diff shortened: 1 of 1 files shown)


aarthificial (changes requested):
One more thing I just remembered - we should also do the same in other branches, so `delete` and `create`, otherwise font ligatures will appear/disappear during transitions. It could also be a good idea to extract the token rendering code to a separate method.

hhenrichsen:
I'm not 100% sure how to get ligatures rendering so that I can test that -- are those supported right now? Even then, I'd propose that that be filed as its own issue. That way this fix can go out quick. I'm happy to look into the other issue, too, once I can get ligatures rendering.

aarthificial:
If you're using a font that supports ligatures, say JetBrains Mono, then something like this: `=>` should be turned into an arrow.
Here's an example from the newest video:
image
https://github.com/motion-canvas/examples

Issue #239, opened 2023-02-07
Allow Video Component properties to be accessed


Tandashi:
**Description**
It would be nice to be able to access different properties of the Video Component such as `time`, `playing`, ...

As of right now it is possible with a 'hacky' way:
```ts
videoRef()['playing']()   // returns if the video is playing
```

but a more official way would be good.

Something like this would be the preferred way:
```ts
videoRef().playing() 
```

**Proposed solution**
Change the visibility of the properties. 
Currently they are marked as `protected`.

PR #240, opened 2023-02-07
feat(2d): add video component property getter

Tandashi:
## What has been changed in this Pull Request

Getter functions (`isPlaying`, `getCurrentTime` and `getDuration`) were added to provide access to these properties, which is needed in some cases.

Before there was a 'hacky' way of accessing them via the
`indexing operator` (`[]`).

```ts
videoRef()['playing']()
```

With these changes, these property values can be retrieved in a cleaner and secure way without interfering with core logic.

```ts
videoRef().isPlaying()
```

---

Resolves: https://github.com/motion-canvas/motion-canvas/issues/239

### DIFF
--- a/packages/2d/src/components/Video.ts
+++ b/packages/2d/src/components/Video.ts
@@ -51,6 +51,18 @@ export class Video extends Rect {
     super(props);
   }
 
+  public isPlaying(): boolean {
+    return this.playing();
+  }
+
+  public getCurrentTime(): number {
+    return this.time();
+  }
+
+  public getDuration(): number {
+    return this.video().duration;
+  }
+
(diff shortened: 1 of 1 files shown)


aarthificial (changes requested):
Instead of exposing these internal signals, we should add dedicated methods for retrieving them. These signals should be modified in a specific way via `play`, `pause`, and `seek` but now users can modify them manually, possibly breaking the component.

I propse adding: `getDuration`, `getCurrentTime` and `isPlaying` instead.

Tandashi:
Oh yeah, you're absolutely correct. No clue what I was doing here... Will update in a bit.

aarthificial (approved):

PR #245, opened 2023-02-08
fix(packaging): change ui:dev script to ui:serve based on README

andrewpap22:
Review request: @aarthificial 

Not sure if it was better to change the README.md or the package.json but I went with package.json for consistency with the other scripts. 

Quoting from the README: 

```md
If you want to develop the UI, first build the template project by running: template:build. Next, start ui:serve. ⬅️ This was ui:dev in the package.json
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "2d:build": "npm run build -w packages/2d",
     "2d:watch": "npm run watch -w packages/2d",
     "ui:build": "npm run build -w packages/ui",
-    "ui:dev": "npm run dev -w packages/ui",
+    "ui:serve": "npm run dev -w packages/ui",
     "ui:showcase": "npm run showcase -w packages/ui",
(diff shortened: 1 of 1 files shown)


aarthificial:
Thanks for the PR!
As a side note: `fix(packaging)` is not a valid scope. To quote the [Angular Commit Message Conventions](https://github.com/angular/angular/blob/main/CONTRIBUTING.md#commit):
> The scope should be the name of the npm package affected

You can find all valid scopes here: https://github.com/motion-canvas/motion-canvas/blob/4c7dcc34272429900f4b52166571da304f4a83b3/commitlint.config.js#L8-L19

Additionally, since we use commit messages to generate changelogs, it's best to reserve the `fix` and `feat` types for changes that affect the user. This is a change that only affects collaborators running the project locally, so `build` would be a more suitable type.

PR #255, opened 2023-02-08
Project variables

### LINKED ISSUE #222: Pass variables from player package to project scenes
**Description**
I would like to be able to pass arguments to a projects scenes via the player package such as image sources, aspect ratios or colours. 
I feel this would open up some interesting possibilities for the package such as dynamically changing images, colours or text.

**Proposed solution**
Perhaps adding an attribute to the player package component to provide a single props object:

```
<motion-canvas-player
      src="/@id/__x00__virtual:template"
      style="aspect-ratio: 16 / 9"
      data-props="{ imageSource: "https://mycdn.com/image", colours: ['red', 'blue'] }"
></motion-canvas-player>
```

This would then be available to each scene in the project:

```
export default makeScene2D(function* (view, props) {
      const image = createRef<Image>()
      view.add(
        <Image
          ref={image}
          src={props.imageSource}
          width={200}
          height={200}
        />
      )
});
```

Sorry this is light on actual implementation details, really happy to be guided around the best solution here.

**Considered alternatives**

**Additional context**
Thanks again for this project, it's great work!

therealtimhawkins:
Closes #222 

Adds project variables :) Happy to add documentation in this PR or separately.

Can be tested by updating the example to include:
```
  const circleFill = useScene().variables.get('circleFill');
  view.add(<Circle ref={circle} width={320} height={320} fill={circleFill} />);
```

And either by adding variables to makeProject like:
```
export default makeProject({
  scenes: [example],
  background: '#141414',
  variables: {circleFill: 'teal'},
});
```

Or by updating the player html (this shows variables being updated during the animation):
```
  <body>
    <script type="module" src="/src/main.ts"></script>
    <button id="change-colour-variable">Change Colour</button>
    <motion-canvas-player
      src="/@id/__x00__virtual:template"
      style="aspect-ratio: 16 / 9"
      variables='{"circleFill":"red"}'
    ></motion-canvas-player>

    <script>
      const colourButton = document.getElementById('change-colour-variable');

      colourButton.addEventListener('click', () => {
        console.log('clicked!');
        const player = document.querySelector('motion-canvas-player');
        const currentVariables = player.getAttribute('variables');

        currentVariables.includes('teal')
          ? player.setAttribute(
              'variables',
              JSON.stringify({circleFill: 'red'}),
            )
          : player.setAttribute(
              'variables',
              JSON.stringify({circleFill: 'teal'}),
            );
      });
    </script>
  </body>
```

### DIFF
--- a/packages/core/src/Project.ts
+++ b/packages/core/src/Project.ts
@@ -27,6 +27,7 @@ export interface ProjectConfig {
   canvas?: HTMLCanvasElement;
   size?: Vector2;
   background?: string | false;
+  variables?: {[key: string]: any};
 }
 
 export enum PlaybackState {
@@ -148,6 +149,17 @@ export class Project {
     return this.size;
   }
 
+  public setVariables(variables: object) {
+    this.variables = variables;
+    if (this.scenes.current[0]) {
+      this.scenes.current[0].variables.updateSignals();
+    }
+  }
+
+  public getVariables(): {[key: string]: any} {
+    return this.variables;
+  }
+
   public readonly name: string;
   public readonly audio = new AudioManager();
   public readonly logger = new Logger();
@@ -170,6 +182,7 @@ export class Project {
   private previousBufferContext: CanvasRenderingContext2D | null = null;
   private exportCounter = 0;
   private size = Vector2.zero;
+  private variables: {[key: string]: any} = {};
 
   /**
    * @deprecated Use {@link makeProject} instead.
@@ -194,6 +207,7 @@ export class Project {
       canvas,
       size = ProjectSize.FullHD,
       background = false,
+      variables = {},
     } = typeof name === 'string' ? config! : name;
 
     this.name = typeof name === 'string' ? name : '';
@@ -202,6 +216,7 @@ export class Project {
     this.creationStack = new Error().stack ?? '';
     this.setCanvas(canvas);
     this.setSize(size);
+    this.setVariables(variables);
     this.background = background;
 
     if (audio) {
--- a/packages/core/src/Project.ts
+++ b/packages/core/src/Project.ts
@@ -27,7 +27,7 @@ export interface ProjectConfig {
   canvas?: HTMLCanvasElement;
   size?: Vector2;
   background?: string | false;
-  variables?: {[key: string]: any};
+  variables?: Record<string, unknown>;
 }
 
 export enum PlaybackState {
@@ -149,11 +149,9 @@ export class Project {
     return this.size;
   }
 
-  public setVariables(variables: object) {
+  public setVariables(variables: Record<string, unknown>) {
     this.variables = variables;
-    if (this.scenes.current[0]) {
-      this.scenes.current[0].variables.updateSignals();
-    }
+    this.currentScene.current?.variables.updateSignals();
   }
 
   public getVariables(): {[key: string]: any} {
@@ -182,7 +180,7 @@ export class Project {
   private previousBufferContext: CanvasRenderingContext2D | null = null;
   private exportCounter = 0;
   private size = Vector2.zero;
-  private variables: {[key: string]: any} = {};
+  private variables: Record<string, unknown> = {};
 
   /**
    * @deprecated Use {@link makeProject} instead.
--- a/packages/core/src/scenes/GeneratorScene.ts
+++ b/packages/core/src/scenes/GeneratorScene.ts
@@ -7,6 +7,7 @@ import {
 } from '../threading';
 import {Meta} from '../Meta';
 import {TimeEvents} from './TimeEvents';
+import {Variables} from './Variables';
 import {EventDispatcher, ValueDispatcher} from '../events';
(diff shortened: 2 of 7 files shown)


packages/core/src/Project.ts
```diff
@@ -170,6 +182,7 @@ export class Project {
   private previousBufferContext: CanvasRenderingContext2D | null = null;
   private exportCounter = 0;
   private size = Vector2.zero;
+  private variables: {[key: string]: any} = {};
```
hhenrichsen, line 185: I would suggest that we use a type like `unknown` here instead of `any` -- `any` turns off the type system, and since we return this to user-code that means we can pollute the typing in that code, too. The way using `unknown` would require users to ensure that the project variables are the types they are expecting (or they can cast them if they so choose, but then they're making the choice).

packages/core/src/Project.ts
```diff
@@ -148,6 +149,17 @@ export class Project {
     return this.size;
   }
 
+  public setVariables(variables: object) {
+    this.variables = variables;
+    if (this.scenes.current[0]) {
+      this.scenes.current[0].variables.updateSignals();
+    }
```
aarthificial, line 156: It should be:
```suggestion
    this.currentScene.current?.variables.updateSignals();
```
There can be more than one scene in a project. We're interested in updating the current one, not the first one.

packages/core/src/Project.ts
aarthificial: Btw, should be `Record<string, unknown>` to stay consistent with the rest of the codebase.

packages/core/src/scenes/GeneratorScene.ts
```diff
@@ -263,6 +266,7 @@ export abstract class GeneratorScene<T>
     } else {
       this.state = SceneState.Initial;
     }
+    this.variables.destroySignals();
```
aarthificial, line 269: Similarly to `TimeEvents`, `Variables` should subscribe to scene events themselves.
In the constructor of `Varialbes` you should do:
```ts
scene.onReset.subscribe(this.handleReset);
```
`GeneratorScene` is only the default implementation of the `Scene` interface. There will be more in the future. Doing registration in the `Variables` class makes it easier to reuse them in different scene implementations.

packages/core/src/scenes/Variables.ts
```diff
@@ -0,0 +1,47 @@
...
+  private signals: {[key: string]: SimpleSignal<any>} = {};
+
+  public constructor(private readonly scene: Scene) {}
+
+  /**
+   * Get variable signal if exists or create signal if not
+   *
+   * @param name - The name of the variable.
+   */
+  public get(name: string) {
```
aarthificial, line 14: This method should have a generic type representing the returned value:
```ts
public get<T>(name: string): SimpleSignal<T> {
```

packages/core/src/scenes/Variables.ts
```diff
@@ -0,0 +1,47 @@
...
+    const variables = this.scene.project.getVariables();
+    if (!(name in variables)) {
+      this.scene.project.logger.warn(
```
aarthificial, line 17: I'd consider adding another argument to this method to specify the default value of a variable:
```
public get<T>(name: string, initial: T): SimpleSignal<T> {
```
and return it here instead.

aarthificial (changes requested):

packages/core/src/scenes/Variables.ts
```diff
@@ -0,0 +1,49 @@
...
+  public constructor(private readonly scene: Scene) {
+    scene.onReset.subscribe(this.handleReset);
+  }
+
+  /**
+   * Get variable signal if exists or create signal if not
+   *
+   * @param name - The name of the variable.
+   */
+  public get<T>(name: string, initial: T): SimpleSignal<T> | null {
```
aarthificial, line 16: One last tiny issue, I believe it's no longer possible for this method to return null:
```suggestion
  public get<T>(name: string, initial: T): SimpleSignal<T> {
```

aarthificial (changes requested):

aarthificial:
I've just noticed your question about documentation. Feel free to add it to this PR if you'd like. There aren't many conventions in this repo yet, so whatever works for you.

therealtimhawkins:
Added the docs in here just for simplicity 👍 Thanks for your patience mate!

aarthificial (approved):

aarthificial:
Awesome, thanks so much for the PR!

flomero:
i tried to get the example from the docs to work and got the following error. is there a second argument needed?
if so I think this should be added to the docs

```
Expected 2 arguments, but got 1.ts(2554)
Variables.d.ts(12, 26): An argument for 'initial' was not provided.
```

aarthificial:
@flomero Seems like there's a small mistake, you need to provide a default value as well:
```ts
const circleFill = useScene().variables.get('circleFill', '#fff');
```

PR #257, opened 2023-02-09
Warn when tweening undefined values

### LINKED ISSUE #243: Attempting to lerp without an initial value results in an opaque error message
**Description**
Trying to lerp a value without an initial value results in an error message like:
```
Cannot use 'in' operator to search for 'lerp' in null
```

**Proposed solution**
The error message should instead check for a from value and throw a clearer error when trying to lerp a value that does not yet exist.

hhenrichsen:
Closes #243.

I'm not 100% happy on this, but it's tricky to get a warning / error in place without disrupting intended functionality with lerping Maps. I'm open to suggestions here on what might be better.

### DIFF
--- a/packages/core/src/tweening/interpolationFunctions.ts
+++ b/packages/core/src/tweening/interpolationFunctions.ts
@@ -1,3 +1,4 @@
+import {useLogger} from '../utils';
 import {Vector2} from '../types';
 
 export interface InterpolationFunction<T, Rest extends any[] = any[]> {
@@ -56,6 +57,10 @@ export function deepLerp<
   TFrom extends Record<any, unknown>,
   TTo extends Record<any, unknown>,
 >(from: TFrom, to: TTo, value: number): TFrom | TTo;
+export function deepLerp<
+  TFrom extends Record<any, unknown>,
+  TTo extends Record<any, unknown>,
+>(from: TFrom, to: TTo, value: number, suppressWarnings: boolean): TFrom | TTo;
 /**
  * Interpolate between any two values, including objects, arrays, and Maps.
  *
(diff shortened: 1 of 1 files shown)


packages/core/src/tweening/interpolationFunctions.ts
```diff
@@ -66,11 +71,27 @@ export function deepLerp<
...
+export function deepLerp(
+  from: any,
+  to: any,
+  value: number,
+  suppressWarnings = false,
+): any {
   if (value === 0) return from;
   if (value === 1) return to;
 
-  if (typeof from === 'undefined' || typeof to === 'undefined') {
```
hhenrichsen, line 73: This was letting `null` slip through as well, which caused issues further down the function.

packages/core/src/tweening/interpolationFunctions.ts
```diff
@@ -102,7 +123,7 @@ export function deepLerp(from: any, to: any, value: number): any {
       if (from instanceof Map && to instanceof Map) {
         const result = new Map();
         for (const key of new Set([...from.keys(), ...to.keys()])) {
-          const inter = deepLerp(from.get(key), to.get(key), value);
+          const inter = deepLerp(from.get(key), to.get(key), value, true);
```
aarthificial, line 126: Shouldn't the value of `suppressWarnings` be passed here recursively?

hhenrichsen: No. I'm intentionally suppressing warnings on `Map`s, where we intentionally use the nullish behavior.

aarthificial (approved):

PR #259, opened 2023-02-09
feat(core): add @inherited() decorator

### LINKED ISSUE #246: Text fontSize property can't be updated
When trying to update the fontSize motion-canvas freezes.

**To reproduce:**
- Create a text element and update it's fontSize in a `yield* all(...)` such as:
```ts
yield* all(
    ..., 
    Text().fontSize(24, 1)
)
``` 

**Expected behavior**
The fontSize will update like other properties such as Size, Position, etc...

**Console errors**
> Cannot use 'in' operator to search for 'lerp' in null

**Package versions**
> Core: v2.0.0
> UI: v2.0.0

exdeejay:
Adds an @inherited() decorator that, when applied to a property, creates a signal of type InheritedSignal. This signal can have an initial value, but if it doesn't, it will call `compute*PropertyName*()` on the owner class to compute the inherited value. Setting a value on the signal disables the inheritance on the signal until `signal.reset()` is called.

This fixes #246 while preserving font size inheritance,  and can also be used in place of the `customX` and `customY` signals for positioning. Not tested with inheriting layout properties yet, though.

Currently this PR isn't ready to be merged (since I'd like to change some more font properties over to using @inherited() before then); it's just to start some discussion on if this is a good pattern to pursue.


aarthificial:
I'd consider whether the additional Signal Context class is necessary, the same can be written using basic signals:
```ts
protected readonly inheritedFontSize: () => number;
  
public constructor(props: LayoutProps) {
  const inheritedFontSize = () => {
    this.requestLayoutUpdate();
    return parseFloat(this.styles.getPropertyValue('font-size').slice(0, -2));
  };
  
  super({
    fontSize: inheritedFontSize,
    ...props,
  });
  
  this.inheritedFontSize = inheritedFontSize;
}
  
@computed()
protected applyFont() {
  // this does the same thing as `isInheriting`:
  // (`collect` would have to be made public)
  this.fontSize.context.collect();
  this.element.style.fontSize =
    this.fontSize.context.raw() !== this.inheritedFontSize
      ? this.parsePixels(this.fontSize())
      : '';
}
```

Both solutions require roughly the same amount of code to implement.

aarthificial:
On a separate note: `customX` and `customY` work slightly differently - they are used depending on whether the node is part of a layout.

exdeejay:
The separate InheritedSignalContext would be for ease of use, since having to call `collect()` and check `raw()` manually is an arcane way of using signals for something that could be used for any number of CSS styles.

I took another look at `customX` and `customY`, and I think you're right in that this wouldn't apply there.

exdeejay:
I'd say it's more about making an easier way to have a default computed value. The only good way to do that with signals is by writing out all the default value functions in the `super()` call (since we can't use/bind `this` until after `super()`), but with this, you can declare a default function for the signal in-class and also easily tell in user code if the value you're looking at is that default calculated value. I'd rename this to ComputedSignal and `isComputed()` if there wasn't already one.

exdeejay:
How about a static util function?

```ts
@computed()
protected applyFont() {
  this.element.style.fontSize = isDefault(this.fontSize) ? '' : this.parsePixels(this.fontSize());
}

function isDefault<T>(signal: SimpleSignal<T>) {
  signal.context.collect();
  return signal.context['compute'] && signal.context.raw() === signal.context['compute'];
}
```

aarthificial:
Okay 👍

packages/core/src/signals/isDefault.ts
```diff
@@ -0,0 +1,11 @@
+import {Signal} from './SignalContext';
+
+export function isDefault<TSetterValue, TValue extends TSetterValue, TOwner>(
```
aarthificial, line 3: Okay, if we're putting it in the `core` package, we might as well make it a method of `SignalContext`

packages/2d/src/utils/patchSignal.ts
```diff
@@ -27,5 +27,12 @@ export function patchSignal<TSetterValue, TValue extends TSetterValue>(
     if (tweener) {
       signal.doTween = tweener.bind(owner);
     }
+    const computer = owner?.[`compute${capitalize(name)}`];
+    if (computer) {
+      signal.compute = computer.bind(owner);
+      if (signal['initial'] === undefined) {
```
aarthificial, line 33: I think we can just make `initial` public since there's already a setter for it and now we're also getting it.

packages/2d/src/components/Layout.ts
```diff
@@ -157,18 +158,32 @@ export class Layout extends Node {
...
   public declare readonly fontStyle: SimpleSignal<string | null, this>;
   @initial(null)
   @signal()
   public declare readonly fontWeight: SimpleSignal<number | null, this>;
-  @initial(null)
   @signal()
-  public declare readonly lineHeight: SimpleSignal<number | null, this>;
+  public declare readonly lineHeight: SimpleSignal<number, this>;
+
+  protected computeLineHeight(): number {
```
aarthificial, line 178: There's already a way to make the line height depend on the font size in CSS. I believe that's how we should do that, similarly to how `width` can have different units.

packages/2d/src/components/Layout.ts
```diff
@@ -157,18 +158,32 @@ export class Layout extends Node {
...
   @signal()
   public declare readonly fontFamily: SimpleSignal<string | null, this>;
-  @initial(null)
   @signal()
-  public declare readonly fontSize: SimpleSignal<number | null, this>;
+  public declare readonly fontSize: SimpleSignal<number, this>;
+
+  protected computeFontSize(): number {
+    this.requestLayoutUpdate();
+    return Number(this.styles.getPropertyValue('font-size').slice(0, -2));
```
aarthificial, line 166: You could use `parseFloat`:
```suggestion
    return parseFloat(this.styles.getPropertyValue('font-size'));
```

packages/2d/src/components/Layout.ts
```diff
@@ -157,18 +158,32 @@ export class Layout extends Node {
   @initial(null)
   @signal()
   public declare readonly fontFamily: SimpleSignal<string | null, this>;
-  @initial(null)
   @signal()
-  public declare readonly fontSize: SimpleSignal<number | null, this>;
+  public declare readonly fontSize: SimpleSignal<number, this>;
+
+  protected computeFontSize(): number {
+    this.requestLayoutUpdate();
```
aarthificial, line 165: This is a problem btw. I've noticed this when testing. Technically, `this.requestFontUpdate()` should be enough to update all important font values but for some reason it doesn't work. `this.requestLayoutUpdate()` is a bit more expensive so it would be nice to get the former to work.

packages/2d/src/utils/patchSignal.ts
```diff
@@ -27,5 +27,12 @@ export function patchSignal<TSetterValue, TValue extends TSetterValue>(
     if (tweener) {
       signal.doTween = tweener.bind(owner);
     }
+    const computer = owner?.[`compute${capitalize(name)}`];
```
aarthificial, line 30: Maybe it should be called `initial[Property]` or `default[Property]` or `getDefault[Property]`?
`compute` may be a bit confusing as it is only used when the signal is not overridden by the user.

aarthificial (changes requested):

aarthificial:
> lineHeight isn't updated in the inspector until it's read.

Wait, why would it be updated? The value is only applied to CSS when the lineHeight is NOT default. In other words, the value computed by `computeLineHeight` will never be assigned to the element's style.

packages/core/src/signals/isDefault.ts
exdeejay: I'd argue for using a separated method, since it doesn't feel right to have normal users accessing `signal.context` methods when there can be an exposed method for doing so. Putting it on the `invokable` would also be fine.

packages/2d/src/components/Layout.ts
exdeejay: I abstracted these functions into `@computeStyle`, which always uses this.requestLayoutUpdate(). Do you think the performance gains here warrant the creation of a `@computeFontStyle` too?

exdeejay:
Triaged the `computeLineHeight` not updating in the inspector: It seems like it only updated when tweening (which implicitly makes the value real). That being said, it actually does update in the inspector--but the bounding box of the text in the view window does not. So the caching issue is with the drawing of the bounding box.

packages/2d/src/utils/patchSignal.ts
exdeejay: I'll leave that up to you; I don't really have a good idea for what to call it

exdeejay:
Turns out computed style values have the nice side effect of populating `unset` values in the inspector!

brandonpelfrey:
@exdeejay would you like to take ownership of the lineHeight issue if this change will solve it? Glad we can solve multiple things at once!

exdeejay:
Flip-flopped a lot in the course of implementing this one. Returning default values make it easy to read in the inspector and user code, but also some default values (like `auto` and `none`) are meaningless and are better off represented as `null` so they show as `unset` in the inspector. Therefore I relegated computing default values to only the font styling signals.

I think computing values could also work for `gap`, `rowGap`, and `columnGap`, but since `gap` combines `rowGap` and `columnGap` it should be made into a compound signal, and that's a bit too out of scope for this PR so I'll leave them alone.

I also tried to emphasize nulls wherever possible, i.e. making `Length` into `Length | null` for values that can have defaults, and removing the null check from `parseLength()`. I'm worried that throwing around nulls everywhere will eventually cause problems such as #257, so value types should be null-less by default. I can revert these if you don't think it's necessary.

aarthificial:
I don't really understand how moving `null` out of `Length` prevents things like #257
The issue is caused by a lacking implementation of tweening for `CanvasStyle` and is not inherently caused by `null`s - tweening between a `Gradient` and a `Color` also won't work.

`null` is a valid option for length that represents `auto` and tweening works as expected when using it. Using `null`s to represent unset values was a design decision I made and I'm not convinced to change it so please revert these changes.

exdeejay:
Pardon the flood of commits, rebased what I had onto main and changed appropriate commits to fixups. The null stuff has been reverted and I think this PR's ready to go.

exdeejay:
> The issue is caused by a lacking implementation of tweening for CanvasStyle and is not inherently caused by nulls - tweening between a Gradient and a Color also won't work.

To elaborate on what I mean, currently if you tween gap like so:
```ts
view.add(
  <Rect ref={rect} layout>
    {range(5).map((i) => 
      <Circle size={100} fill="#3cf" />
    )}
  </Rect>
);
yield* rect().gap(200, 1);
```

The tween will not work correctly and will just jump to 200px wide at 1 second, since the default value for `gap` is null. Coming from programming in Rust, I'm just uneasy with representing automatic values as `null` due to `null`'s terrible track record for causing bugs like these.

aarthificial:
We can consider changing the type of `gap` to something else (in a different issue of course) because in this case there's no difference between the `normal` and `0` CSS values. 

My point is that `null` should be treated as one of the possible Length values and handled properly by the signals that use it, instead of writing `Length | null` everywhere. Especially since the type system won't cover custom tween methods. Notice how you updated the size signal to `Length | null`, but you left `tweenSize`, etc. with just `Length`, which would imply that the "valid" Length is the only one the developer should expect.

exdeejay:
True. I'd love to discuss this more in the future but I'll leave it for now since it's not relevant to this PR.

aarthificial (changes requested):
Some of my comments are still unresolved

aarthificial:
I'm taking a fresh look at this and something that I don't understand is how to make these signals inherit their values again.
Say I do something like this:
```tsx
<Layout layout fontSize={48}>
  <Text fontSize={24} />
</Layout>
```
How do I make `Text` inherit the `fontSize` again?
Or if I want to do a tween: `inherited size -> 24px -> back to inherited size`?

exdeejay:
My first thought is to have some sort of constant that represents an inherited value, so that tweening would look like this:
```ts
yield* value(Signal.DEFAULT, 1.0);
```
Of course, `null` could be that constant (as much as I hate to admit it), but it may cause problems with non-inherited properties being able to actually contain null.

aarthificial:
I really like the idea of a `DEFAULT` constant, but I think we need to simplify some things.
Right now we have two duplicated concepts. There's `null` for sizes and `DEFAULT` for inherited properties. Also `initial` for normal signals and `getDefault` for property signals.

I think we can merge `initial` and `default` together if we stop treating user-provided properties as the default values of signals.
Right now if you do:
```tsx
<Text fontSize={24} />
```
`24` will become the initial value of `fontSize`. It made sense when I implemented this because it allowed custom components to provide their own initial values through a call to `super`. But with your `getDefault[Property]` functionality, this problem is now solved. So we could modify the signal decorator:
```ts
const signal = new SignalContext<T, T, any>(
  owner[`getDefault${proeprty}`] ?? meta.default, // to always use the `getDefault` 
                                                  // or `@initial()` as the initial signal value
  meta.interpolationFunction ?? deepLerp,
  instance,
);
if (context.defaults[key]) {
  signal(context.defaults[key]); // and simply set the user-provided value
}
```
This would merge the `default` and `initial` signal properties into one.

I'm also okay with switching to the `DEFAULT` constant for Length as it sounds much more elegant. 
This way we'd end up with a unified approach to all signal properties.

I'm curious what you think about this.
Also, I know this PR is taking ages so let me know if you're not interested in doing this in the context of this issue.

exdeejay:
> I know this PR is taking ages so let me know if you're not interested in doing this in the context of this issue.
Nah, I like that we're making a better implementation to begin with. I feel like we're near the finish line, anyways.

> I think we can merge initial and default together if we stop treating user-provided properties as the default values of signals.
The problem with this is that calling `signal.reset()` won't reset the signal to the user-provided value, which may be confusing. Also, users may perceive the provided `fontSize` value in the `view.add()` to be the default and get confused when assigning the `DEFAULT` constant to a signal and getting the inherited value. This is initially why I was more biased to using the words "inherited" and "computed". However, maintaining both an initial and a default value is just as confusing and cumbersome, so this may just be a design decision to make. I'll think about it for a bit longer and see if I come up with a more elegant solution.

For the `DEFAULT` constant: ideally it should be a type outside of the normal assignable values for a signal. I was thinking using a `Symbol` like the signal metadata, but I've never used them before so I don't know if they are appropriate here. Thoughts?

aarthificial:
> The problem with this is that calling signal.reset() won't reset the signal...

I'd say it's arguable, `reset` resets the value to the one provided when the signal was created - the users don't create the signal, there's no reason for them to expect that their value is somehow used during creation. Either way, people will get confused once and then they will learn. Also, speaking from experience, the `reset` method is rarely useful. I don't think I've ever used it.

Yes, a symbol would be perfect.

exdeejay:
If I set a default font size, I think I'd want `Signal.DEFAULT` to mean the default I set, to eliminate the need for me to store my own constant with my default font size.

Another idea is to simply not provide the ability to go back to inheriting a value once an initial value is provided:
```tsx
const text = createRef<Text>();
view.add(
  <Text ref={text} text="Hello world" fontSize={64} />
)
yield* text().fontSize(Signal.DEFAULT, 1); // would lerp to 64 instead of an inherited value
```

Then, if a user wants to actually go back and forth between an inherited value, they would have to defer putting in their initial size:
```tsx
const text = createRef<Text>();
view.add(
  <Text ref={text} text="Hello world" />
)
text().fontSize(64); // sets the desired font size while keeping the inherited as a default
yield* text().fontSize(Signal.DEFAULT, 1); // would lerp to the inherited value
```

Or, we could add two separate symbolic constants: `Signal.DEFAULT` and `Signal.INHERITED` (or a different name), to allow for lerping to both user-provided and computed defaults, while disambiguating what each one does.

aarthificial:
Storing your own constant:
```ts
const text = createRef<Text>();
const fontSize = 64;
view.add(
  <Text ref={text} text="Hello world" fontSize={fontSize} />
)
yield* text().fontSize(fontSize, 1);
```
has the same complexity as deferring putting the value:
```ts
const text = createRef<Text>();
view.add(
  <Text ref={text} text="Hello world" />
)
text().fontSize(64);
yield* text().fontSize(Signal.DEFAULT, 1);
```
IMO, we should always reset to the Node's default value because it makes the implementation straightforward.
If we'll notice that the community often needs to reset properties to the values they set, then we'll add a way to easily do that.
I don't think we should "what if" about this too much.

exdeejay:
Sounds good, I'll make it so 👍

exdeejay:
One more question: should `isDefault()` remain as a way to check if the signal is the initial value, or should that go back to being a call to `raw()`? `this.collect()` still needs to be called when default checking to update the computed methods it's used in.

aarthificial:
I think it should stay. But if you're using the `initial` property on the `SignalContext`, it should probably be called `isInitial` and not `isDefault`

exdeejay:
The `DEFAULT` symbol is currently on `SignalContext.DEFAULT`. Though, I feel like putting such a value on what I consider to be an internal class is a bad idea. I also thought about putting it in `core/lib/symbols` or as a property on the signal object (`signalName.default()` always returns the `DEFAULT` symbol). Let me know your thoughts.

Besides that, I brought back `setInitial()` and made `initial` private again, since nothing outside of the class is using it now.

GufNZ:
> tweening between a Gradient and a Color also won't work.
Random thought that I wasn't sure where else to put:  Gradient -> Color could tween all the gradient's colour stops ->newColour then switch at the end when the difference is no longer visible.  Similarly in reverse.

exdeejay:
That's a good idea that deserves its own issue. Can you make one @GufNZ?

packages/core/src/signals/SignalContext.ts
```diff
@@ -101,21 +113,26 @@ export class SignalContext<
     this.markDirty();
   }
 
-  public set(value: SignalValue<TSetterValue>): TOwner {
+  public set(value: SignalValue<TSetterValue> | symbol): TOwner {
```
aarthificial, line 116: Should be:
```suggestion
  public set(value: SignalValue<TSetterValue> | typeof SignalContext.DEFAULT): TOwner {
```

packages/core/src/signals/SignalContext.ts
```diff
@@ -257,6 +279,27 @@ export class SignalContext<
     return this.set(this.get());
   }
 
+  /**
+   * Checks if the signal is currently using its initial value.
```
aarthificial, line 283: Docs for functions should use imperative present tense:
```suggestion
   * Check if the signal is currently using its initial value.
```

packages/core/src/signals/SignalContext.ts
```diff
@@ -49,6 +54,10 @@ export class SignalContext<
   TValue extends TSetterValue = TSetterValue,
   TOwner = void,
 > extends DependencyContext<TOwner> {
+  public static readonly DEFAULT = Symbol.for(
```
aarthificial, line 57: As you've mentioned in your comment, I think `DEFAULT` should be exported as a separate symbol, not a class member.

packages/2d/src/partials/types.ts
```diff
@@ -44,7 +46,6 @@ export type LayoutMode = boolean | null;
  * - `number` - the desired length in pixels
  * - `${number}%` - a string with the desired length in percents, for example
  *                  `'50%'`
- * - `null` - an automatic length (equivalent to `auto` in CSS)
```
aarthificial, line 47: `null` should stay until we switch to `DEFAULT`

packages/2d/src/decorators/signal.ts
```diff
@@ -88,8 +88,10 @@ export function signal<T>(): PropertyDecorator {
   return (target: any, key) => {
     const meta = getPropertyMetaOrCreate<T>(target, key);
     addInitializer(target, (instance: any, context: any) => {
```
aarthificial, line 90: `context` is no longer necessary here

packages/2d/src/decorators/defaultStyle.ts
```diff
@@ -0,0 +1,18 @@
...
+export function defaultStyle<T>(
+  styleName: string,
+  parse: (value: string) => T = value => value as T,
+): PropertyDecorator {
+  return (target: any, key) => {
+    target[`getDefault${capitalize(<string>key)}`] = function (this: Layout) {
+      this.requestLayoutUpdate();
+      const old = (<any>this.element.style)[styleName];
+      (<any>this.element.style)[styleName] = '';
+      const ret = parse.apply(this, [this.styles.getPropertyValue(styleName)]);
```
aarthificial, line 13: Could be:
```suggestion
      const ret = parse.call(this, this.styles.getPropertyValue(styleName));
```

packages/2d/src/components/Layout.ts
```diff
@@ -729,29 +736,45 @@ export class Layout extends Node {
...
+      : this.fontFamily();
+    this.element.style.fontSize = this.fontSize.isInitial()
+      ? ''
+      : `${this.fontSize()}px`;
+    this.element.style.fontStyle = this.fontStyle.isInitial()
+      ? ''
+      : this.fontStyle();
+    this.element.style.lineHeight =
+      typeof this.lineHeight() === 'string'
+        ? String(Number((this.lineHeight() as string).slice(0, -1)) / 100)
```
aarthificial, line 757: The rest of the code base uses `parseFloat`, `parseInt` and `toString`, so we should try to be consistent:
```suggestion
        ? (parseFloat(this.lineHeight() as string) / 100).toString()
```

packages/2d/src/components/Layout.ts
```diff
@@ -154,25 +156,33 @@ export class Layout extends Node {
...
   @signal()
-  public declare readonly fontSize: SimpleSignal<number | null, this>;
-  @initial(null)
+  public declare readonly fontSize: SimpleSignal<number, this>;
+  @defaultStyle('font-style')
   @signal()
-  public declare readonly fontStyle: SimpleSignal<string | null, this>;
-  @initial(null)
+  public declare readonly fontStyle: SimpleSignal<string, this>;
+  @defaultStyle('font-weight', Number)
```
aarthificial, line 168:
```suggestion
  @defaultStyle('font-weight', parseInt)
```

aarthificial (changes requested):

packages/core/src/signals/SignalContext.ts
exdeejay: moved to `core/src/signals/symbols.ts`

packages/core/src/signals/SignalContext.ts
```diff
@@ -101,21 +110,26 @@ export class SignalContext<
     this.markDirty();
   }
 
-  public set(value: SignalValue<TSetterValue>): TOwner {
+  public set(value: SignalValue<TSetterValue> | typeof DEFAULT): TOwner {
+    if (value === DEFAULT) {
+      value = this.initial!;
+    }
+    const resolvedValue = value as SignalValue<TSetterValue>;
```
aarthificial, line 117: With `typeof DEFAULT` this is no longer necessary bc typescript will do it's magic type narrowing.

aarthificial (changes requested):
Last small thing

aarthificial (approved):

PR #263, opened 2023-02-09
update Project to makeProject

younissk:
I think it might have used to benn "Project" in the project.ts file, but has been changed, since when I just first initialsed the project, it was different.


aarthificial:
Thanks for the PR!

PR #275, opened 2023-02-09
docs: add easing function visualiser animation

DoctorBracewell:
When developing additional easing functions for #274 I created a easing visualiser animation, and I thought it might be good to include a version of it in the tweening docs to illustrate how the eased value and time are related.

It's definitely not absolutely necessary to include but I feel it helps visualise how easing works for people learning about programmatic animations.


DoctorBracewell:
Ah, the build fails because currently the animation demonstrates the `easeInOutBounce` function - which isn't yet implemented from #274. This could be switched to an existing function or delayed until that function is merged. My bad for creating the PR with the invalid function.

aarthificial:
If you're interested, we now have [project variables](https://motion-canvas.github.io/docs/project-variables) that let you customize the animation after it was embedded. You could use that to add a select on the documentation page that would dynamically pick different easing functions to preview.

aarthificial:
Thanks for the PR! Let me know if you're interested in my comment above.

PR #279, opened 2023-02-09
feat: added a theme property to the CodeBlock component

Ross-Esmond:


Ross-Esmond:
I don't know if this is right. Signals require a few fiddly bits that I haven't kept up to date with.

aarthificial (approved):

aarthificial:
Looks good 👍

PR #280, opened 2023-02-09
fix(2d): call spawnChildren() in constructor to initialize spawner parents

### LINKED ISSUE #270: Element disappears when scrolling forward, works as expected when scrolling backwards.
**Describe the bug**
When creating an element and changing the absolute position of an element, the element dissapears when scrolling forwards and works as expected when scrolling backwards.

**To Reproduce**
Dont know the exact piece of my code that causes it so I will just leave all the code used here.

Components/Array.tsx
```ts
import { Layout, LayoutProps, Node, Rect, ShapeProps, Text } from "@motion-canvas/2d/lib/components";
import { easeInOutCubic, tween } from "@motion-canvas/core/lib/tweening"
import { Color } from "@motion-canvas/core/lib/types/Color"
import { initial, signal } from "@motion-canvas/2d/lib/decorators";
import { SignalValue, SimpleSignal } from "@motion-canvas/core/lib/signals";
import { ColorSignal, Spacing } from "@motion-canvas/core/lib/types";
import { makeRef } from "@motion-canvas/core/lib/utils";
import { range } from '@motion-canvas/core/lib/utils';
import { all } from "@motion-canvas/core/lib/flow"
// import { Colors, BlackLabel, WhiteLabel } from "../styles";

export interface ArrayProps extends ShapeProps, LayoutProps {
    values?: SignalValue<number[] | string[] >;
    boxWidth?: SignalValue<number>;
    boxHeight?: SignalValue<number>;
    boxGap?: SignalValue<number>;
}

export class Array extends Layout {
    @initial([])
    @signal()
    public declare readonly values: SimpleSignal<number[] | string[], this>

    @initial(128)
    @signal()
    public declare readonly boxWidth: SimpleSignal<number, this>

    @initial(128)
    @signal()
    public declare readonly boxHeight: SimpleSignal<number, this>

    @initial(28)
    @signal()
    public declare readonly boxGap: SimpleSignal<number, this>

    // Array that stores the references to the Rects
    public readonly boxArray: Rect[] = [];
    
    textStyle = {
        paddingTop: 10,
        fontFamily: 'JetBrains Mono',
        fill: 'rgba(255, 255, 255, 0.6)',
    };

    // @brief Returns value of the original array
    private getValue(Index: number){
        if(this.values()[Index] == undefined){
            return -1;
        }
        return this.values()[Index];
    }
    
    // @brief Pool to reduce playback lag in the animator
    public pool = range(64).map(i => (
        <Rect
            ref={makeRef(this.boxArray, i)}
            size={[this.boxWidth(), this.boxWidth()]}
            x={-((this.values().length * (this.boxWidth() + this.boxGap())) / 2) + i * (this.boxWidth() + this.boxGap()) + (this.boxWidth() + this.boxGap()) / 2}
            lineWidth={8}
            stroke={'#242424'}
            radius={new Spacing(4)}
        >
            <Text
                text={this.getValue(i).toString()}
                {...this.textStyle}
            />
        </Rect>
    ));


    public constructor(props?: ArrayProps) {
        super({
            ...props
        }); 

        this.add(
            <Node
                spawner={() => this.pool.slice(0, this.values().length)}
            >
            </Node>
        )
    }

    public * HighLight(Index: number, Duration: number, highlightColor: Color){
        yield* tween(Duration, color =>{
            this.boxArray[Index].stroke(
                Color.lerp(
                    new Color('#242424'),
                    new Color(highlightColor), 
                    easeInOutCubic(color),
                )
            )
        })
    }

    public * deHighLight(Index: number, Duration: number, highlightColor: Color){
        yield* tween(Duration, color =>{
            this.boxArray[Index].stroke(
                Color.lerp(
                    new Color(highlightColor), 
                    new Color('#242424'),
                    easeInOutCubic(color),
                )
            )
        })
    }

    public * Swap(Index1: number, Index2: number){
        let tempPosition = this.boxArray[Index1].absolutePosition();
        yield* all(
            this.boxArray[Index1].absolutePosition(this.boxArray[Index2].absolutePosition(), 1),
            this.boxArray[Index2].absolutePosition(tempPosition, 1),
        ) 
     }
}
```

Scenes/example.tsx
```ts
import {makeScene2D} from '@motion-canvas/2d/lib/scenes';
import { Color } from '@motion-canvas/core/lib/types';
import { createRef } from '@motion-canvas/core/lib/utils';
import {Array} from '../components/Array'

  // Create your animations here

  const Array1 = [1, 2]
  const ArrayRef = createRef<Array>();

  view.add(
    <Array
      ref={ArrayRef}  
      values={[0,1,2,3]}
    />
  )

  yield* ArrayRef().Swap(1, 2);
  yield* waitFor(5);
});
```

**Expected behavior**
The element should not disappear when scrolling forward.

**Console errors**
No console errors.

- core: [e.g. 2.1.0]
- ui: [e.g. 2.1.0]

**Additional context**

exdeejay:
Fixes #270 by calling the spawner function once during construction, to initialize the parents of the spawner objects. A warning in the docs about making sure to null check the parent of spawners would be prudent as well.

### DIFF
--- a/packages/2d/src/components/Node.ts
+++ b/packages/2d/src/components/Node.ts
@@ -33,12 +33,10 @@ import {Filter} from '../partials';
 import {filtersSignal, FiltersSignal} from '../decorators/filtersSignal';
 import {
   createSignal,
-  Signal,
   DependencyContext,
   SignalValue,
   SimpleSignal,
   isReactive,
-  Computed,
 } from '@motion-canvas/core/lib/signals';
 
 export interface NodeProps {
@@ -393,6 +391,7 @@ export class Node implements Promisable<Node> {
     this.add(children);
     if (spawner) {
       this.children(spawner);
+      this.spawnChildren();
     }
   }
 
--- a/packages/2d/src/components/Node.ts
+++ b/packages/2d/src/components/Node.ts
@@ -33,10 +33,12 @@ import {Filter} from '../partials';
 import {filtersSignal, FiltersSignal} from '../decorators/filtersSignal';
 import {
   createSignal,
+  Signal,
   DependencyContext,
   SignalValue,
   SimpleSignal,
   isReactive,
+  Computed,
 } from '@motion-canvas/core/lib/signals';
 
 export interface NodeProps {
@@ -391,7 +393,6 @@ export class Node implements Promisable<Node> {
     this.add(children);
     if (spawner) {
       this.children(spawner);
-      this.spawnChildren();
     }
   }
 
--- a/packages/docs/docs/advanced/spawners.mdx
+++ b/packages/docs/docs/advanced/spawners.mdx
@@ -77,4 +77,8 @@ yield* all(...spawnedCircles.map(function* (circle) {
 }));
 ```
 
+Be aware that the references returned by a call to `children()` may be invalidated
+when the number of spawned objects changes, and accessing the invalidated objects 
+may cause undefined behavior. Try not to save references to spawned objects for too long,
+and use `children()` wherever possible to get the updated list of spawned objects.
 
--- a/packages/docs/docs/advanced/spawners.mdx
+++ b/packages/docs/docs/advanced/spawners.mdx
@@ -62,5 +62,19 @@ const pool = range(64).map(i => (
   <Circle x={i * 32} width={32} height={32} fill={'lightseagreen'} />
 ));
 
-view.add(<Layout layout spawner={() => pool.slice(0, count())} />);
+const layout = createRef<Layout>();
+view.add(<Layout layout ref={layout} spawner={() => pool.slice(0, count())} />);
 ```
+
+Apart from the spawner function, the pool should never be accessed directly.
+Use the `children()` signal of the parent object to get references to the spawned objects:
+
+```tsx
+// ... continuing from above ...
+let spawnedCircles = layout().children() as Circle[];
+yield* all(...spawnedCircles.map(function* (circle) {
+  yield* circle.scale(1.5, 1).to(1, 1);
+}));
+```
+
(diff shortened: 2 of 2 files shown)


aarthificial (changes requested):
This only fixes one side effect and not the core issue.
The problem will come back when you change the signal that the spawner depends on:
```ts
view.add(<Array ref={ArrayRef} values={[0, 1]} />);
ArrayRef().values([0, 1, 2, 3, 4]); // happens after the constructor

yield* ArrayRef().Swap(1, 2);
```

The real issue here lies in the user land. When using spawners with a pool, you shouldn't access the pool directly. Instead, you should access the nodes thorough the `children` property of the spawner node:
```ts
export class Array {
  // ...
  
  // a reference to the spawner node
  public boxes: Node;

  // access the boxes through its `children`
  public get boxArray() {
    return this.boxes.children() as Rect[];
  }

  public constructor(props?: ArrayProps) {
    super({
      ...props,
    });

    this.add(
      <Node
        ref={makeRef(this, 'boxes')}
        spawner={() => this.pool.slice(0, this.values().length)}
      ></Node>,
    );
  }
}
```
This will make sure that everything works correctly.

Adding `this.spawnChildren()` to the base constructor comes with some issues on its own (can't use `this` in the spawner function, for example). 
It may also result in people developing bad habits of accessing objects from the pool. 

Let's add an explanation of this to the documentation instead.

exdeejay:
Reverted the above change and added guidance to use `children()` wherever possible. I'm not too familiar with how spawners work and there aren't any public examples I can find, so I can't add more guidance than that.

aarthificial (approved):

PR #282, opened 2023-02-09
feat(2d): add antialiased signal to Shape

### LINKED ISSUE #235: Option to _disable_ Anti-Aliasing
**Description**
I like to make short "explainer gifs" when making a big write-up and have great interest in motion-canvas.
I come from utilizing Adobe After Effects to make gifs such as this:
AVX2
SAD

The problem and biggest challenge I had with making these gifs though is that their file-size is very correlated by the characteristics of the motion and the amount of colors that are to be animated. One way that helps, in my experience, is to _disable_ anti-aliasing to reduce the amount of colors in exchange for "quality". This also helps when exporting a gif with transparency and there is no support for partial-transparency. This might also aid in viewport performance.

**Proposed solution**
An option to _disable_ anti-aliasing for the viewport and for rendering to turn off Anti-Aliasing on shapes, lines, text, and images.

**Additional context**
For SVGs, I believe the [shape-rendering](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/shape-rendering) option may be utilized which accomplishes what I am looking for:
image
Though, ideally this is a systematic option and not something that has to be configured per-shape.

For images, there is [CanvasRenderingContext2D.imageSmoothingEnabled](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/imageSmoothingEnabled) which would maintain aliasing on images.
image

There is also [CanvasRenderingContext2D.textRendering](https://devdocs.io/dom/canvasrenderingcontext2d/textrendering)

The feature, as displayed in After Effects
AfterFX_lfTZ0Q2C33

exdeejay:
Adds an `antialiased` signal to all Shapes that enables/disables antialiasing for the shape. Confirmed working on Firefox, not tested across other browsers. Fixes #235.

### DIFF
--- a/packages/2d/src/components/Shape.ts
+++ b/packages/2d/src/components/Shape.ts
@@ -24,6 +24,7 @@ export interface ShapeProps extends LayoutProps {
   lineCap?: SignalValue<CanvasLineCap>;
   lineDash?: SignalValue<number>[];
   lineDashOffset?: SignalValue<number>;
+  antialiased?: SignalValue<boolean>;
 }
 
 export abstract class Shape extends Layout {
@@ -49,6 +50,9 @@ export abstract class Shape extends Layout {
   @initial(0)
   @signal()
   public declare readonly lineDashOffset: SimpleSignal<number, this>;
+  @initial(true)
+  @signal()
+  public declare readonly antialiased: SimpleSignal<boolean, this>;
 
   protected readonly rippleStrength = createSignal<number, this>(0);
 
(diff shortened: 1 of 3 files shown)


aarthificial:
I'll have to add this to the Contribution Guide, but Chrome is the main browser we support, you can't just *not* test it.

exdeejay:
Fair enough. Tested just now and it works on Chrome.

packages/2d/src/components/Shape.ts
```diff
@@ -69,6 +73,12 @@ export abstract class Shape extends Layout {
     context.lineCap = this.lineCap();
     context.setLineDash(this.lineDash());
     context.lineDashOffset = this.lineDashOffset();
+    if (!this.antialiased()) {
+      // from https://stackoverflow.com/a/68372384
+      context.imageSmoothingEnabled = false;
```
aarthificial, line 78: I think we should merge `Shape.antialiased` and `Image.smoothing` together. They both control the `imageSmoothingEnabled` option and `Image` extends `Rect` which extends `Shape` so it will now have two duplicated properties.

aarthificial (changes requested):

aarthificial (changes requested):
The name should stay as `smoothing` though, otherwise this is a breaking change.

exdeejay:
As far as I can tell, the `imageSmoothingEnabled` in the antialiased handling code is unnecessary, as the context gets reset before drawing any images. I've reverted the previous change so that `antialiased` and `smoothing` are back to doing their own separate things. If you think they should do the same thing, we can make `smoothing` an alias of `antialiased`.

PR #287, opened 2023-02-10
fix(2d): make Text respect textWrap="pre"

### LINKED ISSUE #284: Text never wrapping on Firefox.
**Describe the bug**
Text just does not wrap. It always renders in one line, no matter if you use preformatted text or try wrapping it automatically.
(Chrome on the left, Firefox on the right)
image

**To Reproduce**
```ts
function* problem(scene: View2D) {
  scene.add(
    <Text
      fill="white"
      fontFamily="monospace"
      textWrap="pre"
      text={`hello {
  world
}`}
    />
  );
}
```

**Expected behavior**
The text should wrap.

**Package versions:**
- core: [e.g. 2.2.0]
- ui: [e.g. 2.2.0]

**Additional context**
This only happens on Firefox.

exdeejay:
Fixes #284. `context.fillText()` is not supposed to support newlines in text, so I'm not sure how it worked on Chrome before. `pre` text wrapping is preserved in the same way as text wrapping is done with the segmenter: one text node is added per line of text.

### DIFF
--- a/packages/2d/src/components/Text.ts
+++ b/packages/2d/src/components/Text.ts
@@ -78,7 +78,7 @@ export class Text extends Shape {
     rect: Rect,
   ) {
     const y = rect.y + rect.height / 2;
-    const oldBaseline = context.textBaseline;
+    context.save();
     context.textBaseline = 'middle';
 
     if (this.lineWidth() <= 0) {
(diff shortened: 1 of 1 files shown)


aarthificial:
> I'm not sure how it worked on Chrome before

Chrome used the segmenter for both types of wrapping.

aarthificial (approved):

PR #227, opened 2023-02-06
docs: add logging page

Tandashi:
**What has changed?**
A new page has been added to the documentation describing the process on how to access and use the built-in `Logger`.

Closes: #199

PR #293, opened 2023-02-10
feat(core): add `debug` helper function

### LINKED ISSUE #267: Allow logging of non-string values
**Description**

Currently, the logger only accepts a parameter of type `string | LogPayload`. I personally use the logger mostly when trying to debug a wonky animation. In these cases, I will almost always want to log complex types like `Vector2` or `Rect`s. What I need to do in these cases is to manually `JSON.stringify` the payload before passing it to the logger. 

```ts
const logger = useLogger();
logger.debug(JSON.stringify(rect().cacheRect());
``` 

While this is not a huge deal, it's starting to become death by a thousand paper cuts.

**Proposed solution**

Allow the logger to also accept non-string values and call `JSON.stringify` on them internally. 

```ts
const logger = useLogger();

// Would call `JSON.stringify` on the input internally
logger.debug(rect().cacheRect());
```

The big problem here is what the new method signature of these methods should become. Since we would essentially be widening the type to `any`, we'd lose the ability to distinguish between `LogPayload`s and any other payload.

ksassnowski:
## Summary

Adds a new `debug` helper function to easily log debug messages with arbitrary payloads to the UI. See #267 for the rationale.

The function uses the logger's built-in functionality to pretty-print complex types by providing the `object` property in the log payload. Since it takes an additional mouse click to open the object dropdown in the UI, the `message` will be the JSON serialized payload itself. This way, you would still get a useful output at a glance which can often be enough when logging simpler types like `Vector2`.

```typescript
import {debug} from "@motion-canvas/core/lib/utils";

const v = Vector2.one;

debug(v);

// Is equivalent to
useLogger().debug({ message: JSON.serialize(v), object: v });
```

When logging primitive values like `string` and `number`,  providing the `object` property would only clutter up the output without adding any useful information. In these cases, the payload will simply be converted to a string using `JSON.stringify` while taking care of various edge-cases like `undefined`, `NaN` and not double quoting `string`s.

```ts
debug("Hello world");
// Is equivalent to
useLogger().debug("Hello world");

debug(500);
// Is equivalent to
useLogger().debug(JSON.stringify(500));

debug(undefined)
// Is equivalent to
useLogger().debug("undefined");
// since JSON.stringify(undefined) would return `null` which would lose information
```

I have included a test suite for this new function as well that should outline the important cases.

## Documentation

I've added a doc block to the function but I'm not sure what I need to do for it to show up in the API reference of the documentation. Just rebuilding the project locally didn't seem to do anything.

Closes #267

### DIFF
--- a/packages/core/src/utils/debug.test.ts
+++ b/packages/core/src/utils/debug.test.ts
@@ -1,0 +1,51 @@
+import {describe, test, beforeEach, expect, vi} from 'vitest';
+import {setProject, useLogger, debug} from '../utils';
+import {Project, ProjectMetadata} from '../Project';
+import {Meta} from '../Meta';
+import {LogLevel, LogPayload} from '../Logger';
+import {Rect, Vector2} from '../types';
+
+describe('debug()', () => {
+  beforeEach(() => {
+    setProject(new Project({name: 'test', scenes: []}));
+  });
+
+  test.each([
+    ['Strings', 'test', {message: 'test'}],
+    ['Numbers', 5, {message: '5'}],
+    ['Null', null, {message: 'null'}],
+    ['Undefined', undefined, {message: 'undefined'}],
+    ['NaN', NaN, {message: 'NaN'}],
+    ['Vector2', Vector2.one, {message: '{"x":1,"y":1}', object: Vector2.one}],
+    [
+      'Rect',
+      new Rect([10, 20, 30, 40]),
+      {
+        message: '{"x":10,"y":20,"width":30,"height":40}',
+        object: new Rect([10, 20, 30, 40]),
+      },
+    ],
+    [
+      'Arrays',
+      [5, 'test', Vector2.one],
+      {message: '[5,"test",{"x":1,"y":1}]', object: [5, 'test', Vector2.one]},
+    ],
+    ['Empty Array', [], {message: '[]', object: []}],
+    [
+      'Objects',
+      {foo: 'bar', baz: 'qux'},
+      {message: '{"foo":"bar","baz":"qux"}', object: {foo: 'bar', baz: 'qux'}},
+    ],
+    ['Empty Objects', {}, {message: '{}', object: {}}],
+  ])('log: %s', (_, payload, expected) => {
+    const spy = vi.fn();
+    useLogger().onLogged.subscribe(spy);
+
+    debug(payload);
+
+    expect(spy).toHaveBeenNthCalledWith(1, {
+      level: LogLevel.Debug,
+      ...expected,
+    });
+  });
+});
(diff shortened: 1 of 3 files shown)


Tandashi:
When this is merged I can also add a short paragraph to the logging documentation page (#227).

Regarding the API Reference not showing up make sure the remove the `generated` folder. If it exists the `api.json` (which is used for the API Section in the docs) will not be regenerated. Probably something we should change in the near future so one doesn't have to do it manually.

packages/core/src/utils/debug.test.ts
```diff
@@ -0,0 +1,57 @@
...
+          level: LogLevel.Debug,
+          ...expected,
+        });
+        resolve();
+      });
+    });
+
+    debug(payload);
+
+    return promise;
```
aarthificial, line 55: Logging is synchronous, so I think we could just use a mock function:
```suggestion
    const spy = vi.fn();
    useLogger().onLogged.subscribe(spy);

    debug(payload);

    expect(spy).toHaveBeenCalledWith({
      level: LogLevel.Debug,
      ...expected,
    });
```

packages/core/src/utils/debug.test.ts
```diff
@@ -0,0 +1,57 @@
...
+import {setProject, useLogger, debug} from '../utils';
+import {Project, ProjectMetadata} from '../Project';
+import {Meta} from '../Meta';
+import {LogLevel, LogPayload} from '../Logger';
+import {Rect, Vector2} from '../types';
+
+describe('debug()', () => {
+  beforeEach(() => {
+    setProject(
+      new Project('test', new Meta<ProjectMetadata>('test'), {scenes: []}),
```
aarthificial, line 11: This is a deprecated signature, we should pass a config object instead:
```suggestion
      setProject(new Project({name: 'test', scenes: []}));
```
Also, metadata is not necessary.

aarthificial (changes requested):

packages/core/src/utils/debug.test.ts
ksassnowski: Oh, weird. VSCode was telling me that _that's_ the deprecated signature. Also the deprecation notice in `Project.ts` seems to agree with that, too: https://github.com/motion-canvas/motion-canvas/blob/main/packages/core/src/Project.ts#L190

packages/core/src/utils/debug.test.ts
ksassnowski: Neat. I didn't realize vitest hat built-in mocks. Will change.

packages/core/src/utils/debug.test.ts
aarthificial: Right, I most likely forgot that each signature should have its own typedoc and thought I was deprecating all of them.
But the one with the config object is the correct one.

aarthificial (approved):

aarthificial:
Awesome, thanks for the PR!

PR #295, opened 2023-02-10
docs: Component documentation

### LINKED ISSUE #202: Documentation for components
Components are undocumented. I will most likely write docs for them soon :D

I plan to use two examples, one of a simple text component with a colour and text and another with an attached function based of of the original component which will animate it

nick-1666:
Added a page under the "advanced" tab documenting components. Heavily based off of the random page.

Closes #202


nick-1666:
also `sputnick#1666` for green role

packages/examples/src/scenes/components-switch.tsx
```diff
@@ -0,0 +1,85 @@
...
+  public declare readonly accent: ColorSignal<this>;
+
+  private isOn: boolean;
+  private readonly indicatorPosition = createSignal(0);
+  private readonly offColor = new Color('#242424');
+  private readonly indicator = createRef<Circle>();
+  private readonly container = createRef<Rect>();
+
+  public constructor(props?: SwitchProps) {
+    super({
```
Tandashi, line 38: Super minor detail: Is there a reason why you create a new object and spread the `props` instead of just passing the props directly?

```ts
super(props);
```

packages/docs/docs/advanced/custom-components.mdx
```diff
@@ -0,0 +1,126 @@
...
+    @initial(false)
+    // @signal - is required by motion canvas
+    // for every prop that was passed in.
+    @signal()
+    public declare readonly initialState: SimpleSignal<boolean, this>
+
+    @initial("#68ABDF")
+    @signal()
+    public declare readonly accent: ColorSignal<this>
+```
```
Tandashi, line 66: Yet another super minor detail: Personally I am missing a `}`. Or 
```ts
// ...
}
```
Since you move on to a different section in the next code snipped.

packages/examples/src/scenes/components-switch.tsx
ksassnowski: This and the the empty `<>` when adding only a single root element also apply to a few more places in this file.

nick-1666: > Super minor detail: Is there a reason why you create a new object and spread the `props` instead of just passing the props directly?
> 
> ```ts
> super(props);
> ```

because if you were to make a layout component and wanted to ensure it was a layout, you would use 
```tsx
super({
   layout
   ...props
})
```

packages/examples/src/scenes/components-switch.tsx
```diff
@@ -0,0 +1,85 @@
...
+      ...props,
+    });
+
+    this.indicatorPosition(this.initialState ? 50 : -50);
+    this.isOn = this.initialState();
+
+    this.add(
+      <>
```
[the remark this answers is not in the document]
nick-1666: because if you were to make a layout component and wanted to ensure it was a layout, you would use :
```tsx
super({
   layout
   ...props
})
```
just makes it a bit more obvious not only props can go here

packages/examples/src/scenes/components-switch.tsx
ksassnowski: I'm not sure I'm following. Where is the `layout` variable coming from in this example?

packages/examples/src/scenes/components-switch.tsx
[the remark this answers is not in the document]
nick-1666: this is true, it was 11 and it slipped past me lol. will be fixed

packages/examples/src/scenes/components-switch.tsx
nick-1666: because if you were to make a layout component and wanted to ensure it was a layout, you would use :
```tsx
super({
   layout
   ...props
})
```
just makes it a bit more obvious not only props can go here

packages/docs/docs/advanced/custom-components.mdx
nick-1666: okay. will be in the next commit in the name of consistency if its like this everywhere else

packages/examples/src/scenes/components-switch.tsx
ksassnowski: If you were making a layout component, you'd also be extending from `Layout` and not `Node`, however. So the props in this case would also extends `LayoutProps` and not `NodeProps`. At which point `super(props)` would still work as expected.

Tandashi: Yeah no I get that but was confused because in that concrete example it is not needed. But if you want to add it just to provide a generic way to make it easier understandable for someone that isn't so much into typescript / coding I am all for it. But providing an example / paragraph describing this could be useful then as well.

Not sure however at what persona the documentation is targeted. Maybe @aarthificial could shed some light on this.

nick-1666:
@Tandashi fixed in new commits. Review?

packages/docs/docs/advanced/custom-components.mdx
```diff
@@ -0,0 +1,126 @@
...
+import AnimationPlayer from '@site/src/components/AnimationPlayer';
+import CodeBlock from '@theme/CodeBlock';
+import source from '!!raw-loader!@motion-canvas/examples/src/scenes/components-switch';
+
+# Components
+
+<AnimationPlayer name="components" banner />
+
+Components are reusable parts of code such as `Rect` and `Circle` which can be
+reused throughout scenes. To use a component in a scene, add it to the view and
```
hhenrichsen, line 14: nit: reusable/reused is repeated

hhenrichsen:
As per the contributing guide, [any commits addressing code reviews should be fixup commits](https://github.com/motion-canvas/motion-canvas/blob/main/CONTRIBUTING.md#addressing-review-feedback). This helps us to keep git history clean, and tell a consistent story.

packages/docs/docs/advanced/custom-components.mdx
```diff
@@ -0,0 +1,126 @@
...
+export interface SwitchProps extends NodeProps {
+    initialState?: SignalValue<boolean>
+
+    // We dont use color here because we want
+    // to be able to pass hex strings and rgb
+    // values to accent rather than a `Color`
+    accent?: SignalValue<PossibleColor>
+}
+```
+Next, create a component class for your components. It must extend an existing
```
hhenrichsen, line 39: I would start this section with the requirements (i.e., "must extend Node or one of its subclasses"), and then work towards the recommendation to pick the most closely associated subclass. This sets up a good flow from hard rules to recommendations.

packages/docs/docs/advanced/custom-components.mdx
```diff
@@ -0,0 +1,126 @@
...
+to use a component most similar to yours, for instance if you were to make a component
+including a `Layout`, you should extend `Layout` and `LayoutProps`.
+```tsx
+export interface SwitchProps extends NodeProps { ... }
+
+export class Switch extends Node { ... }
+```
+
+To use the properties defined in the interface, your class _must_ contain a property with the
+same name. Motion Canvas provides type decorators to facilitate this. Here is an example of
```
hhenrichsen, line 51: I would reword the middle portion of this to read "Motion Canvas provides type decorators to facilitate this, like `@signal()` and `@initial()`. For more info on Signals, look here." with a link to the signals page. You may also want to explicitly state that `@initial()` is optional below.

packages/docs/docs/advanced/custom-components.mdx
```diff
@@ -0,0 +1,126 @@
...
+public constructor(props?: SwitchProps) {
+    super({
+        ...props
+    })
+    ...
+}
+```
+Now the properties will follow the schema defined in your interface and they will be
+passed to the superclass, in our case `Node`. The `props` parameter can also be useful
+outside of the super call for instance, setting styles on every rect in an array.
```
hhenrichsen, line 99: nit: This may read better as 
> The `props` parameter can also be useful outside of the super call to access your data elsewhere. For example, if you wanted to set a color set in the `props` on every `Rect` in an array.

"For instance" is normally a suffix.

packages/docs/docs/advanced/custom-components.mdx
```diff
@@ -0,0 +1,126 @@
...
+            <Rect
+                ...
+            >
+                <Circle ... />
+            </Rect>
+        </>
+    )
+}
+```
+Since this is still a regular typescript class, you can add methods. This is especially useful
```
hhenrichsen, line 117: nit: "Since this is still a regular typescript class, you can also add methods."

"Since this is still..." normally implies some base level of information, so "also" makes sense here.

hhenrichsen (commented):
Not going to approve or deny since most of what I'm leaving is grammatical over content, but here are some of my thoughts on how we can make this a little cleaner.

packages/docs/docs/advanced/custom-components.mdx
```diff
@@ -0,0 +1,135 @@
...
+passed to the superclass, in our case `Node`. The `props` parameter can also be useful
+outside of the super call for instance, setting styles on every rect in an array.
+
+Now we can add elements to the view by using `this.add()`, much like you would add to a
+scene's view:
+```tsx
+public constructor(props?: SwitchProps) {
+    ...
+    this.add(
+        <>
```
Tandashi, line 107: nit: empty element `<>` which isn't really needed.

Tandashi:
Also a person opinion where others also are free to voice their opinion.

I **personally** like when things are linked. You used a lot of classes like `Layout`, `Node`, `Circle`, `Rect`, ...
I think it would be a good idea to make them clickable and link them to the respective API Reference.
If I am not mistaken that is how other documentation pages handle this as well.

packages/docs/docs/advanced/custom-components.mdx
nick-1666: idk what i was thinking 💀

nick-1666:
> As per the contributing guide, [any commits addressing code reviews should be fixup commits](https://github.com/motion-canvas/motion-canvas/blob/main/CONTRIBUTING.md#addressing-review-feedback). This helps us to keep git history clean, and tell a consistent story.

fixed!

hhenrichsen:
It looks like something went really wrong in the process because there are now 13 commits on this PR instead of the 2 or 3 there should be.

Tandashi:
> > As per the contributing guide, [any commits addressing code reviews should be fixup commits](https://github.com/motion-canvas/motion-canvas/blob/main/CONTRIBUTING.md#addressing-review-feedback). This helps us to keep git history clean, and tell a consistent story.

> fixed!

Ehm... Shouldn't be 13 commits now. You probably merged the `main` into your branch I assume.

packages/docs/docs/advanced/custom-components.mdx
nick-1666: Fixed this code sample in the new commit and changed the formatting to improve clarity

nick-1666:
uh oh

hhenrichsen:
Also please go read the contributing guide. `fixup` commits are not the same as `fix:`-scoped commits. Mention me on the discord if you want me to walk you through how fixup commits work.

nick-1666:
well uhh i didn't mean to spam request reviews. It was a visual glitch :/
my bad

packages/docs/docs/advanced/custom-components.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+import AnimationPlayer from '@site/src/components/AnimationPlayer';
+import CodeBlock from '@theme/CodeBlock';
+import source from '!!raw-loader!@motion-canvas/examples/src/scenes/components-switch';
+
+# Components
+
+<AnimationPlayer name="components" banner />
+
+Components are reusable parts of code such as `Rect` and `Circle` which can be
```
hhenrichsen, line 13: > Components are reusable parts of code such as `Rect` and `Circle` which can be scenes.

Looks like something got lost here 😅

hhenrichsen, line 13: Additionally: did we want these to link to the API pages for `Rect` and `Circle`?

packages/docs/docs/advanced/custom-components.mdx
```diff
@@ -0,0 +1,141 @@
...
+ scenes. To use a component in a scene, add it to the view and
+provide arguments to the component.
+
+```tsx
+<Switch
+    initialState={false}
+/>
+```
+To define what arguments a component will take in, first define an interface. All
```
hhenrichsen, line 22: super nit: "in" is redundant here

packages/docs/docs/advanced/custom-components.mdx
```diff
@@ -0,0 +1,141 @@
...
+export interface SwitchProps extends NodeProps {
+    initialState?: SignalValue<boolean>
+
+    // We dont use color here because we want
+    // to be able to pass hex strings and rgb
+    // values to accent rather than a `Color`
+    accent?: SignalValue<PossibleColor>
+}
+```
+Next, create a component class for your components. The component class must extend [Node](/api/2d/components/Node)
```
hhenrichsen, line 40: nit: "Next, create a ~~component~~ class for your components."

packages/docs/docs/advanced/custom-components.mdx
```diff
@@ -0,0 +1,141 @@
...
+or one of its subclasses. If you dont want to inherit any methods from an existing component, extend your class from
+`Node`. It is advisable to extend from a component most similar to the custom component you are building. For instance,
```
hhenrichsen, line 42: We should probably keep with the plural first person here: "We advise extending from ~~a~~ the component most similar to the ~~custom~~ component you are building"

packages/docs/docs/advanced/custom-components.mdx
```diff
@@ -0,0 +1,141 @@
...
+To use the properties defined in the interface, your class _must_ contain a property with the
+same name. Motion Canvas provides type decorators to facilitate this like `@initial()` and
+`@signal()`. Click [here](/docs/signals) for more information on signals.
+
+Here is an example of how you would define such properties:
+```tsx
+export class Switch extends Node {
+    // @initial - sets the property to an
+    // initial value if it was not provided.
+    // The @initial decorator is optional.
```
hhenrichsen, line 60: Since this is on a comment on the `@initial` decorator, it's probably fine to say "This decorator is optional," or even update the beginning of this comment to "@initial - optional, sets the property to an..."

packages/docs/docs/advanced/custom-components.mdx
```diff
@@ -0,0 +1,141 @@
...
+                ...
+            >
+                <Circle 
+                    ... 
+                />
+            </Rect>
+    )
+}
+```
+Since this is also a regular typescript class, you can add methods. This is especially useful
```
hhenrichsen, line 123: Did you decide not to use "also" here?

hhenrichsen (commented):
Again some more grammar things I found, but not going to block on any of them.

packages/docs/docs/advanced/custom-components.mdx
nick-1666: Bro i swear i try not to be superficial but they just keep getting past me lol

packages/docs/docs/advanced/custom-components.mdx
nick-1666: youre right

packages/docs/docs/advanced/custom-components.mdx
nick-1666: okay true

packages/docs/docs/advanced/custom-components.mdx
nick-1666: ill go with the latter

packages/docs/docs/advanced/custom-components.mdx
nick-1666: Since this is still... was transformed into Since this is also. is that not what you meant?

hhenrichsen: I was intending "Since this is a regular typescript class, you can also add methods."

nick-1666: should typescript be omitted since its implied that typescript is used throughout the project?

packages/docs/docs/advanced/custom-components.mdx
nick-1666: done!

packages/docs/docs/advanced/custom-components.mdx
nick-1666: yes we do

packages/docs/docs/advanced/custom-components.mdx
hhenrichsen: I think typescript is a good distinction because we're talking about component classes as well, so having a distinction between a "regular TypeScript class" and the rules for a component class is good. We might capitalize TypeScript, though.

nick-1666:
fixed-up

hhenrichsen (commented):
Couple extra thoughts but no blockers.

packages/docs/docs/advanced/custom-components.mdx
```diff
@@ -0,0 +1,140 @@
...
+    public declare readonly accent: ColorSignal<this>
+
+    ...
+
+}
+```
+Notice how colors are wrapped in `ColorSignal<>` whilst any other type (even user-defined ones)
+are wrapped in `SimpleSignal<>`. The type does not need to be passed to color signal as Motion
+Canvas knows that it must be of a color-resolvable type. In both, the class is passed at the end
+of the wrapper to register the signal to the class. The properties must also be _declared_, _public_
```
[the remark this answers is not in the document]
nick-1666: no cause its not an actual key word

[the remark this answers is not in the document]
nick-1666: i dont think so. If we wanted to use keywords the sentence could be changed to:
"The properties must be (initialised/initialized depending on uk or us) with the `public` `declare` and `readonly` keywords"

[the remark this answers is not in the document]
nick-1666: or better yet just "Properties ..."

nick-1666:
this should have fixed the formatting

Issue #294, opened 2023-02-10
Make building API Reference locally less confusing


Tandashi:
**Description**
It would be nice if one doesn't need to manually remove the `generated` folder in order for the `api.json` (and other things).
Having to do it manually / not knowing about this can easily cause confusion why things are not showing up in the documentation when building locally.

**Proposal**
Add a remove for the `generated` directory in the `package.json` build script.

---

(Would love to claim this if this should get accepted)

aarthificial:
It may be possible to detect if we're in build mode directly from our plugin: https://github.com/facebook/docusaurus/discussions/7621

Tandashi:
Having it directly in the plugin seems like the better option. Will do it that way if possible.

PR #298, opened 2023-02-10
feat(docs): always re-build api references in `build` mode

Tandashi:
## What happens in this Pull Request

In this Pull Request the `typedoc` plugin has been changed to always re-build generated files when
running in `build` mode.

Before the API References did not get re-build when running `npm run docs:build` or using `npx lerna run build` which caused some confusion.

### Explanation
One can check if `docusaurus` runs in `build` mode using the `NODE_ENV` environment variable, which will be set to `production` if that's the case. If the variable is set, we ignore previously build files and re-build them, including the `api.json`.

---

Closes: #294


aarthificial:
Awesome, thanks!

PR #299, opened 2023-02-10
Fix (UI) : Play-Pause with Spacebar , Fixes #273

### LINKED ISSUE #273: Video doesn't pause in the player, when pressing spacebar while the play/pause button is :active
If I have already clicked the pause button and then use the spacebar instead, it neutralizes the pause, since it gets triggered twice.

miikkuu:
Fixed play-pause focus issue with spacebar when the element is already selected with mouse.
Added preventDefault funcion to the case.
Solves #273

### DIFF
--- a/packages/ui/src/components/playback/PlaybackControls.tsx
+++ b/packages/ui/src/components/playback/PlaybackControls.tsx
@@ -32,6 +32,7 @@ export function PlaybackControls() {
         }
         switch (event.key) {
           case ' ':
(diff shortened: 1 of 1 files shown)


aarthificial:
Nice, thanks for the PR!
Small note: next time make sure to follow the [Angular Commit Message Conventions](https://github.com/angular/angular/blob/main/CONTRIBUTING.md#commit) as described in our contribution guide.

Issue #11, opened 2022-07-19
Add code style docs, expand lint, reconfigure prettier.


Ross-Esmond:
**Description**
I'll have trouble remembering the code styles unless I take notes. This issue should allow us to keep track of the code styles that come up.

I would that whenever possible the code styles should go first into Prettier, as it has an autofix, then into lint, as it has autodetection, and then go into a `CODE_STYLE.md` (working name), as a fallback.

**Proposed code styles so far**

- [ ] Capitalize Interface names. (This can be set up in prettier)
- [ ] Use concise sentences in the imperative present tense for function JSDocs. "Run the function." "Do the thing." "Clear the items."
- [x] Add eslint-plugin-tsdoc

For Jacob (the original dev) if you think of any more comment about them and I'll add them. If you are contributing this can operate as a work-in-progress code style guide, subject to change.

PR #306, opened 2023-02-10
feat(2d): add Icon Component (#305)

### LINKED ISSUE #305: feat(2d): Add Icon Component
**Description**

This feature will add the `<Icon />` component. It acts as a wrapper around the `<Image/>` and provides a simple API
for including SVG Icons in Motion Canvas.

**Proposed solution**

An `<Icon/>` component that makes use of the iconify API to request SVG Icons.
One can declare an Icon by using one of the 150k+ identifiers. One can also set the color of the icon using a Prop.
```tsx
view.add(
  <Layout layout>
    <Icon icon={"mdi:abjad-arabic"} size={200} color="red" />
    <Icon icon={"mdi:typewriter"} size={200} color="#bada55" />
    <Icon icon={"mdi:abugida-thai"} size={200} color="#bada55aa" />
    <Icon icon={"mdi:access-point"} size={200} color="green" />
    <Icon icon={"mdi:access-point"} size={200} color="blue" />
  </Layout>
)
```
A list of all Icons that would be accessible using this approach can be found on [Icônes](https://icones.js.org/).

The component converts the icon identifier (eg `mdi:access-point` and the provided color (eg `red`) into a URL that would look like this: `https://api.iconify.design/mdi/access-point.svg?color=red`

The Icon would be implemented as a subclass of Image. Image simply consumes the generated URL as a `src` of the image.

**Considered alternatives**
no alternatives have been considered.

**Additional context**
This feature has been discussed on Discord in the following Forum Topic: https://discord.com/channels/1071029581009657896/1073653313335218177

WaldemarLehner:
This introduces the new `<Icon/>` component, a subclass of `Image`.
Closes #305 

Blocked by #357

### DIFF
--- a/packages/2d/src/components/Icon.ts
+++ b/packages/2d/src/components/Icon.ts
@@ -1,0 +1,85 @@
+import {Image, ImageProps} from '../components/Image';
+import {colorSignal, computed, initial, signal} from '../decorators';
+import {SignalValue, SimpleSignal} from '@motion-canvas/core/lib/signals';
+import {useLogger} from '@motion-canvas/core/lib/utils';
+import {ColorSignal, PossibleColor} from '@motion-canvas/core/lib/types';
+
+export interface IconProps extends ImageProps {
+  /**
+   * {@inheritDoc Icon.icon}
+   */
+  icon: SignalValue<string>;
+
+  /**
+   * {@inheritDoc Icon.color}
+   */
+  color?: SignalValue<PossibleColor>;
+}
+
+/**
+ * An Icon Component that provides an easy access to over 150k icons.
+ * See https://icones.js.org/collection/all for all available Icons.
+ */
+export class Icon extends Image {
+  /**
+   * The identifier of the icon.
+   *
+   * @remarks
+   * You can find identifiers on [Icônes](https://icones.js.org).
+   * They can look like this:
+   * * `mdi:language-typescript`
+   * * `ph:anchor-simple-bold`
+   * * `ph:activity-bold`
+   */
+  @signal()
+  public declare icon: SimpleSignal<string, this>;
+
+  /**
+   * The color of the icon
+   *
+   * @remarks
+   * Provide the color in one of the following formats:
+   * * named color like `red`, `darkgray`, …
+   * * hexadecimal string with # like `#bada55`, `#141414`
+   *   Value can be either RGB or RGBA: `#bada55`, `#bada55aa` (latter is partially transparent)
+   *   The shorthand version (e.g. `#abc` for `#aabbcc` is also possible.)
+   *
+   * @default 'white'
+   */
+  @initial('white')
+  @colorSignal()
+  public declare color: ColorSignal<this>;
+
+  public constructor(props: IconProps) {
+    super(props);
+  }
+
+  /**
+   * Create the URL that will be used as the Image source
+   * @returns Address to Iconify API for the requested Icon.
+   */
+  @computed()
+  protected svgUrl(): string {
+    const iconPathSegment = this.icon().replace(':', '/');
+    const encodedColorValue = encodeURIComponent(this.color().hex());
+    // Iconify API is documented here: https://docs.iconify.design/api/svg.html#color
+    return `https://api.iconify.design/${iconPathSegment}.svg?color=${encodedColorValue}`;
+  }
+
+  /**
+   * overrides `Image.src` getter
+   */
+  protected getSrc(): string {
+    return this.svgUrl();
+  }
+
+  /**
+   * overrides `Image.src` setter to warn the user that the value
+   * is not used
+   */
+  protected setSrc() {
+    useLogger().warn(
+      "The Icon Component does not accept setting the `src`. If you need access to `src`, use '<Image/>` instead.",
+    );
+  }
+}
--- a/packages/2d/src/components/index.ts
+++ b/packages/2d/src/components/index.ts
@@ -1,4 +1,5 @@
 export * from './Circle';
+export * from './Icon';
 export * from './Image';
 export * from './Grid';
 export * from './Latex';


packages/2d/src/components/Icon.ts
```diff
@@ -0,0 +1,130 @@
...
+      if (str.length !== 6 && str.length !== 8) {
+        return false;
+      }
+      // If we do not have a hexadecimal string here, parseInt will return Not a Number (NaN)
+      // so we have a valid hexadecimal string if this returns false
+      return !Number.isNaN(parseInt(str, 16));
+    }
+
+    const colorName = canBeInterpretedAsHexadecimal(withoutHashTag)
+      ? '%23' + withoutHashTag // as hexadecimal value
```
hhenrichsen, line 104: You might replace `%23` with `encodeURIComponent('#')` just so it's clear where this string is coming from.

packages/2d/src/components/Icon.ts
```diff
@@ -0,0 +1,130 @@
...
+      : withoutHashTag; // as named color
+
+    return colorName;
+  }
+
+  protected override image(): HTMLImageElement {
+    const src = this.svgUrl();
+
+    if (!Icon.iconContentsPool[src]) {
```
hhenrichsen, line 113: I would replace this with the guard pattern; something like this:
```tsx
if (Icon.iconContentsPool[src]) {
  return Icon.iconContentsPool[src];
}

const image = document.createElement('img');
image.src = src;
if (!image.complete) {
  DependencyContext.collectPromise(
    new Promise((resolve, reject) => {
        image.addEventListener('load', resolve); // also be careful of comma's undefined behavior
        image.addEventListener('error', reject);
    }),
  );
}
Icon.iconContentsPool[src] = image;
return image;
```
This way the cognitive load of reading this is a little bit lower; if we hit a case where we can early return, we early return.

hhenrichsen (approved):
Couple thoughts, but generally looks good to me.

packages/2d/src/components/Icon.ts
```diff
@@ -0,0 +1,130 @@
...
+      this.color(),
+    )}`;
+    return url;
+  }
+
+  private static normalizeColorInput(colorString: string) {
+    // This assumes that no HTML named color is a valid hexadecimal string.
+    // First, check if a # is present. If yes, we can assume that this is not a
+    // named color.
+    const hasHashtag = colorString.trimStart().startsWith('#');
```
hhenrichsen, line 87: super nit: This can just be `hasHash` -- the tag part isn't relevant in code that doesn't actually do any tagging 😛

packages/2d/src/components/Icon.ts
```diff
@@ -0,0 +1,130 @@
...
+    const iconPathSegment = this.icon().replace(':', '/');
+
+    // iconify API is documented here: https://docs.iconify.design/api/svg.html#color
+    const url = `https://api.iconify.design/${iconPathSegment}.svg?color=${Icon.normalizeColorInput(
+      this.color(),
+    )}`;
+    return url;
+  }
+
+  private static normalizeColorInput(colorString: string) {
```
hhenrichsen, line 83: Looking at this again, we probably can condense this into two steps:
1. Check if there's a hash, and if there is make sure that it's a valid hex number of 6 or 8 digits.
    * If there's not, warn the user that they used an invalid hex using the logger and don't bother updating the value.
2. Send the result to the `encodeURIComponent` function. If there's a `#`, it'll get converted. If not, it stays as the input string.

Tandashi: Also just found a simpler way to check if its a valid color.
Assuming that every CSS color is valid you could use:

```ts
function isColor(color: string){
  const style = new Option().style;
  style .color = color;
  return style .color === color;
}
```

Tandashi:
Please use Git `fixup` commits to make review changes as described in the [Contribution Guide])https://github.com/motion-canvas/motion-canvas/blob/main/CONTRIBUTING.md#addressing-review-feedback).
Just helps us of having a cleaner commit history

packages/2d/src/components/Icon.ts
WaldemarLehner: >  warn the user that they...

How would I access the console inside the UI? printing to `console.error` does not seem to print the Error there.

Tandashi: You can use the built-in logger as follows:
```ts
useLogger().error('Some error here');
```
There is also a PR (#227) on the way for further references.

WaldemarLehner: > Assuming that every CSS color is valid you could use:

Just checked, and yup :) The API just passes through whatever you provide it as the color to the SVG Path's `fill` property. 

e.g. `https://api.iconify.design/mdi/access-point.svg?color=testtest` 
sets the fill to testtest
image

looks like the most elegant solution in that case :+1:

hhenrichsen:
> Please use Git `fixup` commits rather than making a new commit to make review changes as described in the [Contribution Guide](https://github.com/motion-canvas/motion-canvas/blob/main/CONTRIBUTING.md#addressing-review-feedback). Just helps us of having a cleaner commit history

If you need help doing this, I'm happy to hop on a call in a little while, or write out a description on Discord (I'm Hunter on there).

WaldemarLehner:
just saw the comment about fixup. Will do :)

hhenrichsen:
For the commits you've already pushed, you can do `git rebase --interactive upstream/main` (if `upstream` is your remote for the `motion-canvas/motion-canvas` repo) and change the operation to `fixup` (they should default to `pick`) on any commits that should be merged back into the original commit, like the two new `fix`-scoped commits.

WaldemarLehner:
Done. Force-Pushed with fixup commits. Did I do it right? :smiley:

packages/2d/src/components/Icon.ts
WaldemarLehner: > Should work with Hex Colors

On Firefox it returns the color in the rgb() syntax unfortunately. 
image

hhenrichsen:
> Done. Force-Pushed with fixup commits. Did I do it right? :smiley: 

It looks like you rebased them away which is a good step for the old ones. Further commits should be done with fixup commits:
```
git commit --fixup=[commit id to fix here]
```

Then when it comes time to merge, you can do
```
git rebase --interactive --autosquash upstream/main
```
Which will do the process I just had you do for you!

packages/2d/src/components/Icon.ts
```diff
@@ -0,0 +1,152 @@
...
+   * Color of the Icon. Defaults to white.
+   * Provide the color in one of the following formats:
+   * * named color like `red`, `darkgray`, …
+   * * hexadecimal string, with or without # `#bada55`,`141414`
+   *
+   * Values can be either RGB or RGBA: `#bada55` or `#bada55aa` (latter is partially transparent)
+   */
+  @initial('white')
+  @signal()
+  public declare color: SimpleSignal<string, this>;
```
aarthificial, line 63: Any reason we're not using the `Color` signal?
```suggestion
  @initial('white')
  @colorSignal()
  public declare color: ColorSignal<this>;
```

This automates the conversion from all string color formats to a common `Color` class provided by `chroma-js`

packages/2d/src/components/Icon.ts
aarthificial: With a color signal, you could just do `this.color().hex()` to retrieve the hex value.

packages/2d/src/components/Icon.ts
```diff
@@ -0,0 +1,152 @@
...
+  public constructor(props: IconProps) {
+    super(props);
+  }
+
+  /**
+   * This creates the URL that will be used as the Image source
+   * @returns Address to Iconify API for the requested Icon.
+   */
+  @computed()
+  protected svgUrl(): string {
```
aarthificial, line 74: Instead of overriding the `image` method and creating this new computed method, you should be able to simply override the getter and setter for `src`:
```ts
protected getSrc(): string {
  // return the url to svg
}

protected setSrc() {
  // ignore the setter
}
```

aarthificial (changes requested):

packages/2d/src/components/Icon.ts
WaldemarLehner: Reason being that I did not know about `colorSignal` :D will do

packages/2d/src/components/Icon.ts
WaldemarLehner: Yup. This does work. It's just that my IDE (VSCode) doesn't consider this to an override.

aarthificial: Awesome!
> It's just that my IDE (VSCode) doesn't consider this to an override.

Signal setters and getters are magic methods so unfortunately no IDE will mark them correctly. But whenever there's a signal property, you can implement `get[Property]` and `set[Property]` methods that will override the signals default behavior.

Tandashi:
Be sure to squash the `fixup` commits together so they are just one commit (`feat(2d) add Icon Component`).
You should be able to achieve with what @hhenrichsen mentioned: `git rebase --interactive --autosquash upstream/main`

WaldemarLehner:
image

done :)

packages/2d/src/components/Icon.ts
```diff
@@ -0,0 +1,89 @@
...
+   * Color of the Icon. Defaults to white.
+   * Provide the color in one of the following formats:
+   * * named color like `red`, `darkgray`, …
+   * * hexadecimal string with # `#bada55`,`#141414`
+   * Values can be either RGB or RGBA: `#bada55` or `#bada55aa` (latter is partially transparent)
+   * The shorthand version (e.g. `#abc` for `#aabbcc`) is also possible
+   */
+  @initial('white')
+  @colorSignal()
+  public declare color: ColorSignal<this>;
```
exdeejay, line 55: Even though there's no corresponding `stroke` property (which would be an awesome feature for icons but that's a feature for another day), I'd say that this should be called `fill` to be consistent with the majority of the other components.

WaldemarLehner: `Image`, the parent class of `Icon`, uses `fill` to set the color of the background image, so I don't think that is a good idea.

See for example:
```jsx
<Layout layout direction={"row"}>
      <Icon icon="mdi:alert-minus" color={"red"} size={200} fill={"yellow"} />
      <Image src={"https://upload.wikimedia.org/wikipedia/commons/2/2a/Alpha_compositing.svg"} size={[400, 200]} fill={"gray"} />
    </Layout>  
```
results in 

image

exdeejay: ah ok, disregard then

packages/2d/src/components/Icon.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export interface IconProps extends ImageProps {
+  /**
+   * Expects an Icon Identifier.
+   * You can find identifiers on [Icônes](https://icones.js.org/collection/all).
+   * They can look like this:
+   * * mdi:language-typescript
+   * * ph:anchor-simple-bold
+   * * ph:activity-bold
+   */
```
aarthificial, line 15: The properties from the interfaces should inherit the documentation from the class:
```suggestion
  /**
   * {@inheritDoc Icon.icon}
   */
```

packages/2d/src/components/Icon.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+/**
+ * An Icon Component that provides an easy access to over 150k icons.
+ * See https://icones.js.org/collection/all for all available Icons.
+ */
+export class Icon extends Image {
+  /**
+   * Expects an Icon Identifier.
+   * You can find identifiers on [Icônes](https://icones.js.org).
```
aarthificial, line 36: The header should simply state what the property is and remarks should be in their own section:
```suggestion
   * The identifier of the icon.
   *
   * @remarks
   * You can find identifiers on [Icônes](https://icones.js.org).
```

packages/2d/src/components/Icon.ts
```diff
@@ -0,0 +1,89 @@
...
+  @signal()
+  public declare icon: SimpleSignal<string, this>;
+
+  /**
+   * Color of the Icon. Defaults to white.
+   * Provide the color in one of the following formats:
+   * * named color like `red`, `darkgray`, …
+   * * hexadecimal string with # `#bada55`,`#141414`
+   * Values can be either RGB or RGBA: `#bada55` or `#bada55aa` (latter is partially transparent)
+   * The shorthand version (e.g. `#abc` for `#aabbcc`) is also possible
```
aarthificial, line 51: The default value should be in its own section:
```suggestion
   * The color of the icon.
   * 
   * @remarks
   * Provide the color in one of the following formats:
   * * named color like `red`, `darkgray`, …
   * * hexadecimal string with # `#bada55`,`#141414`
   * Values can be either RGB or RGBA: `#bada55` or `#bada55aa` (latter is partially transparent)
   * The shorthand version (e.g. `#abc` for `#aabbcc`) is also possible

   * @default `'white'`
```

packages/2d/src/components/Icon.ts
```diff
@@ -0,0 +1,89 @@
...
+  @initial('white')
+  @colorSignal()
+  public declare color: ColorSignal<this>;
+
+  public constructor(props: IconProps) {
+    super(props);
+  }
+
+  /**
+   * This creates the URL that will be used as the Image source
```
aarthificial, line 62: Function description should use imperative present tesnse:
```suggestion
   * Create the URL that will be used as the Image source
```

aarthificial (changes requested):
I slept on it and we'll be merging this.

I left some comments concerning the documentation, though.
Apologies for not having the conventions written down somewhere, there's an issue for this in #11

WaldemarLehner:
@aarthificial 
I applied all changes you requested. I also changed the color's type in the interface:
```diff
- color?: SignalValue<string>;
+ color?: SignalValue<PossibleColor>;
```
Given that `color` is a `colorSignal`, this should be the correct type, right?

aarthificial:
> Given that color is a colorSignal, this should be the correct type, right?

Yes

WaldemarLehner:
Okay… there is a problem with this implementation:

It makes use for remote resources. When the App tries to export the canvas, we get a Tainted Canvas error.

#338  has to be solved before this can be merged.

aarthificial (approved):

Issue #204, opened 2023-02-05
Playback speed is reset after saving with erroneous/faulty code and correcting it


Gaweringo:
**Describe the bug**
When making changes in the `view.add()` section that are erroneous and saving them the playback stops, as the code has errors, which is expected.
When correcting the code and saving again, playback resumes automatically but not at the before-selected playback speed.
Pausing and unpausing the playback does not fix this. To fix, a different playback speed needs to be selected and then the old, desired playback speed has to be applied again to get the desired playback speed back.

 **To Reproduce** 
<details>
  <summary>Expand for details:

Change e.g fill to an invalid hex color while playback is running at x0.25. Change the color back to a valid one, save and playback resumes at 1x although x0.25 is still selected.</summary>

1. Create a working example:
```ts
export default function* problem(scene: Scene) {
  const myText = createRef<Text>();

  view.add(<Text ref={myText} fill={`#ffff`}>Test</Text>);

  yield* myText().rotation(360, 1).to(0, 1);
}
```
Result:
![The text rotating at x1 speed as expected](https://user-images.githubusercontent.com/40121865/216845795-8b119cbf-b027-4f04-852b-d6bdf1e6ec2d.gif)

2. Set the playback speed to x0.25
![The text rotating slowly at x0.25 playback speed](https://user-images.githubusercontent.com/40121865/216845871-f17a3d06-ffcb-47a5-8ce9-1707d4edf870.gif)

3. Keep the playback running at x0.25 and change the fill color of the text to `#ff` which is not a valid hex color
```ts
export default function* problem(scene: Scene) {
  const myText = createRef<Text>();

  view.add(<Text ref={myText} fill={`#ff`}>Test</Text>);

  yield* myText().rotation(360, 1).to(0, 1);
}
```
Save the file and playback automatically stops as an error is present:
![Image of the paused player with the error shown](https://user-images.githubusercontent.com/40121865/216846027-fe6c2431-1969-4b1a-a167-bb33ee83652b.png)

4. Change the fill back to a correct hex color like in `1.`
```ts
export default function* problem(scene: Scene) {
  const myText = createRef<Text>();

  view.add(<Text ref={myText} fill={`#ffff`}>Test</Text>);

  yield* myText().rotation(360, 1).to(0, 1);
}
```
Save the file -> playback resumes but at 1x speed, although x0.25 is still selected
[...]

PR #339, opened 2023-02-14
fix(core): playback speed is reset after saving with faulty code (#204).

AshishBarvaliya:
Removing this code fixes the problem, but it's unclear why it was added in the first place. I don't see any issue by removing this.
However, please do let me know if this code is essential for a particular scenario. If that is the case, we'll explore alternative solutions to fix the issue.

Issue(#204)

### DIFF
--- a/packages/core/src/Project.ts
+++ b/packages/core/src/Project.ts
@@ -331,14 +331,11 @@ export class Project {
   public async recalculate() {
     this.previousScene = null;
 
-    const speed = this._speed;
-    this._speed = 1;
     this.frame = 0;
     const scenes = [...this.scenes.current];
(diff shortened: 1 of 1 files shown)


aarthificial (changes requested):
It's set to `1` to make things faster - for example, using 0.25 would result in a 4x longer recalculation. While using 2 would result in the scene duration being rounded.

I suggest we wrap the for loop in a `try/catch` to make sure that the speed is always reset even when an error occurs, and then simply rethrow the error.

AshishBarvaliya:
updated the changes. working fine now.

aarthificial (approved):

PR #340, opened 2023-02-14
feat(core): add Matrix2D type

ksassnowski:
Adds a new `Matrix2D` type to represent a 2D transformation. This is a specialized 2x3 matrix based on the `mat2d` found in [gl-matrix](https://github.com/toji/gl-matrix).

## Constructing the matrix

The matrix can be constructed in a number of ways. Rather than listing them all here, I'm just going to paste the test cases that deal with constructing a matrix:

```
     ✓ construction (16)
       ✓ can be constructed from an existing matrix
       ✓ can be constructed from an array of three PossibleVector2
       ✓ can be constructed from individual numbers
       ✓ can be constructed from a three sets of two numbers
       ✓ can be constructed from array of six numbers
       ✓ can be constructed from three PossibleVector2 [number, number]
       ✓ can be constructed from three PossibleVector2 SerializedVector2
       ✓ can be constructed from three PossibleVector2 {width: T, height: T}
       ✓ can be constructed from three PossibleVector2 Vector2
       ✓ can be constructed from three PossibleVector2 number
       ✓ can be constructed from three PossibleVector2 undefined
       ✓ can be constructed from a rotation
       ✓ can be constructed from a translation: non-uniform
       ✓ can be constructed from a translation: uniform
       ✓ can be constructed from a scaling: non-uniform
       ✓ can be constructed from a scaling: uniform
```

### Static constructors

- `Matrix2D.identity` returns the 2d transformation identity matrix
- `Matrix2D.zero` returns a 2x3 zero matrix
- `Matrix2D.fromTranslation` creates a matrix from an initial translation by setting `tx` and `ty` accordingly
- `Matrix2D.fromRotation` creates a rotation matrix from an angle (in degrees)
- `Matrix2D.fromScale` creates a transformation matrix for the provided scale

## Properties

The matrix exposes a handful of getters and setters to retrieve and manipulate the various component vectors of the matrix. Unless otherwise specified, all properties also expose a respective setter. Not that using a setter will _mutate_ the matrix. This is also why `Matrix2D.identity` and `Matrix2D.zero` cannot be singletons.

- `Matrix2D.translateX` - The horizontal translation (`tx`)
- `Matrix2D.translateY` - The vertical translation (`ty`)
- `Matrix2D.translation` - The x and y translation as a `Vector2`
- `Matrix2D.scaleX` - The horizontal scaling (`a`)
- `Matrix2D.scaleY` - The vertical scaling (`d`)
- `Matrix2D.scaling` - The x and y scaling as a `Vector2`
- `Matrix2D.skewX` - The horizontal scaling (`c`)
- `Matrix2D.skewY` - The vertical scaling (`b`)
- `Matrix2D.skew` - The x and y skew as a `Vector2`
- `Matrix2D.rotation` - The rotation described by the matrix in degrees. Note that this can also be set in degrees and will be translated to the correct rotation matrix internally.

For compatibility reasons, there is also a `domMatrix` getter to retrieve the equivalent `DOMMatrix` of the `Matrix2D`.

Furthermore, there are getters for common operations on a matrix, as well:

- `Matrix2D.determinant` - Returns the determinant of the matrix
- `Matrix2D.inverse` - Returns the inverse of the matrix if the matrix is invertible (determinant != 0). Otherwise returns `null`

There is no transpose defined for a `Matrix2D`.

## Methods

All methods are pure, meaning they will return a new matrix that represents the result of the computation instead of mutating the matrix.

- `Matrix2D.mul` - Calculates the matrix product with another `Matrix2D`
- `Matrix2D.scale` - Scales the x and y component vectors of the matrix. Does not affect `tx` and `ty` (see `mulScalar` for that )
- `Matrix2D.mulScalar` - Scales every value of the matrix by a value (including `tx` and `ty`)
- `Matrix2D.add` - Matrix addition
- `Matrix2D.sub` - Matrix subtraction
- `Matrix2D.equals` - Checks if the matrix is equal to another matrix by comparing their values. Includes a small tolerance to account for floating point imprecisions (default `0.000001`)
- `Matrix2D.exactlyEquals` - Like `equals` but without the tolerance

## Next steps

In the next PR, I would like to update most method signatures that currently only accept `DOMMatrix` to also accept a `Matrix2D`. This allows us to slowly start migrating over to this new type without introducing a massive BC break by changing the signatures outright.

I don't want to include these changes in this PR as it would introduce way to many changes and make the PR essentially unreviewable.


packages/core/src/types/Matrix2D.ts
```diff
@@ -0,0 +1,663 @@
...
+ *   - A(B + C) = AB + AC does not hold for a Matrix2D
+ *   - (rA)^-1 = r^-1 A^-1, r != 0 does not hold for a Matrix2D
+ *   - r(AB) = (rA)B = A(rB) does not hold for a Matrix2D
+ */
+export class Matrix2D {
+  public readonly values: Float32Array = new Float32Array(6);
+
+  public static get identity(): Matrix2D {
+    return new Matrix2D(1, 0, 0, 1, 0, 0);
+  }
```
aarthificial, line 30: This could be a constant:
```suggestion
  public static readonly identity = new Matrix2D(1, 0, 0, 1, 0, 0);
```

packages/core/src/types/Matrix2D.ts
```diff
@@ -0,0 +1,663 @@
...
+
+  public static get zero(): Matrix2D {
+    return new Matrix2D(0, 0, 0, 0, 0, 0);
+  }
+
+  public static fromRotation(angle: number): Matrix2D {
+    const matrix = Matrix2D.identity;
+    matrix.rotation = angle;
+    return matrix;
+  }
```
aarthificial, line 40: Immutability:
```suggestion
  public static fromRotation(angle: number): Matrix2D {
    return Matrix2D.identity.rotate(angle);
  }
```

packages/core/src/types/Matrix2D.ts
```diff
@@ -0,0 +1,663 @@
...
+ *
+ * Note that because a Matrix2D ignores the z-values of each component vectors,
+ * it does not satisfy all properties of a "real" 3x3 matrix.
+ *
+ *   - A Matrix2D has no transpose
+ *   - A(B + C) = AB + AC does not hold for a Matrix2D
+ *   - (rA)^-1 = r^-1 A^-1, r != 0 does not hold for a Matrix2D
+ *   - r(AB) = (rA)B = A(rB) does not hold for a Matrix2D
+ */
+export class Matrix2D {
```
aarthificial, line 25: Could we also add some getters for retrieving the x and y unit vectors from a matrix?

packages/core/src/types/Matrix.ts
```diff
@@ -1,4 +1,14 @@
...
+import {PossibleVector2, Vector2} from './Vector';
+
+export * from './Matrix2D';
+
+export const EPSILON = 0.000001;
+
+export type PossibleMatrix2D =
+  | Matrix2D
+  | [number, number, number, number, number, number]
+  | [PossibleVector2, PossibleVector2, PossibleVector2];
```
aarthificial, line 11: `undefined` should also be a possible matrix:
```suggestion
export type PossibleMatrix2D =
  | Matrix2D
  | [number, number, number, number, number, number]
  | [PossibleVector2, PossibleVector2, PossibleVector2]
  | undefined;
```

packages/core/src/types/Matrix.ts
```diff
@@ -1,4 +1,14 @@
-import {Vector2} from './Vector';
+import {Matrix2D} from './Matrix2D';
+import {PossibleVector2, Vector2} from './Vector';
+
+export * from './Matrix2D';
+
+export const EPSILON = 0.000001;
+
+export type PossibleMatrix2D =
```
aarthificial, line 8: This declaration should be moved to `Matrix2D.ts`

packages/core/src/types/Matrix.ts
```diff
@@ -1,4 +1,14 @@
-import {Vector2} from './Vector';
+import {Matrix2D} from './Matrix2D';
+import {PossibleVector2, Vector2} from './Vector';
+
+export * from './Matrix2D';
```
aarthificial, line 4: The export should be added to `index.ts` instead.

packages/core/src/types/Matrix2D.ts
ksassnowski: That won't work if we also want to include the setters as that would mean you could do

```ts
Matrix2D.identity.translation = [100, 100];

Matrix.identity.translation;
// => Vector2(100, 100)
```

aarthificial: Yes, but the types are meant to be immutable. If you mutate a type it means that you know what you're doing.
It should be:
```ts
Matrix2D.identity.translate(100, 100)
```
We have the same thing going on in `Vector2`

packages/core/src/types/Matrix2D.ts
ksassnowski: So the equivalent of calling `.column(0)` and `.column(1)`?

aarthificial: Nevermind, I forgot it's a 3x2 matrix and `column` already returns a Vector2

packages/core/src/types/Matrix2D.ts
```diff
@@ -0,0 +1,663 @@
...
+    const vec = new Vector2(translation);
+    this.values[4] = vec.x;
+    this.values[5] = vec.y;
+  }
+
+  public get translation(): Vector2 {
+    return new Vector2(this.values[4], this.values[5]);
+  }
+
+  public set scaling(value: PossibleVector2<number>) {
```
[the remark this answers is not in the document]
ksassnowski: Yes, the idea behind the setters was that they _set_ the value to a specific value, essentially overriding it. To transform the existing matrix you would use the `matrix.scale()` method, instead. I don't feel super strongly about this, but

```
matrix.scaling = 2;
```

implies that I want this matrix to represent a scaling of 2, i.e. _set_ the scaling to 2, not scale the current matrix by 2.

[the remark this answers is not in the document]
aarthificial: I see, but technically, couldn't you normalize the 0th and 1st columns and multiply that by the `value`? 
This would achieve the same thing and would still work for rotated matrices, I think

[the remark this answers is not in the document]
ksassnowski: I guess the general question is if we even want to keep the setters. Because there's no point in having them behave the same way the methods do.

packages/core/src/types/Matrix2D.ts
```diff
@@ -0,0 +1,663 @@
...
+      this.values[0],
+      this.values[1],
+      this.values[2],
+      this.values[3],
+      this.values[4],
+      this.values[5],
+    ]);
+  }
+
+  public constructor(matrix: PossibleMatrix2D);
```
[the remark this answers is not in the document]
ksassnowski: I assume an empty constructor would return the identity matrix?

packages/core/src/types/Matrix2D.ts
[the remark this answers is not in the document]
ksassnowski: Oh, sorry I had a bit of a brain fart there. You're saying that this method currently wouldn't work for a matrix that's already rotated because it's ignoring the horizontal and vertical skew, resulting in a shear. You're absolutely right.

Yeah, normalizing and multiplying should work. Do you think the setters are useful to keep around?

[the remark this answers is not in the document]
ksassnowski: Alright, I've updated both the setters for `scale`, `scaleX`, `scaleY`, and `rotation` so that they set the absolute value of the transformation while keeping the other transformations proportinal.

I've decided to remove the `skew` getters and setters for now as they seemed a little awkward and not that useful overall. Let me know if you'd like them back or not.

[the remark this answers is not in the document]
ksassnowski: I'm also an idiot and left a `test.only` in there 🤦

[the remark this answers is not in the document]
ksassnowski: Fixed 👍

aarthificial (approved):

PR #341, opened 2023-02-14
feat(core): error double event name

Thomsr:
Error message when you use an event name that has already been used.

### DIFF
--- a/packages/core/src/scenes/TimeEvents.ts
+++ b/packages/core/src/scenes/TimeEvents.ts
@@ -1,6 +1,5 @@
 import type {Scene, SceneMetadata} from './Scene';
 import {ValueDispatcher} from '../events';
-import {useLogger} from '../utils';
 
 /**
  * Represents a time event at runtime.
@@ -115,12 +114,10 @@ export class TimeEvents {
     if (!this.collisionLookup.has(name)) {
       this.collisionLookup.add(name);
     } else {
-      useLogger().error(
-        'name: "' +
-          name +
-          '"' +
-          ' has already been used for another event name.',
-      );
+      this.scene.project.logger.error({
+        message: `name "${name}" has already been used for another event name.`,
+        stack: new Error().stack,
+      });
       return 0;
     }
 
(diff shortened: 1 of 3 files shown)


packages/core/src/scenes/TimeEvents.ts
```diff
@@ -109,6 +112,18 @@ export class TimeEvents {
...
   public register(name: string): number {
+    if (!this.collisionLookup.has(name)) {
+      this.collisionLookup.add(name);
+    } else {
+      useLogger().error(
+        'name: "' +
+          name +
+          '"' +
+          ' has already been used for another event name.',
+      );
```
aarthificial, line 123: Three things:
- you can access the logger through `this.scene.project.logger`
- it's better to use template literals for messages like this:
- you can add a stack trace to the log to show people the where in the code this event was:
```suggestion
      this.scene.project.logger.error({
        message: `name "${name}" has already been used for another event name.`,
        stack: new Error().stack,
      });
```

packages/template/src/scenes/example.meta
```diff
@@ -3,7 +3,7 @@
   "timeEvents": [
     {
       "name": "circle",
-      "targetTime": 3
+      "targetTime": 0.9940212833664561
```
aarthificial, line 6: You can't change the `template` package because it's used for testing the rendering.
So any visual/timing change will make the tests fail.

aarthificial (changes requested):

packages/template/src/scenes/example.meta
Thomsr: _facepalm_, I forgot sorry

packages/core/src/scenes/TimeEvents.ts
```diff
@@ -109,6 +111,16 @@ export class TimeEvents {
    * @internal
    */
   public register(name: string): number {
+    if (!this.collisionLookup.has(name)) {
```
ksassnowski, line 114: This might be personal preference, but how about inverting this condition and instead return early if the name has already been used? That would allow us to get rid of one level of indentation since we can get rid of the `else` branch.

```suggestion
    if (this.collisionLookup.has(name)) {
      this.scene.project.logger.error({
        message: `name "${name}" has already been used for another event name.`,
        stack: new Error().stack,
      });
      return 0;
    }
 
    this.collisionLookup.add(name);
 
 // rest...
```

Thomsr: Good suggestion, I would agree to change it to use one less indentation.

aarthificial: I agree

PR #349, opened 2023-02-14
docs: add update guide

### LINKED ISSUE #288: Docs: Add an updating guide
**Description**
We probably want a page that describes an update process, as well as any migration steps for breaking changes when they happen.

Tandashi:
## What changed in this Pull Request
A new page for the update guide has been added to the documentation.
In the sidebar it is located below the `Advanced` section since it doesn't really fit to any of the other topics. (Inspired by NextJS Documentation)

This page might contain migration guides as well as breaking changes once we have some.

---

Closes #288

### DIFF
--- a/packages/docs/docs/updating.mdx
+++ b/packages/docs/docs/updating.mdx
@@ -1,0 +1,21 @@
+---
+sidebar_position: 100
+slug: /updating
+---
+
+# Update Guide
+
+The latest patch notes can be found
+[here](https://github.com/motion-canvas/motion-canvas/releases).
+
+To update all the `@motion-canvas` packages to the latest version use the
+following:
+
+```bash
+npm update --save @motion-canvas/2d @motion-canvas/core @motion-canvas/ui @motion-canvas/vite-plugin
+```
+
+## Toubleshooting
+
+If you get any errors, try removing `package-lock.json` and the `node_modules`
+folder, then run `npm install` again.


packages/docs/docs/updating.mdx
```diff
@@ -0,0 +1,22 @@
+---
+sidebar_position: 3
```
aarthificial, line 2: Let's give it some bigger number so that it's always at the end:
```suggestion
sidebar_position: 100
```

aarthificial (changes requested):

packages/docs/docs/updating.mdx
```diff
@@ -0,0 +1,22 @@
...
+To update all the `@motion-canvas` packages to the latest version use the
+following:
+
+```bash
+npm update --save @motion-canvas/2d @motion-canvas/core @motion-canvas/ui @motion-canvas/vite-plugin
+```
+
+## Toubleshooting
+
+If you get any errors try to remove the `package-lock.json` and the
```
hhenrichsen, line 20: nit: "If you get any errors, try removing `package-lock.json` and the `node_modules` folder, then run `npm install` again.

packages/docs/docs/updating.mdx
```diff
@@ -0,0 +1,22 @@
...
+`node_modules` folder.  
+Then run `npm install` again.
```
hhenrichsen, line 22: nit: This probably should be on the previous line since it's not double spaced (and therefore should render on the same line as above)

Tandashi: Should be fixed with the one-liner you proposed :)

Tandashi:
No clue what went wrong here and why the PR suddenly closed...

aarthificial (approved):

PR #351, opened 2023-02-15
fix(vite-plugin): Ignore query param in devserver

rmhartog:
The previous implementation tested the full url, including query parameters against the '/' path. Any query parameter in the URL would cause a 404 instead of loading the editor.

Parsing the request URL with the URL class gives us access to the pathname, which does not include query parameters and can be safely compared.

---

I ran into the issue when testing in VS Code and opening `http://localhost:9000/` in the builtin 'Simple Browser'. It simply resulted in a blank page. After some digging I noticed that it automatically appends the query parameter `/?vscodeBrowserReqId=<random number>`, causing the vite server to respond with 404.

### DIFF
--- a/packages/vite-plugin/src/main.ts
+++ b/packages/vite-plugin/src/main.ts
@@ -296,13 +296,16 @@ export default ({
           return;
         }
 
-        if (req.url === '/') {
+        const url = req.url
+          ? new URL(req.url, `http://${req.headers.host}`)
+          : undefined;
+        if (url?.pathname === '/') {
           res.setHeader('Content-Type', 'text/html');
           res.end(createHtml('/@id/__x00__virtual:editor'));
           return;
         }
 
-        const name = req.url?.slice(1);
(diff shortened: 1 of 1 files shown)


packages/vite-plugin/src/main.ts
```diff
@@ -296,13 +296,14 @@ export default ({
           return;
         }
 
-        if (req.url === '/') {
+        const url = new URL(req.url, `http://${req.headers.host}`);
```
aarthificial, line 299: You need to account for `req.url` possibly being `undefined`

aarthificial (changes requested):

packages/vite-plugin/src/main.ts
rmhartog: Oops, missed that case, added a check and let `url` be `undefined` otherwise.

aarthificial (approved):

aarthificial:
Thanks for the fix!

Issue #353, opened 2023-02-15
Allow doing `shift + right arrow` to go to the end of the timeline


BlockOG:
**Description**
Currently `shift + right arrow` does the same thing as just `right arrow` (moves you 1 frame forward), while `shift + left arrow` moves you to the start of the timeline (and just `left arrow` moves you 1 frame back)

**Proposed solution**
Make `shift + right arrow` move you to the end of the timeline

aarthificial:
For reference, playback shortcuts are handled here:
https://github.com/motion-canvas/motion-canvas/blob/18f61a668420dec8afba52d52a6557e7a7919ba2/packages/ui/src/components/playback/PlaybackControls.tsx#L38-L50

PR #354, opened 2023-02-15
feat(ui): shift + right arrow moves to last frame

ribru17:
Press shift + right arrow in the editor to move to the last frame for easier navigation.

Implements #353


aarthificial:
Thanks for the PR!

Issue #352, opened 2023-02-15
Switch from `null` to `DEFAULT`


aarthificial:
**Description**
#259 introduced a new `DEFAULT` primitive for representing default values of signals.

**Proposed solution**
Layout properties utilizing the `Length` type (`width`, `lineHeight`, etc.) should switch from using `null` to `DEFAULT`.

PR #355, opened 2023-02-15
feat(2d): unify layout properties

aarthificial:
- External interfaces for `size` and `lineHeight` no longer use `null` values. Users can use `DEFAULT` to achieve the same thing.
- `lineHeight` supports inheritance like other CSS properties.
- `gap` is now a compound `Vector2` signal.
- Added `width` and `height` as shorthands for `size.x` and `size.y` respectively.

Closes: #352

### DIFF
--- a/packages/2d/src/components/Layout.ts
+++ b/packages/2d/src/components/Layout.ts
@@ -44,6 +44,7 @@ import {drawLine, lineTo} from '../utils';
 import {spacingSignal} from '../decorators/spacingSignal';
 import {
   createSignal,
+  Signal,
   SignalValue,
   SimpleSignal,
 } from '@motion-canvas/core/lib/signals';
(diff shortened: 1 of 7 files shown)


aarthificial:
@exdeejay, could you take a look?

packages/2d/src/components/Layout.ts
```diff
@@ -146,15 +148,19 @@ export class Layout extends Node {
...
-  @initial(null)
+  @initial(0)
   @signal()
   public declare readonly gap: SimpleSignal<Length, this>;
-  @initial(null)
   @signal()
   public declare readonly rowGap: SimpleSignal<Length, this>;
-  @initial(null)
+  protected getDefaultRowGap() {
+    return this.gap();
```
exdeejay, line 157: Having both a horizontal and vertical gap makes me think a compound signal would be better here, especially since this will result in:
```tsx
object.rowGap(10);
object.gap(20);
debug(object.rowGap()); // 10
```

exdeejay (approved):

PR #356, opened 2023-02-15
feat(core): added spring interpolation

### LINKED ISSUE #334: Physical spring animations
**Description**
Spring animations using real physical spring simulations look often better than using approximations using a mix of squareroots and sines like elastic easing. The down site to this is that you are not able to set the duration exactly but just wait the simulation to cool down.

**Proposed solution**
I have a working example already built having a `tween`-like API.
```ts
yield* springInterpolation(spring, x => {
   circle().position.x(map(0, 300, x));
})
```

with `spring` being type of `Spring`:

```ts
interface Spring {
  mass: number,
  stiffness: number,
  damping: number,
  initalVelocity: number
}
```
This way it is using a known spring syntax and you can easily import different springs from generators like [Couverture](https://apps.apple.com/us/app/couverture/id1552415914).

What would be nice too, to have an implementation of an spring generator like mentioned [here](https://github.com/jenox/UIKit-Playground/tree/master/01-Demystifying-UIKit-Spring-Animations/), which helps to generate springs in a more human friendly way. 

**Considered alternatives**
We could use `react-spring`, but I would love to come up with an own, light and more motion-canvasy way.

**Additional context**

https://user-images.githubusercontent.com/35671734/218494364-6ef5bac0-0881-4a2a-98ad-74357d4d328e.mov

I would really love to claim this implementation and contribute to this project. :-)

gustavneustadt:
Resolves #334

I am very excited to contribute to the project again. I would really appreciate a code review. Please feel free and do not hesitate to point out best practices or wrong concepts.

I've added partly random example springs to choose from and an interface, but I really don't know if this really fits the overall code style of MC. 

Some notes to the tests: 
I noticed that the springs are very sensitive to the projects frame rate. I assume that the higher the frame rate, the steps are smaller and result in a more accurate simulation. Actually MC crashes at 10 fps and I do not have any clue why this happens – maybe a floating point issue.

 
Usage:
```ts
yield* spring(smoothSpring, 0, 100, .5, x => circle().position.x(x))
yield* spring({
   mass: .3,
   stiffness: 10,
   damping: .1
}, 0, 100, .5, x => circle().position.x(x))
yield* spring(makeSpring(.3,10,.1), 0, 100, .5, x => circle().position.x(x))
```

### DIFF
--- a/packages/core/src/tweening/spring.ts
+++ b/packages/core/src/tweening/spring.ts
@@ -1,0 +1,116 @@
+import {decorate, threadable} from '../decorators';
+import {ThreadGenerator} from '../threading';
+import {useProject, useThread, useLogger} from '../utils';
+
+decorate(spring, threadable());
+export function* spring(
+  spring: Spring | null,
+  from: number,
+  to: number,
+  settleTolerance: number | null,
+  onProgress: (value: number, time: number) => void,
+  onEnd?: (value: number, time: number) => void,
+): ThreadGenerator {
+  spring = spring ?? {
+    mass: 0.05,
+    stiffness: 10,
+    damping: 0.5,
+  };
+
+  if (spring.mass <= 0) {
+    useLogger().error(new Error('Spring mass must be greater 0.'));
+    return;
+  }
+  if (spring.stiffness < 0) {
+    useLogger().error(
+      new Error('Spring stiffness must be greater or equal to 0.'),
+    );
+    return;
+  }
+  if (spring.damping < 0) {
+    useLogger().error(
+      new Error('Spring damping must be greater or equal to 0.'),
+    );
+    return;
+  }
+
+  const project = useProject();
+  const thread = useThread();
+
+  const startTime = thread.time();
+
+  // figure out the step length for each frame
+  const timeStep = 1 / project.framerate;
+
+  settleTolerance = settleTolerance ?? 0.001;
+  let position = from;
+  let velocity = spring.initalVelocity ?? 0;
+  let settled = false;
+
+  onProgress(from, 0);
+  while (!settled) {
+    const positionDelta = position - to;
+
+    // using hooks law: F=-kx; with k being the spring constant and x the offset to the settling position
+    const force = -spring.stiffness * positionDelta - spring.damping * velocity;
+
+    //updating the velocity
+    velocity += (force / spring.mass) * timeStep;
+
+    // apply velocity to the position
+    position += velocity * timeStep;
+
+    // the simulation would run forever but is often barely noticable after some seconds
+    // we fix this by checking if the delta delta between $to and $position is smaller than the given $settleTolerance
+    // but we also need to check if the velocity is below the tolerance too
+    if (
+      Math.abs(to) - Math.abs(position) < settleTolerance &&
+      Math.abs(velocity) < settleTolerance
+    ) {
+      settled = true;
+    }
+
+    const time = project.time - startTime;
+    const value = position;
+
+    if (time > 0) {
+      onProgress(value, time);
+    }
+    yield;
+  }
+  const endTime = thread.time();
+
+  thread.time(endTime);
+
+  onProgress(to, endTime);
+  onEnd?.(to, endTime);
+}
+
+export interface Spring {
+  mass: number;
+  stiffness: number;
+  damping: number;
+  initalVelocity?: number;
+}
+
+export function makeSpring(
+  mass: number,
+  stiffness: number,
+  damping: number,
+  initalVelocity?: number,
+): Spring {
+  return {
+    mass,
+    stiffness,
+    damping,
+    initalVelocity,
+  };
+}
+
+export const beatSpring: Spring = makeSpring(0.13, 5.7, 1.2, 10.0);
+export const ploppSpring: Spring = makeSpring(0.2, 20.0, 0.68, 0.0);
+export const bounceSpring: Spring = makeSpring(0.08, 4.75, 0.05, 0.0);
+export const swingSpring: Spring = makeSpring(0.39, 19.85, 2.82, 0.0);
+export const jumpSpring: Spring = makeSpring(0.04, 10.0, 0.7, 8.0);
+export const strikeSpring: Spring = makeSpring(0.03, 20.0, 0.9, 4.8);
+export const smoothSpring: Spring = makeSpring(0.16, 15.35, 1.88, 0.0);
--- a/packages/core/src/tweening/spring.ts
+++ b/packages/core/src/tweening/spring.ts
@@ -2,15 +2,50 @@ import {decorate, threadable} from '../decorators';
 import {ThreadGenerator} from '../threading';
 import {useProject, useThread, useLogger} from '../utils';
 
+type ProgressFunction = (value: number, time: number) => void;
+
 decorate(spring, threadable());
+export function spring(
+  spring: Spring | null,
+  from: number,
+  to: number,
+  settleTolerance: number,
+  onProgress: ProgressFunction,
+  onEnd?: ProgressFunction,
+): ThreadGenerator;
+
+export function spring(
+  spring: Spring | null,
+  from: number,
+  to: number,
+  onProgress: ProgressFunction,
+  onEnd?: ProgressFunction,
+): ThreadGenerator;
+
 export function* spring(
   spring: Spring | null,
   from: number,
   to: number,
-  settleTolerance: number | null,
-  onProgress: (value: number, time: number) => void,
-  onEnd?: (value: number, time: number) => void,
+  settleToleranceOrOnProgress: number | ProgressFunction,
+  onProgressOrOnEnd?: ProgressFunction,
+  onEnd?: ProgressFunction,
 ): ThreadGenerator {
+  const settleTolerance =
+    typeof settleToleranceOrOnProgress === 'number'
+      ? settleToleranceOrOnProgress
+      : 0.001;
+
+  onEnd =
+    typeof settleToleranceOrOnProgress === 'number' ? onEnd : onProgressOrOnEnd;
+
+  const onProgress: ProgressFunction = (value: number, time: number) => {
+    if (typeof settleToleranceOrOnProgress === 'function') {
+      settleToleranceOrOnProgress(value, time);
+    } else if (typeof onProgressOrOnEnd === 'function') {
+      onProgressOrOnEnd(value, time);
+    }
+  };
+
   spring = spring ?? {
     mass: 0.05,
     stiffness: 10,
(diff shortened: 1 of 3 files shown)


aarthificial:
> I noticed that the springs are very sensitive to the projects frame rate

I think the best way to address this would be by using a fixed time step for physics calculations.
In other words, assume that the delta time is always `(1 / 60)s` and perform the update loop however many times necessary to match the framerate. So:
- `60fps = (1 / 60)s = 1 * (1 / 60)s` - 1 iteration
- `30fps = (1 / 30)s = 2 * (1 / 60)s` - 2 iterations
- `10fps = (1 / 10)s = 6 * (1 / 60)s` - 6 iterations

an so on.

gustavneustadt:
makes sense! I give it a shot.

packages/core/src/tweening/spring.ts
```diff
@@ -0,0 +1,116 @@
...
+    useLogger().error(
+      new Error('Spring damping must be greater or equal to 0.'),
+    );
+    return;
+  }
+
+  const project = useProject();
+  const thread = useThread();
+
+  const startTime = thread.time();
```
aarthificial, line 40: You'll need to account for the time offset of the current thread.
Threads can be offset relative to the global time (`project.time`) by values shorter than the duration of one frame. This allows you to do things like `waitFor(0.005)` that work independently of the framerate.
The current implementation will produce incorrect results when dispatching springs in quick succession (the line is not smooth):
Now | Should be
--- | ---
image | image

Repro:
```tsx
import {Circle, Node} from '@motion-canvas/2d/lib/components';
import {sequence, waitFor} from '@motion-canvas/core/lib/flow';
import {makeRef, range} from '@motion-canvas/core/lib/utils';
import {spring} from '@motion-canvas/core/lib/tweening';

  const circles: Circle[] = [];

  view.add(
    <Node>
      {range(9).map(i => (
        <Circle
          ref={makeRef(circles, i)}
          x={(i - 4) * 100}
          y={-400}
          size={50}
          fill={'white'}
        />
      ))}
    </Node>,
  );

  yield* sequence(
    0.03,
    ...circles.map(circle =>
      spring(null, -400, 400, null, value => {
        circle.position.y(value);
      }),
    ),
  );

  yield* waitFor(5);
});
```

One way to do that would be by calculating the time step manually:
```ts
let lastUpdateTime = thread.time();
while (!settled) {
  // this timeStep will account for the difference between the global time and this thread's time.
  const timeStep = project.time - lastUpdateTime; 
  if (timeStep <= 0) {
    yield;
    continue;
  }
  lastUpdateTime = project.time;
  // rest of the calculation
}
```

packages/core/src/tweening/spring.ts
```diff
@@ -0,0 +1,116 @@
...
+    const value = position;
+
+    if (time > 0) {
+      onProgress(value, time);
+    }
+    yield;
+  }
+  const endTime = thread.time();
+
+  thread.time(endTime);
```
aarthificial, line 83: This shouldn't be necessary.

packages/core/src/tweening/spring.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    //updating the velocity
+    velocity += (force / spring.mass) * timeStep;
+
+    // apply velocity to the position
+    position += velocity * timeStep;
+
+    // the simulation would run forever but is often barely noticable after some seconds
+    // we fix this by checking if the delta delta between $to and $position is smaller than the given $settleTolerance
+    // but we also need to check if the velocity is below the tolerance too
```
aarthificial, line 65: Unfortunately, prettier won't do this for you, but comments should also be limited to 80 characters.
Also, a few minor mistakes in spelling/grammar:
```suggestion
    // The simulation would run forever but is often barely noticeable after
    // some seconds. We fix this by checking if the delta between `to` and
    // `position` is smaller than the given `settleTolerance`. But we also need
    // to check if the velocity is below the tolerance too.
```

packages/core/src/tweening/spring.ts
```diff
@@ -0,0 +1,116 @@
...
+  settleTolerance = settleTolerance ?? 0.001;
+  let position = from;
+  let velocity = spring.initalVelocity ?? 0;
+  let settled = false;
+
+  onProgress(from, 0);
+  while (!settled) {
+    const positionDelta = position - to;
+
+    // using hooks law: F=-kx; with k being the spring constant and x the offset to the settling position
```
aarthificial, line 54:
```suggestion
    // Using hooks law: F=-kx; with k being the spring constant and x the offset 
    // to the settling position
```

packages/core/src/tweening/spring.ts
```diff
@@ -0,0 +1,116 @@
...
+  onEnd?: (value: number, time: number) => void,
+): ThreadGenerator {
+  spring = spring ?? {
+    mass: 0.05,
+    stiffness: 10,
+    damping: 0.5,
+  };
+
+  if (spring.mass <= 0) {
+    useLogger().error(new Error('Spring mass must be greater 0.'));
```
aarthificial, line 21:
```suggestion
    useLogger().error(new Error('Spring mass must be greater than 0.'));
```

packages/core/src/tweening/spring.ts
```diff
@@ -0,0 +1,116 @@
...
+  };
+}
+
+export const beatSpring: Spring = makeSpring(0.13, 5.7, 1.2, 10.0);
+export const ploppSpring: Spring = makeSpring(0.2, 20.0, 0.68, 0.0);
+export const bounceSpring: Spring = makeSpring(0.08, 4.75, 0.05, 0.0);
+export const swingSpring: Spring = makeSpring(0.39, 19.85, 2.82, 0.0);
+export const jumpSpring: Spring = makeSpring(0.04, 10.0, 0.7, 8.0);
+export const strikeSpring: Spring = makeSpring(0.03, 20.0, 0.9, 4.8);
+export const smoothSpring: Spring = makeSpring(0.16, 15.35, 1.88, 0.0);
```
aarthificial, line 116: `const`s that are not functions should use PascalCase:
```suggestion
export const BeatSpring: Spring = makeSpring(0.13, 5.7, 1.2, 10.0);
export const PlopSpring: Spring = makeSpring(0.2, 20.0, 0.68, 0.0);
export const BounceSpring: Spring = makeSpring(0.08, 4.75, 0.05, 0.0);
export const SwingSpring: Spring = makeSpring(0.39, 19.85, 2.82, 0.0);
export const JumpSpring: Spring = makeSpring(0.04, 10.0, 0.7, 8.0);
export const StrikeSpring: Spring = makeSpring(0.03, 20.0, 0.9, 4.8);
export const SmoothSpring: Spring = makeSpring(0.16, 15.35, 1.88, 0.0);
```

packages/core/src/tweening/spring.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  onProgress(to, endTime);
+  onEnd?.(to, endTime);
+}
+
+export interface Spring {
+  mass: number;
+  stiffness: number;
+  damping: number;
+  initalVelocity?: number;
```
aarthificial, line 93:
```suggestion
  initialVelocity?: number;
```

aarthificial (changes requested):

gustavneustadt:
Thanks a lot! @aarthificial

packages/core/src/tweening/spring.ts
```diff
@@ -0,0 +1,116 @@
+import {decorate, threadable} from '../decorators';
+import {ThreadGenerator} from '../threading';
+import {useProject, useThread, useLogger} from '../utils';
+
+decorate(spring, threadable());
+export function* spring(
+  spring: Spring | null,
+  from: number,
+  to: number,
+  settleTolerance: number | null,
```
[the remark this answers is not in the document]
gustavneustadt: Thought separating function and spring might be useful – but that's totally made up, so … idk if this is really a good idea.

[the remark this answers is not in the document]
gustavneustadt: Do you mean the settleTolerance, if so:
The settleTolerance is related to the from & to attributes. So imo it may be better to keep them separate. Some would like to spring between 0 and 1 and so they may want a very small tolerance.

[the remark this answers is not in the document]
aarthificial: Maybe the value of `settleTolerance` should be relative to the distance between `from` and `to`?
It just seems like a rarely used option, so having to write `null` all the time could be a bit annoying.
Alternatively, maybe we could provide multiple signatures for `spring`?
One with `spring, from, to, onProgress` and one with `spring, from, to, settleTolerance, onProgress`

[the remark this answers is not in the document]
gustavneustadt: Yea, I see, indeed. 
Ok lets try that.

packages/core/src/tweening/spring.ts
gustavneustadt: Is there another way to account for the offset?

gustavneustadt: Modifying the timestep is breaking the solution I have for a constant simulation time. Maybe it's all nonsense I came up with … so I would really appreciate another code review. :-)

gustavneustadt:
The stair-like animation is not resolved but at least the simulation has the same duration on every framerate now. Maybe you have another idea to solve this.

packages/core/src/tweening/spring.ts
```diff
@@ -0,0 +1,200 @@
...
+      // Sub step until there are no longer sub steps available
+      while (accumulation / step >= 1) {
+        update(step);
+        // remove the substeps from the stash
+        accumulation -= step;
+      }
+
+      // Increase the divider
+      divider *= 10;
+    }
```
aarthificial, line 150: I think we can simplify that by calling `update` once, for the missing fraction of a frame.
For example, let's assume that a frame takes 33ms (30FPS) while the physics should be integrated using 10ms intervals (100FPS), then we should perform four updates per frame:
```
update(10ms);
update(10ms);
update(10ms);
update(3ms); // missing fraction
// 10 + 10 + 10 + 3 = 33ms
```
An update loop for something like this could look as follows:
```ts
const startTime = thread.time();
let simulationTime = startTime;

while (!settled) {
  // while our simulation is behind the project time
  while (simulationTime < project.time) {
    // calculate the remaining difference:
    const difference = project.time - simulationTime;

    // if the remaining time is smaller than our simulation step
    if (timeStep > difference) {
      // update using the remaining difference
      update(difference);
      // make the simulation time equal to the project time to make sure the while loop ends.
      simulationTime = project.time;
    } else {
      // otherwise update using the simulation step
      update(timeStep);
      // progress the simulation time by the same amount
      simulationTime += timeStep;
    }
  }
  
  // handle settling and call onProgress
}
```
Because we initialize the `simulationTime` using the thread time, it will also account for the thread offset.

aarthificial (changes requested):

packages/core/src/tweening/spring.ts
aarthificial: This should help:
https://github.com/motion-canvas/motion-canvas/pull/356#discussion_r1110978669

packages/core/src/tweening/spring.ts
```diff
@@ -0,0 +1,200 @@
...
+  onEnd =
+    typeof settleToleranceOrOnProgress === 'number' ? onEnd : onProgressOrOnEnd;
+
+  const onProgress: ProgressFunction = (value: number, time: number) => {
+    if (typeof settleToleranceOrOnProgress === 'function') {
+      settleToleranceOrOnProgress(value, time);
+    } else if (typeof onProgressOrOnEnd === 'function') {
+      onProgressOrOnEnd(value, time);
+    }
+  };
```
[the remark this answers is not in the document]
gustavneustadt: Weird, I think I tried this approach but got an compiling error. But I could be wrong, will try it.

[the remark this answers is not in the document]
aarthificial: Note the `!` at the end of `onProgressOrOnEnd`
It tells the compiler that you're sure this variable won't be undefined.

packages/core/src/tweening/spring.ts
gustavneustadt: I tried your approach but I don't really know whats wrong with it.

https://user-images.githubusercontent.com/35671734/219866199-a083c89c-5c3a-4e28-9b8b-d02bd1f7b790.mov

aarthificial: Oh yeah, forgot about this, you also need to change this:
```ts
Math.abs(to) - Math.abs(position) < settleTolerance
```
to
```ts
Math.abs(to - position) < settleTolerance
```

Otherwise, a spring starting at -400 with the goal of 400 will end immediately, because the check would succeed

gustavneustadt: Okay that works. But now the animation is not that constant in duration anymore across different framerates. 
But yeah, I assume that smooth animations is more important than duration consistency.

packages/core/src/tweening/spring.ts
```diff
@@ -0,0 +1,200 @@
...
+    if (spring === null) {
+      return;
+    }
+    const positionDelta = position - to;
+
+    // Using hooks law: F=-kx; with k being the spring constant and x the offset
+    // to the settling position
+    const force = -spring.stiffness * positionDelta - spring.damping * velocity;
+
+    //updating the velocity
```
ksassnowski, line 93: I feel that a lot of these comments don't really add much information. Most of them seem to be simply be describing what the next line does. If it were up to me (which it isn't), I'd say that we should remove these "what" comments and focus more on the "why" comments.

For instance, the comment on line 89 is a good comment because it adds context that isn't immediately obvious by the code itself.

gustavneustadt: Yea, that’s the style I often comment. But I am not a good programmer at all, so i‘ll love to take advices.

Besides that, this whole PR is more of an draft at the moment. So feel free to review again when a final solution is found, I will mention you ;-)

packages/core/src/tweening/spring.ts
aarthificial: You're right, I absolutely forgot about the duration.
This would require some modifications to the loop:

```ts
const startTime = thread.time();
let simulationTime = startTime;

while (!settled) {
  while (simulationTime < project.time) {
    const difference = project.time - simulationTime;

    if (timeStep > difference) {
      update(difference);
      simulationTime = project.time;
    } else {
      update(timeStep);
      simulationTime += timeStep;
    }

    // perform the check during every iteration:
    if (
      Math.abs(to - position) < settleTolerance &&
      Math.abs(velocity) < settleTolerance
    ) {
      // set the thread time to simulation time:
      thread.time(simulationTime);
      settled = true;
      // break out when settled
      break;
    }
  }

  // only yield if we haven't settled yet.
  if (!settled) {
    onProgress(position, project.time - startTime);
    yield;
  }
}
```
This will make the durations much more aligned between framerates.
The solution is not perfect, but the problem is really complex and so I think we can leave it like this for now.

gustavneustadt: Okay, will apply your modifications. Thanks a lot! :-)

packages/core/src/tweening/spring.ts
gustavneustadt: I removed some comments or changed them and just kept the comments from jacob‘s changes.

PR #357, opened 2023-02-16
feat(2d, core, vite-plugin): add cors proxy and integrated into Image component

### LINKED ISSUE #234: Can't render size animations if a child contains an image with a src from a URL
**Describe the bug**
When attempting to render an animation of the width of a parent that has an image child element with the src property set to a url, the renderer will crash with the error below. It will work correctly inside the preview, but will error when rendering.

**To Reproduce**
Below is a more or less minimal example that will cause the error

```ts
const USERNAME = 'Reina';
const PICTURE = 'https://media.kitsu.io/users/avatars/171606/large.jpeg';

const fontFamily = 'sans-serif';
const BACKGROUND = '#312631';

	const background = createRef<Layout>();

	view.add(
		<Layout layout justifyContent='end' margin={[0, 80, 80, 0]}>
			<Layout clip width='0%' ref={background}>
				<Rect
					height={122}
					width='100%'
					padding={[11, 25, 11, 20]}
					fill={BACKGROUND}
					layout
					gap={25}
					radius={{
						top: 5,
						left: 5,
						bottom: 0,
						right: 0,
					}}>

					{/* This clip property seem to be the culprit */}
					<Circle width={100} height={100} fill='#000' clip>
						<Image src={PICTURE} width={100} height={100} />
					</Circle>

					<Text
						fill='#ffffff'
						fontFamily={fontFamily}
						fontSize={40}
						fontWeight={700}
						lineHeight={48}
						margin={0}
						padding={0}>
						{USERNAME}
					</Text>
				</Rect>
			</Layout>
		</Layout>
	);

	// Problematic animation
	yield* tween(0.4, (value) => {
		background().size.x(`${easeInOutCubic(value, 0, 100)}%`);
	});
	yield* tween(0.4, (value) => {
		background().size.x(`${easeInOutCubic(value, 100, 0)}%`);
	});
});
```

**Expected behavior**
This should render like it does in the preview window

**Console errors**
image

- core: 2.0.0
- ui: 2.0.0

### LINKED ISSUE #338: Rendering fails when using remote pictures
**Describe the bug**
If a scene has images from remote source, trying to render will result in an error:

> Failed to execute 'toDataURL' on 'HTMLCanvasElement': Tainted canvases may not be exported.
image

**To Reproduce**
Have a scene with a remote Image, then try to render.

```ts
import { makeScene2D } from "@motion-canvas/2d";
import { Image, Layout } from "@motion-canvas/2d/lib/components";
import { waitFor } from "@motion-canvas/core/lib/flow";

export default makeScene2D(function* (scene) {
    scene.add(<Layout>
        <Image scale={.2} src={"https://images.unsplash.com/photo-1515229144611-617d3ce8e108"}/> {/* any remote picture works */}
    </Layout>)
    yield* waitFor(1)
});
```

**Additional context**
Error occurs both on Firefox and Chromium.

WaldemarLehner:
This PR touches the following packages:

### vite-plugin

Adds a new `proxy-middleware.ts` which defines a Middleware which proxies GET-Requests using the `/cors-proxy` Endpoint — this is the "main" part of this PR

Also adds a new virtual module, `virtual:proxysettings` , which returns `{enabled: boolean}`.
This is used by the changes in core. This value is either true or false depending on if the user has disabled to proxy by setting `proxy:false` as an option of the vite plugin.

### core

Adds a new util, `viaProxy`.
Anything that might touch a remote resource can use this helper to proxy the request.
The idea behind `viaProxy` is that it will only rewrite if a resource is rewritable.

* it only rewrites if enabled (checks via `virtual:proxysettings`)
* it restricts itself to `http` and `https` to not break inline `data:` sources
* it only rewrites images with a different `host` 
* it does not proxy a proxied request (calling viaProxy multiple times does not break the address)

### 2d
A small change has been made to the `<Image/>` component.

The computed prop `image` wraps the `src` signal into `withProxy`.

Closes #338 
Closes #234 

Relevant Discord Thread: https://discord.com/channels/1071029581009657896/1076868990573477998

### DIFF
--- a/packages/core/src/utils/proxyUtils.ts
+++ b/packages/core/src/utils/proxyUtils.ts
@@ -1,0 +1,149 @@
+/**
+ * Utility to redirect remote sources via Proxy
+ *
+ * This utility is used to rewrite a request to be routed through
+ * the Proxy instead.
+ */
+
+import {useLogger} from './useProject';
+
+/**
+ * Route the given url through a local proxy.
+ *
+ * @example
+ * This rewrites a remote url like `https://via.placeholder.com/300.png/09f/fff`
+ * into a URI-Component-Encoded string like
+ * `/cors-proxy/https%3A%2F%2Fvia.placeholder.com%2F300.png%2F09f%2Ffff`
+ */
+export function viaProxy(url: string) {
+  if (!isProxyEnabled()) {
+    // Proxy is disabled, so we just pass as-is.
+    return url;
+  }
+
+  if (url.startsWith('/cors-proxy/')) {
+    // Already proxied, return as-is
+    return url;
+  }
+
+  // window.location.hostname is being passed here to ensure that
+  // this does not throw an Error for same-origin requests
+  // e.g. /some/image -> localhost:9000/some/image
+  const selfUrl = new URL(window.location.toString());
+  // inside a try-catch in case the URL cannot be understood
+  try {
+    const expandedUrl = new URL(url, selfUrl);
+    if (!expandedUrl.protocol.startsWith('http')) {
+      // this is probably some embedded image (e.g. image/png;base64).
+      // don't touch and pass as is
+      return url;
+    }
+    if (selfUrl.host === expandedUrl.host) {
+      // This is a request to a "local" resource.
+      // No need to rewrite
+      return url;
+    }
+
+    // Check if the host matches the Allow List.
+    // if not, no rewrite takes place.
+    // will fail in the Editor if the
+    // remote host does not accept anonymous
+    if (!isInsideAllowList(expandedUrl.host)) {
+      return url;
+    }
+  } catch (_) {
+    // in case of error just silently pass as-is
+    return url;
+  }
+
+  // Everything else is a "remote" resource and requires a rewrite.
+  return `/cors-proxy/${encodeURIComponent(url)}`;
+}
+
+/**
+ * Check the provided host is allowed to be routed
+ * to the Proxy.
+ */
+function isInsideAllowList(host: string) {
+  const allowList = getAllowList();
+  if (allowList.length === 0) {
+    // Allow List defaults to allow all if empty
+    return true;
+  }
+  for (const entry of allowList) {
+    if (entry.toLowerCase().trim() === host) {
+      return true;
+    }
+  }
+  return false;
+}
+
+/**
+ * Check if the proxy is enabled via the plugin by checking
+ * for `import.meta.env.VITE_MC_PROXY_ENABLED`
+ *
+ * @remarks The value can either be 'true' of 'false'
+ * (as strings) if present, or be undefined if not run
+ * from a vite context or run without the MC Plugin.
+ */
+export function isProxyEnabled() {
+  if (import.meta.env) {
+    return import.meta.env.VITE_MC_PROXY_ENABLED === 'true';
+  }
+  return false;
+}
+
+/**
+ * Cached value so getAllowList does not
+ * try to parse the Env var on every call,
+ * spamming the console in the process
+ */
+let getAllowListCache: string[] | undefined = undefined;
+/**
+ * Return the list of allowed hosts
+ * from the Plugin Config
+ */
+function getAllowList() {
+  // Condition should get optimized away for Prod
+  if (import.meta.env.VITEST !== 'true') {
+    if (getAllowListCache) {
+      return [...getAllowListCache];
+    }
+  }
+
+  // Inline function gets immediately invoked
+  // and the result stored in getAllowListCache.
+  // The cached value is used on subsequent requests.
+  const result = (function () {
+    if (!isProxyEnabled() || !import.meta.env) {
+      return [];
+    }
+
+    // This value is encoded as a JSON String.
+    const valueJson = import.meta.env.VITE_MC_PROXY_ALLOW_LIST ?? '[]';
+    const parsedJson = JSON.parse(valueJson);
+    // Do an additional check that only strings are present,
+    // create a warning and ignore the value
+    if (!Array.isArray(parsedJson)) {
+      useLogger().error(
+        'Parsed Allow List expected to be an Array, but is ' +
+          typeof parsedJson,
+      );
+    }
+    const validatedEntries = [];
+    for (const entry of parsedJson) {
+      if (typeof entry !== 'string') {
+        useLogger().warn(
+          'Unexpected Value in Allow List: ' +
+            entry +
+            '. Expected a String. Skipping.',
+        );
+        continue;
+      }
+      validatedEntries.push(entry);
+    }
+    return validatedEntries;
+  })();
+  getAllowListCache = result;
+  return [...getAllowListCache];
+}
--- a/packages/vite-plugin/src/main.ts
+++ b/packages/vite-plugin/src/main.ts
@@ -4,6 +4,11 @@ import fs from 'fs';
 import {Readable} from 'stream';
 import mime from 'mime-types';
 import projectInstance from './__logs__/project-instance.md';
+import {
+  motionCanvasCorsProxy,
+  MotionCanvasCorsProxyOptions,
+  setupEnvVarsForProxy,
+} from './proxy-middleware';
 
 export interface MotionCanvasPluginConfig {
   /**
@@ -67,6 +72,16 @@ export interface MotionCanvasPluginConfig {
    * @default '\@motion-canvas/ui'
    */
   editor?: string;
+  /**
+   * Configuration of the Proxy used for remote sources
+   *
+   * @remarks
+   * This passes configuration to Motion Canvas' proxy.
+   * Note that the proxy is disabled by default.
+   * You can either pass `true` and a config object
+   * to enable it.
+   **/
+  proxy?: boolean | MotionCanvasCorsProxyOptions;
 }
 
 interface ProjectData {
@@ -79,6 +94,7 @@ export default ({
   output = './output',
   bufferedAssets = /\.(wav|ogg)$/,
   editor = '@motion-canvas/ui',
+  proxy,
 }: MotionCanvasPluginConfig = {}): Plugin => {
   const editorPath = path.dirname(require.resolve(editor));
   const editorFile = fs.readFileSync(path.resolve(editorPath, 'editor.html'));
@@ -89,6 +105,7 @@ export default ({
   const createHtml = (src: string) => htmlParts[0] + src + htmlParts[1];
 
   const resolvedEditorId = '\0virtual:editor';
+
   const timeStamps: Record<string, number> = {};
   const outputPath = path.resolve(output);
   const projects: ProjectData[] = [];
@@ -115,6 +132,9 @@ export default ({
       );
     }
   }
+
+  // Initialize the Proxy Module
+  setupEnvVarsForProxy(proxy);
 
   return {
     name: 'motion-canvas',
@@ -314,6 +334,14 @@ export default ({
 
         next();
       });
+
+      // if proxy is unset (undefined), or set to false,
+      // it will not register its middleware — as a result, no
+      // proxy is started.
+      if (proxy !== false && proxy !== undefined) {
+        motionCanvasCorsProxy(server.middlewares, proxy === true ? {} : proxy);
+      }
+
       server.ws.on('motion-canvas:meta', async ({source, data}, client) => {
         timeStamps[source] = Date.now();
         await fs.promises.writeFile(
--- a/packages/vite-plugin/src/proxy-middleware.ts
+++ b/packages/vite-plugin/src/proxy-middleware.ts
@@ -1,0 +1,291 @@
+/**
+ * This module provides the proxy located at
+ * /cors-proxy/...
+ *
+ * It is needed when accessing remote resources.
+ * Trying to access remote resources works while
+ * in preview, but will fail when you try to
+ * output the image (= "read" the canvas)
+ *
+ * See https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_enabled_image
+ * for reasons
+ *
+ * Using the proxy circumvents CORS-issues because
+ * this way all remote resources are served from the
+ * same host as the main app.
+ */
+
+import {Connect} from 'vite';
+import {ServerResponse, IncomingMessage} from 'http';
+import {http, https} from 'follow-redirects';
+
+/**
+ * Configuration used by the Proxy plugin
+ */
+export interface MotionCanvasCorsProxyOptions {
+  /**
+   * Set which types of resources are allowed by default.
+   *
+   * @remarks
+   * Catchall on the right side is supported.
+   * Pass an empty Array to allow all types of resources, although this is not
+   * recommended.
+   *
+   * @default ["image/*", "video/*"]
+   */
+  allowedMimeTypes?: string[];
+  /**
+   * Set which hosts are allowed
+   *
+   * @remarks
+   * Note that the host is everything to the left of the first `/`, and to the
+   * right of the protocol `https://`. AllowList is not used by default,
+   * although you should consider setting up just the relevant hosts.
+   */
+  allowListHosts?: string[];
+}
+
+export function setupEnvVarsForProxy(
+  config: MotionCanvasCorsProxyOptions | undefined | boolean,
+) {
+  // Define Keys for Env Var
+  const prefix = 'VITE_MC_PROXY_';
+  const isEnabledKey = prefix + 'ENABLED';
+  const allowList = prefix + 'ALLOW_LIST';
+
+  if (config === true) {
+    config = {}; // Use Default values
+  }
+
+  process.env[isEnabledKey] = String(!!config); // 'true' or 'false'
+
+  if (config) {
+    // These values are only configured if the Proxy is enabled
+    // You cannot access them via import.meta.env if the Proxy
+    // is set to false
+    process.env[allowList] = JSON.stringify(config.allowListHosts ?? []);
+  }
+}
+
+export function motionCanvasCorsProxy(
+  middleware: Connect.Server,
+  config: MotionCanvasCorsProxyOptions,
+) {
+  // Set the default config if no config was provided
+  config.allowedMimeTypes ??= ['image/*', 'video/*'];
+  config.allowListHosts ??= [];
+
+  // Check the Mime Types to have a correct structure (left/right)
+  // not having them in the correct format would crash the Middleware
+  // further down below
+  if ((config.allowedMimeTypes ?? []).some(e => e.split('/').length !== 2)) {
+    throw new Error(
+      "Invalid config for Proxy:\nAll Entries must have the following format:\n 'left/right' where left may be '*'",
+    );
+  }
+
+  middleware.use((req, res, next) => {
+    if (!req.url || !req.url.startsWith('/cors-proxy/')) {
+      // url does not start with /cors-proxy/, so this
+      // middleware does not care about it
+      return next();
+    }
+
+    // For now, only allow GET Requests
+    if (req.method !== 'GET') {
+      return writeError('Only GET Requests are allowed', res, 405);
+    }
+
+    let sourceUrl: URL;
+    try {
+      sourceUrl = extractDestination(req.url);
+    } catch (err) {
+      return writeError(err as any, res);
+    }
+
+    if (
+      !isReceivedUrlInAllowedHosts(sourceUrl.hostname, config.allowListHosts)
+    ) {
+      return writeError(
+        `Blocked by Proxy: ${sourceUrl.hostname} is not on Hosts Allowlist`,
+        res,
+      );
+    }
+
+    // Get the resource, do some checks. Throws an Error
+    // if the checks fail. The catch then writes an Error
+    return tryGetResource(res, sourceUrl, config).catch(error => {
+      writeError(error, res);
+    });
+  });
+}
+
+/**
+ * Unwrap the destination from the URL.
+ *
+ * @remarks
+ * Throws an Error if the value could not be unwrapped.
+ *
+ * @param url - the entire URL with the `/cors-proxy/` prefix and containing the
+ * url-Encoded Path.
+ *
+ * @returns The URL that needs to be called.
+ */
+function extractDestination(url: string): URL {
+  const withoutPrefix = url.replace('/cors-proxy/', '');
+  const asUrl = new URL(decodeURIComponent(withoutPrefix));
+  if (asUrl.protocol !== 'http:' && asUrl.protocol !== 'https:') {
+    throw new Error('Only supported protocols are http and https');
+  }
+  return asUrl;
+}
+
+/**
+ * A simple Error Helper that will write an Error and close the response.
+ */
+function writeError(
+  message: string,
+  res: ServerResponse<IncomingMessage>,
+  statusCode = 400,
+) {
+  res.writeHead(statusCode, message);
+  res.end();
+}
+
+/**
+ * Check if the Proxy is allowed to get the requested resource based on the
+ * host.
+ */
+function isReceivedUrlInAllowedHosts(
+  hostname: string,
+  allowListHosts: string[] | undefined,
+) {
+  if (!allowListHosts || allowListHosts.length == 0) {
+    // if the allowListHosts is just the predefinedAllowlist, the user has not
+    // set any additional hosts. In this case, allow any hostname
+    return true;
+  }
+  // Check if the hostname is any of the values set in allowListHosts
+  return allowListHosts.some(
+    e => e.toLowerCase().trim() === hostname.toLowerCase().trim(),
+  );
+}
+
+/**
+ * Check if the Proxy is allowed to get the requested resource based on the
+ * MIME-Type.
+ *
+ * @remarks
+ * Also handles catch-All Declarations like `image/*`.
+ */
+function isResultOfAllowedResourceType(
+  foundMimeType: string,
+  allowedMimeTypes: string[],
+) {
+  if (!allowedMimeTypes || allowedMimeTypes.length === 0) {
+    return true; // no filters set
+  }
+
+  if (foundMimeType.split('/').length !== 2) {
+    return false; // invalid mime structure
+  }
+
+  const [leftSegment, rightSegment] = foundMimeType
+    .split('/')
+    .map(e => e.trim().toLowerCase());
+
+  // Get all Segments where the left Part is identical between foundMimeType and
+  // allowedMimeType.
+  const leftSegmentMatches = allowedMimeTypes.filter(
+    e => e.trim().toLowerCase().split('/')[0] === leftSegment,
+  );
+  if (leftSegmentMatches.length === 0) {
+    // No matches at all, not even catchall - resource is rejected.
+    return false;
+  }
+
+  // This just gets the right part of the MIME Types from the
+  // configured allowList, e.g. "image/png" -> png
+  const rightSegmentOfLeftSegmentMatches = leftSegmentMatches.map(
+    e => e.split('/')[1],
+  );
+
+  // if an exact match or a catchall is found, the resource is allowed to be
+  // proxied.
+  return rightSegmentOfLeftSegmentMatches.some(
+    e => e === '*' || e === rightSegment,
+  );
+}
+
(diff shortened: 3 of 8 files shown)


packages/core/src/utils/viaProxy.ts
```diff
@@ -0,0 +1,57 @@
...
+ */
+//@ts-ignore
+import {enabled} from 'virtual:proxysettings';
+
+/**
+ * Helper-Function to route remote requests through a local proxy.
+ *
+ * @example This rewrites a remote url like `https://via.placeholder.com/300.png/09f/fff` into a URI-Component-Encoded string like `/cors-proxy/https%3A%2F%2Fvia.placeholder.com%2F300.png%2F09f%2Ffff`
+ */
+export function viaProxy(url: string, overrideProxySetting?: boolean) {
```
WaldemarLehner, line 20: I had to add `overrideProxySetting` for Unit Tests here. I could not manage to mock the virtual module properly.

It would only ever consider the first mocked value and ignore subsequent mocks. If you have an idea on how to get the mocks to update, please do let me know.

packages/core/src/utils/viaProxy.test.ts
```diff
@@ -0,0 +1,64 @@
...
...QpGtGyiQnu6H/iVo0qlaTdH+tIy+urw7XaeUqS7hmiOcPi1hdk6+lJwfbV27VSXZ/UyjvVP1+SfrZTL7ynx3dKUq9cPTe1s4+Q6tQhRujyqXuwTI/w5QPB1eT7R6hypirK1TdP+VFNGqC10/XoLcGyn++KO+q3bwcbv5+mx76s8kLlRzVlYOyfAw5+qnVHwq/vwAM3Bef0P08Gd8w8tVvbT0rS9MFafHN3TR1ihC6cumfL9H9lkjR/lF6s6GjBqmo9uCluT0TaPDu4QNdWszSzUn85Gn59e24o1t55Ks3VhQbd8qfY3V5jemvXXOVHde6yvviKDtUmf6hUpw4xQldN3eNl+jOhpDUHtHBT6//ouaJZWr5Xi9vcC9Ysff2N4BbkVhqatHhz63Mr1fUJRaTVd6k0V5Ie3R53v+U7Z4MXpcU5Wn1Xu5fmPivVqUOM0CVTZwKeCQPDS/TQOM0YrGElyoqo5ryqavTM/sTn0BXZWVr0Bc0fpbF9VZStE3WqqtEv92jP6a5Z38qPxmv57ZJUVaNp61vfjBKNaMscTRogSY9s0/K9nXrMVKcOMcLnnDoTECFgxstRwIwIATMiBMyIEDAjQsCMCAEzIgTMiBAwI0LAjAgBMyIEzIgQMCNCwIwIATMiBMyIEDAjQsCMCAEzIgTMiBAwI0LAjAgBMyIEzIgQMCNCwIwIATMiBMyIEDAjQsCMCAEzIgTMiBAwI0LAjAgBMyIEzIgQMCNCwIwIATMiBMyIEDAjQsCMCAEzIgTMiBAwI0LAjAgBMyIEzIgQMCNCwIwIATMiBMyIEDAjQsCMCAEzIgTMiBAwI0LAjAgBMyIEzIgQMCNCwIwIATMiBMyIEDAjQsCMCAEzIgTMiBAwI0LAjAgBMyIEzIgQMCNCwIwIATMiBMyIEDAjQsCMCAEzIgTMiBAwI0LAjAgBMyIEzIgQMCNCwIwIATMiBMyIEDAjQsCMCAEzIgTMiBAwI0LAjAgBMyIEzIgQMCNCwIwIATMiBMyIEDAjQsCMCAEzIgTMiBAwI0LAjAgBMyIEzIgQMCNCwIwIATMiBMyIEDAjQsCMCAEzIgTMiBAwI0LAjAgBMyIEzIgQMCNCwIwIATMiBMyIEDAjQsCMCAEzIgTMiBAwI0LAjAgBMyIEzIgQMCNCwIwIATMiBMz+Dzt6qG+Jr4RtAAAAAElFTkSuQmCC';
```
WaldemarLehner, line 51: Should I trim this? 

I dont think the actual payload is relevant for the unit test.

exdeejay: I'd say just put in a smaller image, i.e. 1x1 pixels

packages/core/src/utils/viaProxy.ts
exdeejay, line 20: This method should probably be called `useProxy()` to fit in with the naming convention for the other utility functions.

packages/core/src/utils/viaProxy.ts
```diff
@@ -0,0 +1,57 @@
...
+  const isEnabled = overrideProxySetting ?? enabled;
+  if (!isEnabled) {
+    // Proxy is disabled, so we just pass as-is.
```
exdeejay, line 23: Print a warn message here to ensure the user knows the proxy is disabled.

packages/core/src/utils/viaProxy.ts
```diff
@@ -0,0 +1,57 @@
...
+      return url;
+    }
+    if (selfUrl.host === expandedUrl.host) {
+      // This is a request to a "local" resource.
+      // No need to rewrite
+      return url;
+    }
+  } catch (_) {
+    // in case of error just silently pass as-is
+    return url;
```
exdeejay, line 52: IMO a malformed or non-HTTP(S) URL should have an associated warning that the URL isn't being proxied, but maybe that's too verbose.

packages/vite-plugin/src/main.ts
```diff
@@ -79,6 +93,7 @@ export default ({
   output = './output',
   bufferedAssets = /\.(wav|ogg)$/,
   editor = '@motion-canvas/ui',
+  proxy = undefined,
```
exdeejay, line 96: Remove `= undefined`, it's implied

packages/vite-plugin/src/main.ts
```diff
@@ -121,6 +138,11 @@ export default ({
     async configResolved(resolvedConfig) {
       viteConfig = resolvedConfig;
     },
+    resolveId(id) {
+      if (id === proxySettingsKey) {
+        return '\0' + proxySettingsKey;
```
exdeejay, line 143: Extract this and the one below out into a constant, like `resolvedEditorId` (build it from proxySettingsKey though, you don't have to repeat yourself)

packages/vite-plugin/src/main.ts
```diff
@@ -67,6 +71,16 @@ export interface MotionCanvasPluginConfig {
...
+  /**
+   * Configuration of the Proxy used for remote sources
+   *
+   * @remarks
+   * This passes configuration to Motion Canvas' proxy.
+   * Not setting this value will allow all `image/*` and `video/*` resources from any hostname.
+   * You can also disable the proxy entirely by passing `false`. This will however mean that you
+   * cannot render any animation that relies on remote sources.
+   */
+  proxy?: false | MotionCanvasCorsProxyPluginOptions;
```
exdeejay, line 83: This should be able to handle `true` as well, treat it the same as `undefined`

packages/vite-plugin/src/proxy-middleware.ts
```diff
@@ -0,0 +1,247 @@
...
+ * Trying to access remote resources works while
+ * in preview, but will fail when you try to
+ * output the image (= "read" the canvas)
+ *
+ * See https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_enabled_image
+ * for reasons
+ *
+ * Using the proxy circumvents CORS-issues because
+ * this way all remote resources are served from the
+ * same server (localhost:9000) as the main app.
```
exdeejay, line 15: nit: `localhost:9000` doesn't have to be mentioned here since it's not guaranteed to be the hostname, say "from the same domain as the main app" instead

packages/vite-plugin/src/proxy-middleware.ts
```diff
@@ -0,0 +1,247 @@
...
+   * Pass an empty Array to allow all types of resources, although this is not recommended.
+   *
+   * @default ["image/*", "video/*"]
+   */
+  allowedMimeTypes?: string[];
+  /**
+   * Set which hosts are allowed
+   *
+   *
+   * @remarks Note that the host if everything to the left of the first `/`, and to the right of the protocol `https://`
```
exdeejay, line 40: *"host is"

packages/vite-plugin/src/proxy-middleware.ts
```diff
@@ -0,0 +1,247 @@
...
+
+/**
+ * Unwrap the destination from the URL.
+ *
+ * @param url - the entire URL with the `/cors-proxy/` prefix and containing the url-Encoded Path
+ * @returns the URL that needs to be called
+ */
+function extractDestination(
+  url: string,
+): [URL, undefined] | [undefined, string] {
```
exdeejay, line 116: Don't return a result/error pattern here, just throw the error and catch it where you use it up above.

packages/vite-plugin/src/proxy-middleware.ts
```diff
@@ -0,0 +1,247 @@
...
+      // middleware does not care about it
+      return next();
+    }
+
+    // For now, only allow GET Requests
+    if (req.method !== 'GET') {
+      return writeError('Only GET Requests are allowed', res, 405);
+    }
+
+    const [sourceUrl, error] = extractDestination(req.url);
```
exdeejay, line 86: Wrap this in a try/catch (see below)

exdeejay (changes requested):

exdeejay:
Possible idea: in Image.ts, check if the proxy is disabled and if so, set `img.crossOrigin = 'anonymous'` so that cross-origin images will still work.

packages/core/src/utils/viaProxy.ts
WaldemarLehner: My main motivation behind not using the `use`-Prefix is that, looking at the other utils, `use*` is used for getting some sort of dependency.

`useRandom` returns a Random-Instance, `useThread` returns a reference to the thread, etc.
`useProxy` would, in my opinion, mean that you would get some sort of Proxy-Instance.

packages/core/src/utils/viaProxy.ts
WaldemarLehner: i think that the Browser can do that for us — if it is indeed bad value.
Maybe there is some non-standard way of providing a resource that URL does not handle — in this case, the value should probably not be proxied anyways.

If the Browser cannot understand the passed value it will log to the console, if it can, it will interpret it correctly.

packages/core/src/utils/viaProxy.ts
exdeejay: `useProxy` is kind of returning a dependency, since it's transforming an image URL into a project-dependant and usable URL. Also, this project is heavily based on the architecture of React. In React, hook function start with `use` to signify that they aren't pure, and that you're depending on a functionality of React. In this case, you'd be depending on the proxy functionality, so I'd argue the `use` naming scheme would be appropriate.

packages/core/src/utils/viaProxy.ts
WaldemarLehner: I don't think this is necessary as disabling the proxy is a conscious decision by the user (setting it to `false` in the vite config)

packages/core/src/utils/viaProxy.test.ts
exdeejay: I must have accidentally deleted my comment earlier on this: you could keep it and replace it with a 1x1 image so it's not so long

packages/core/src/utils/viaProxy.ts
exdeejay: True, disregard then

packages/core/src/utils/viaProxy.test.ts
WaldemarLehner: oh… good point!

packages/core/src/utils/viaProxy.ts
aarthificial: I agree with @WaldemarLehner. The `use` prefix is used for hooks - things that can only be called in the context of the animation. Calling `useScene` outside of the animation (for example in a promise) will fail. The same goes for other hooks. `viaProxy` is just a simple wrapper function that can be called anywhere.

packages/core/src/utils/viaProxy.ts
```diff
@@ -0,0 +1,57 @@
...
+ * This utility is used to rewrite a request to be routed through
+ * the Proxy instead.
+ */
+
+/**
+ * Virtual Module provided by the Motion Canvas
+ * Plugin. Returns `false` if proxy is set to `false`, else `true`
+ */
+//@ts-ignore
+import {enabled} from 'virtual:proxysettings';
```
aarthificial, line 13: I'm not sure how I feel about using virtual modules in `core`.
So far, all integration with Vite was done through `import.meta` so it was possible to use `core` like a normal typescript library and build it without Vite. But with virtual modules, one will need to configure their pipeline specifically to account for the `virtual:proxysettings` alias which sounds annoying.

Can't we use `import.meta.env` or something along these lines to pass the configuration?

packages/vite-plugin/src/proxy-middleware.ts
```diff
@@ -0,0 +1,244 @@
...
+ */
+
+import {Connect} from 'vite';
+import axios from 'axios';
+import {ServerResponse, IncomingMessage} from 'http';
+
+/**
+ * Configuration used by the Proxy plugin
+ */
+export interface MotionCanvasCorsProxyPluginOptions {
```
aarthificial, line 25: It should be called `MotionCanvasCorsProxyOptions` since it's not technically a standalone plugin. It's part of the main Vite plugin.

packages/vite-plugin/src/proxy-middleware.ts
```diff
@@ -0,0 +1,244 @@
+/**
+ * This modules provides the proxy located at
+ * /motion-canvas-proxy/...
```
aarthificial, line 3:
```suggestion
 * This module provides the proxy located at
 * /cors-proxy/...
```

packages/vite-plugin/src/proxy-middleware.ts
```diff
@@ -0,0 +1,244 @@
...
+ *
+ * It is needed when acessing remote resources.
```
aarthificial, line 5:
```suggestion
 * It is needed when accessing remote resources.
```

packages/vite-plugin/src/proxy-middleware.ts
```diff
@@ -0,0 +1,244 @@
...
+ * See https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_enabled_image
+ * for reasons
+ *
+ * Using the proxy circumvents CORS-issues because
+ * this way all remote resources are served from the
+ * same host as the main app.
+ */
+
+import {Connect} from 'vite';
+import axios from 'axios';
```
aarthificial, line 19: `axios` is a dependency of the `docs` package. It works for you locally, because in a monorepo dependencies are shared in root node_modules. After publishing, this will fail because npm won't install `axios`.

`axios` is a quite heavy dependency and I'm not sure if we should add it to `vite-plugin`. Can't we just use Node's `https` package?

packages/vite-plugin/src/proxy-middleware.ts
```diff
@@ -0,0 +1,244 @@
...
+/**
+ * Configuration used by the Proxy plugin
+ */
+export interface MotionCanvasCorsProxyPluginOptions {
+  /**
+   * Set which types of resources are allowed by default.
+   *
+   * @remarks
+   * Catchall on the right side is supported.
+   * Pass an empty Array to allow all types of resources, although this is not recommended.
```
aarthificial, line 31: Prettier won't do this for you but comments should wrap at 80 characters too.
```suggestion
   * Pass an empty Array to allow all types of resources, although this is not 
   * recommended.
```

packages/vite-plugin/src/proxy-middleware.ts
```diff
@@ -0,0 +1,244 @@
...
+  allowedMimeTypes?: string[];
+  /**
+   * Set which hosts are allowed
+   *
+   *
+   * @remarks Note that the host is everything to the left of the first `/`, and to the right of the protocol `https://`
+   * Allowlist is not used by default, although you should consider setting up just the relevant hosts.
+   *
+   * Note that `api.iconify.design` is **always** enabled, as it is used by the Icon-Component.
+   * It gets added by the Middleware internally.
```
aarthificial, line 44:
```suggestion
   * @remarks 
   * Note that the host is everything to the left of the first `/`, and to the 
   * right of the protocol `https://`. AllowList is not used by default, 
   * although you should consider setting up just the relevant hosts.

   * Note that `api.iconify.design` is **always** enabled, as it is used by the 
   * Icon-Component. It gets added by the Middleware internally.
```

packages/vite-plugin/src/proxy-middleware.ts
```diff
@@ -0,0 +1,244 @@
...
+   */
+  allowListHosts?: string[];
+}
+
+// Add Hosts required for core functionality to work
+const predefinedAllowlist = [
```
aarthificial, line 50:
```suggestion
const predefinedAllowList = [
```

packages/vite-plugin/src/proxy-middleware.ts
```diff
@@ -0,0 +1,244 @@
...
+  // Icons
+  'api.iconify.design',
```
aarthificial, line 52: iconify is not necessary with `crossOrigin = 'anonymous'`

packages/vite-plugin/src/proxy-middleware.ts
```diff
@@ -0,0 +1,244 @@
...
+  });
+}
+
+/**
+ * Unwrap the destination from the URL.
+ *
+ * @param url - the entire URL with the `/cors-proxy/` prefix and containing the url-Encoded Path
+ * @returns the URL that needs to be called
+ *
+ * @remarks Throws an Error if the value could not be unwrapped.
```
aarthificial, line 116: Remarks should always come second.
Also missing commas/capitalization.
```suggestion
 * Unwrap the destination from the URL.
 *
 * @remarks 
 * Throws an Error if the value could not be unwrapped.
 *
 * @param url - the entire URL with the `/cors-proxy/` prefix and containing the
 * url-Encoded Path.
 *
 * @returns The URL that needs to be called.
```

packages/vite-plugin/src/proxy-middleware.ts
```diff
@@ -0,0 +1,244 @@
...
+ */
+function extractDestination(url: string): URL {
+  const withoutPrefix = url.replace('/cors-proxy/', '');
+  const asUrl = new URL(decodeURIComponent(withoutPrefix));
+  if (asUrl.protocol !== 'http:' && asUrl.protocol !== 'https:') {
+    throw 'Only supported protocols are http and https';
```
aarthificial, line 122:
```suggestion
    throw new Error('Only supported protocols are http and https');
```

packages/vite-plugin/src/proxy-middleware.ts
```diff
@@ -0,0 +1,244 @@
...
+  }
+  return asUrl;
+}
+
+/**
+ * A simple Error Helper that will write an Error and close
+ * the response
```
aarthificial, line 129:
```suggestion
 * A simple Error Helper that will write an Error and close the response.
```

packages/vite-plugin/src/proxy-middleware.ts
```diff
@@ -0,0 +1,244 @@
...
+  message: string,
+  res: ServerResponse<IncomingMessage>,
+  statusCode = 400,
+) {
+  res.writeHead(statusCode, message);
+  res.end();
+}
+
+/**
+ * Check if the Proxy is allowed to get the requested resource based on the host
```
aarthificial, line 141:
```suggestion
 * Check if the Proxy is allowed to get the requested resource based on the 
 * host.
```

packages/vite-plugin/src/proxy-middleware.ts
```diff
@@ -0,0 +1,244 @@
...
+  return allowListHosts.some(
+    e => e.toLowerCase().trim() === hostname.toLowerCase().trim(),
+  );
+}
+
+/**
+ * Check if the Proxy is allowed to get the requested resource based on the MIME-Type
+ *
+ * @remarks
+ * Also handles catch-All Declarations like image/*
```
aarthificial, line 162:
```suggestion
 * Check if the Proxy is allowed to get the requested resource based on the 
 * MIME-Type.
 *
 * @remarks
 * Also handles catch-All Declarations like `image/*`.
```

packages/vite-plugin/src/proxy-middleware.ts
```diff
@@ -0,0 +1,244 @@
...
+
+  if (foundMimeType.split('/').length !== 2) {
+    return false; // invalid mime structure
+  }
+
+  const [leftSegment, rightSegment] = foundMimeType
+    .split('/')
+    .map(e => e.trim().toLowerCase());
+
+  // Get all Segments where the left Part is identical between foundMimeType and allowedMimeType
```
aarthificial, line 180:
```suggestion
  // Get all Segments where the left Part is identical between foundMimeType and 
  // allowedMimeType.
```

packages/vite-plugin/src/proxy-middleware.ts
```diff
@@ -0,0 +1,244 @@
...
+  const leftSegmentMatches = allowedMimeTypes.filter(
+    e => e.trim().toLowerCase().split('/')[0] === leftSegment,
+  );
+  if (leftSegmentMatches.length === 0) {
+    return false; // No matches at all, not even catchall - resource is rejected
```
aarthificial, line 185:
```suggestion
    // No matches at all, not even catchall - resource is rejected.
    return false; 
```

packages/vite-plugin/src/proxy-middleware.ts
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  // This just gets the right part of the MIME Types from the
+  // configured allowList, e.g. "image/png" -> png
+  const rightSegmentOfLeftSegmentMatches = leftSegmentMatches.map(
+    e => e.split('/')[1],
+  );
+
+  // if an exact match, or a catchall is found, the resource is allowed to be proxied.
```
aarthificial, line 194:
```suggestion
  // if an exact match or a catchall is found, the resource is allowed to be 
  // proxied.
```

aarthificial (changes requested):

packages/vite-plugin/src/proxy-middleware.ts
WaldemarLehner: We can! My thought behind using axios was that I would not need to handle all the edge cases, like HTTP Redirects.
I should be able to get rid of the dependency though

packages/core/src/utils/viaProxy.ts
WaldemarLehner: Just gave this a test…
Writing an env var during plugin initialization works.
image


There is also the option to add the string into the config object, but I find that to be a bad solution as Vite just replaces all occurrences of the String, regardless of Context (e.g. variables names as the key get expanded to the value; but expansion also happens when inside a string) 

Only downside: The values can only be Strings. But that is not an issue for this use case

WaldemarLehner: so core could have an import like 

```ts
const proxyEnabled = import.meta?.env?.VITE_MC_ENABLE_PROXY ?? "false" !== "false"
``` 

No virtual module, and a safe fallback if the context doesnt  "know" about the env var

packages/vite-plugin/src/proxy-middleware.ts
WaldemarLehner: @aarthificial 

How would you feel about depending on `follow-redirects` ?

https://github.com/follow-redirects/follow-redirects

It is [28.3 kB unpacked](https://www.npmjs.com/package/follow-redirects) according to npm  while [axios is 1.72MB](https://www.npmjs.com/package/axios)

packages/vite-plugin/src/proxy-middleware.ts
WaldemarLehner: The initial plan was to just reject requests going to remote addresses.

But with anon crossOrigin we can just not use the proxy if it doesn't match the allow list.

I guess the `predefinedAllowList` could be removed entirely in this case.

packages/vite-plugin/src/proxy-middleware.ts
aarthificial: I'm okay with that

packages/core/src/utils/viaProxy.ts
WaldemarLehner: no longer relevant

packages/core/src/utils/proxyUtils.ts
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * This utility is used to rewrite a request to be routed through
+ * the Proxy instead.
+ */
+
+/**
+ * Helper-Function to route remote requests through a local proxy.
+ *
+ * @example This rewrites a remote url like `https://via.placeholder.com/300.png/09f/fff` into a URI-Component-Encoded string like `/cors-proxy/https%3A%2F%2Fvia.placeholder.com%2F300.png%2F09f%2Ffff`
+ */
```
aarthificial, line 12: Function descriptions should use the imperative present tense.
Also wrapping at 80 characters:
```suggestion
/**
 * Route the given url through a local proxy.
 *
 * @example
 * This rewrites a remote url like `https://via.placeholder.com/300.png/09f/fff` 
 * into a URI-Component-Encoded string like 
 * `/cors-proxy/https%3A%2F%2Fvia.placeholder.com%2F300.png%2F09f%2Ffff`
 */
```

packages/core/src/utils/proxyUtils.ts
```diff
@@ -0,0 +1,146 @@
...
+ * exists inside the Vite Server.
+ */
+export function isProxyEnabled() {
+  try {
+    const importedValue = import.meta.env['VITE_MC_PROXY_ENABLED'] ?? 'false';
+    return importedValue === 'true';
+  } catch (_) {
+    return false;
+  }
+}
```
aarthificial, line 88: The `try/catch` here shouldn't be necessary:
```suggestion
export function isProxyEnabled() {
  if (import.meta.env) {
    return import.meta.env.VITE_MC_PROXY_ENABLED === 'true';
  } else {
    return false;
  }
}
```
`import.meta` is a js standard so we can always access it to check if `env` exists.
This will also result in a cleaner output after minification:

```ts
// with `try/catch`:
function or(){try{return"true"==="true"}catch{return!1}}

// with `if`:
function or(){return!0}
```

I also don't know why this bracket notation is necessary:
```ts
import.meta.env['VITE_MC_PROXY_ENABLED']
```
I used a dot notation and it works fine:
```ts
import.meta.env.VITE_MC_PROXY_ENABLED
```
Vite replaces it when building, tests in Vitest pass. Is there any reason for it?

packages/core/src/utils/proxyUtils.ts
```diff
@@ -0,0 +1,146 @@
...
+              '. Expected a String. Skipping.',
+          );
+          continue;
+        }
+        validatedEntries.push(entry);
+      }
+
+      return validatedEntries;
+    } catch (err) {
+      console.error('Failed to parse Allow List from Proxy');
```
aarthificial, line 140:
```suggestion
      useLogger().error('Failed to parse Allow List from Proxy');
```

packages/core/src/utils/proxyUtils.ts
```diff
@@ -0,0 +1,146 @@
...
+      return [...getAllowListCache];
+    }
+  }
+
+  const result = (function () {
+    if (!isProxyEnabled()) {
+      return [];
+    }
+
+    try {
```
aarthificial, line 113: Same thing as in my other comment

aarthificial (changes requested):

packages/core/src/utils/proxyUtils.ts
WaldemarLehner: > Vite replaces it when building, tests in Vitest pass. Is there any reason for it?

Oh . None at all. Dot notation should work fine

aarthificial (approved):

PR #361, opened 2023-02-16
docs: fix #319: hack a fix for bad TOC 

### LINKED ISSUE #319: Doc: Reloading API Reference with unchecked "Inherited members" produces bad TOC
**Describe the bug**
Reloading any of the autogenerated API Reference Pages that have inherited members while having *Inherited members* unchecked in the filters results in wrong Properties on the TOC on the right.

**To Reproduce**

1. Go to any API Reference that inherits Properties / Methods, e.g. https://motion-canvas.github.io/api/2d/components/Rect/
2. By default, inherited Properties / Methods are shown: 
image
3. Uncheck Inherited Members on the Filters: 
image
4. Properties show up correctly on the right
5. Reloading the page results in a bad TOC: image
6. Checking and unchecking Inherited Members does *not* fix the issue.

**Expected behavior**
After Reloading the page the TOC should remain as in Point 3

**Console errors**
no output

v2.3.0

**Additional context**
Browsers used:
Mozilla Firefox 109.0.1 (64-bit), Manjaro Linux
Chromium 109.0.5414.119 (Official Build) Arch Linux (64-bit)

WaldemarLehner:
This rewrites the handler for auto-generated Docs to render out the entire TOC first before rendering just the filtered TOC, which is needed because of some Quirks on Docusaurus' end.

Closes #319

### DIFF
--- a/packages/docs/src/components/Api/ApiItem.tsx
+++ b/packages/docs/src/components/Api/ApiItem.tsx
@@ -10,6 +10,7 @@ import {TOCItem} from '@docusaurus/mdx-loader';
 import Tooltip from '@site/src/components/Tooltip';
 import type {JSONOutput} from 'typedoc';
 import {ReflectionKind} from './ReflectionKind';
+import useIsBrowser from '@docusaurus/useIsBrowser';
 
 interface ApiItemProps {
   route: {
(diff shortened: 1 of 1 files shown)


packages/docs/src/components/Api/ApiItem.tsx
```diff
@@ -23,6 +24,11 @@ export default function ApiItem({route}: ApiItemProps): JSX.Element {
   const reflection: JSONOutput.DeclarationReflection =
     lookup[route.reflectionId];
   const [filters] = useFilters();
+  // This is needed because of some weird internal logic on
+  // Docusaurus' side of things.
+  // If you immediately just pass the filtered values
+  // the TOC will have wrong values mapped to it
+  const [passFilters, setPassFilters] = useState(false);
```
aarthificial, line 31: I did a quick test and it looks like you should be able to use `isBrowser` directly, without the `useEffect` hook.
Just specify it as a dependency of `useMemo` and do:
```ts
(isBrowser && !matchFilters(filters, child))
```

aarthificial (approved):

PR #365, opened 2023-02-17
feat(2d): add moveBelow, moveAbove and moveTo methods to Node

ksassnowski:
This PR adds three methods to the `Node` base class to more easily position a node relative to a specific other node in its parent's hierarchy.

- `moveTo` moves to node to the specified position in the parent's children array
- `moveBelow` moves the node one position below the provided sibling node. Will not move the node if it is already positioned lower than the sibling. This method logs an error if the provided node is not a sibling of the current node.
- `moveAbove` moves the node one position above the provided sibling node. Will not move the node if it is already positioned higher than the sibling. This method logs an error if the provided node is not a sibling of the current node.

### DIFF
--- a/packages/2d/src/components/Node.ts
+++ b/packages/2d/src/components/Node.ts
@@ -21,7 +21,11 @@ import {
   ColorSignal,
   SimpleVector2Signal,
 } from '@motion-canvas/core/lib/types';
-import {DetailedError, ReferenceReceiver} from '@motion-canvas/core/lib/utils';
+import {
+  DetailedError,
+  ReferenceReceiver,
+  useLogger,
+} from '@motion-canvas/core/lib/utils';
 import type {ComponentChild, ComponentChildren, NodeConstructor} from './types';
 import {Promisable} from '@motion-canvas/core/lib/threading';
 import {useScene2D} from '../scenes/useScene2D';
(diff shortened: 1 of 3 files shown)


packages/2d/src/components/Node.ts
```diff
@@ -731,6 +735,115 @@ export class Node implements Promisable<Node> {
...
+  /**
+   * Move the node below the provided node in the parent's layout.
+   *
+   * @remarks
+   * The node will be moved below the provided node and from then on will be rendered
+   * below it. By default, if the node is already positioned lower than the sibling
+   * node, it will not get moved.
+   *
+   * @param node - The sibling node below which to move.
+   * @param directlyAbove - Whether the node should be positioned directly below the
```
aarthificial, line 770:
```suggestion
   * @param directlyBelow - Whether the node should be positioned directly below the
```

packages/2d/src/components/Node.ts
```diff
@@ -731,6 +735,115 @@ export class Node implements Promisable<Node> {
...
+   *                        sibling. When true, will move the node even if it is
+   *                        already positioned below the sibling.
```
aarthificial, line 772: 80 characters line wrapping:
```suggestion
   * Move the node below the provided node in the parent's layout.
   *
   * @remarks
   * The node will be moved below the provided node and from then on will be
   * rendered below it. By default, if the node is already positioned lower than
   * the sibling node, it will not get moved.
   *
   * @param node - The sibling node below which to move.
   * @param directlyBelow - Whether the node should be positioned directly below
   *                        the sibling. When true, will move the node even if
   *                        it is already positioned below the sibling.
```

packages/2d/src/components/Node.ts
```diff
@@ -731,6 +735,115 @@ export class Node implements Promisable<Node> {
...
+   *
+   * @remarks
+   * The node will be moved above the provided node and from then on will be rendered
+   * on top of it. By default, if the node is already positioned higher than the sibling
+   * node, it will not get moved.
+   *
+   * @param node - The sibling node below which to move.
+   * @param directlyAbove - Whether the node should be positioned directly above the
+   *                        sibling. When true, will move the node even if it is
+   *                        already positioned above the sibling.
```
aarthificial, line 815:
```suggestion
   * Move the node above the provided node in the parent's layout.
   *
   * @remarks
   * The node will be moved above the provided node and from then on will be 
   * rendered on top of it. By default, if the node is already positioned higher 
   * than the sibling node, it will not get moved.
   *
   * @param node - The sibling node below which to move.
   * @param directlyAbove - Whether the node should be positioned directly above 
   *                        the sibling. When true, will move the node even if 
   *                        it is already positioned above the sibling.
```

aarthificial (changes requested):
References to these new methods should also be added to the docs:
https://github.com/motion-canvas/motion-canvas/blob/2025adc6e7aa11d81b6f5f6989e8eae18cf86cb7/packages/docs/docs/getting-started/hierarchy.mdx?plain=1#L124-L132

PR #379, opened 2023-02-18
feat(2d): make `View2D` extend `Rect`

### LINKED ISSUE #367: Ability to change background color from within the scenes
I request the ability to be able to change the background color of the main project from within the scene, using the Tweening methods and whatnot instead of having to rely on big rectangles OVER the background.

Example Below:
image

dariasc:
Closes #367.

Worked locally for me, unsure if it could cause issues elsewhere but examples seem fine :D.

### DIFF
--- a/packages/2d/src/components/View2D.ts
+++ b/packages/2d/src/components/View2D.ts
@@ -1,6 +1,6 @@
-import {Layout, LayoutProps} from './Layout';
+import {Rect, RectProps} from './Rect';
 
-export class View2D extends Layout {
+export class View2D extends Rect {
   public static frameID = 'motion-canvas-2d-frame';
   public static shadowRoot: ShadowRoot;
 
@@ -20,7 +20,7 @@ export class View2D extends Layout {
     View2D.shadowRoot = frame.shadowRoot ?? frame.attachShadow({mode: 'open'});
   }
 
-  public constructor(props: LayoutProps) {
+  public constructor(props: RectProps) {
     super({
       composite: true,
       fontFamily: 'Roboto',


strau0106:
This doesn't work. I'm getting an error that a lerp from null to `#ffffff` isn't possible.

I guess this means that the fill property needs to be set in the constructor...

`const circle = createRef<Circle>();

  view.add(
  );

  yield* waitUntil('circle');
  yield* circle().scale(2, 2).to(1, 2);
  yield* waitUntil('color');
  yield* view.fill('white', 1).to('black', 2);
  yield* waitFor(5);`

aarthificial:
Awesome, thanks for the PR!

ribru17:
I tried this solution and when tweening the fill color of the view to change the background I got an error saying `Attempting to lerp null -> #fff000 may result in unexpected behavior.` and a following message 
```
unknown format:
The problem occurred here:

17 |   yield* view.fill('#fff000', 1)
   |          ^
18 |   yield* view.fill('black', 2)
19 | 
```
This also prevented the scene from running.
Naively I tried setting an initial value for the `fill` property from within the `View2D` constructor but this caused other issues with the tweening that I failed to resolve. In order to get everything working properly I always have to call `view.fill('#141414')` (or whichever background color) manually at the beginning of the `makeScene2D` function in order to get tweening to work (mostly: even then tweening from yellow #FFF000 to black #000000 doesn't appear to work properly, it tweens to a dark red and then has a discrete jump to the next color). Just thought I would mention this if anyone comes across the same issue, or maybe someone can point me towards a solution if I am doing something wrong

strau0106:
This was already discussed in the discord: https://discord.com/channels/1071029581009657896/1076166997513093272/1076793321315377244

The problem you are experiencing is related to the fact that tweening between different types (null, color, gradient, etc.) And will be a no fix. You just have to set the inital value with a .fill("color") without tweening. 

It will be addressed for the next major release AFAIK

PR #384, opened 2023-02-19
Add simple fade transition

### LINKED ISSUE #383: feat req: Add fade transitions
**Description**
Add two fade transitions: 
- A simple fade which simply fades in the new scene on top of the old one
- A fade through color, which fades the previous scene to a solid color and then that color to the new scene

**Proposed solution**
Create builtin helper methods similar to `slideTransition` and `zoom[In,Out]Transition`:
- `fadeTransition(duration)`
- `colorFadeTransition(color = "black", duration)`

DanielBreiner:
Closes #383

### DIFF
--- a/packages/core/src/transitions/fadeTransition.ts
+++ b/packages/core/src/transitions/fadeTransition.ts
@@ -1,0 +1,18 @@
+import {useTransition} from './useTransition';
+import {ThreadGenerator} from '../threading';
+import {createSignal} from '../signals';
+
+/**
+ * Perform a transition that fades between the scenes.
+ *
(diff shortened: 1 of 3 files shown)


aarthificial (changes requested):
The new transition should also be mentioned in the docs:
https://github.com/motion-canvas/motion-canvas/blob/d8a408dc231cef6a7b80744bd9bc985144778b03/packages/docs/docs/getting-started/transitions.mdx?plain=1#L71-L86

DanielBreiner:
Doesn't close #383, only partially addresses it.

Issue #380, opened 2023-02-19
Customizable shape for the Line Arrowheads


zweek:
**Description**
Right now there seems to be no way to customize the appearance of the arrowheads (beyond the `arrowSize` property), so they are always drawn as the same right triangle shape

**Proposed solution**
To have maybe 2 extra properties - one for 'sharpness' that controls the angle at the tip and one radius property for rounding the corners of the triangle. That way you could tweak those properties to get an arrowhead that could look like the one in the screenshot below

**Additional context**
![image](https://user-images.githubusercontent.com/70382336/219942849-aa5b2ed3-ab90-479b-95ce-fd26098dec6d.png)

strau0106:
I'd propose that you can pass another shape object as the Arrow Head. Because otherwise we would reimplement features that already exist.

zweek:
Yeah I like that idea a lot more, actually. That also opens up a lot more possibilities for what you can do with Lines if you can pass in whatever shape you want

strau0106:
Actually not a Shape but a Node in general, considering that right now triangles and other non 4-edged shapes are made with lines

aarthificial:
`Line` extends `Shape` but I agree that we should accept any `Node`.

I propose extending the `startArrow` `endArrow` properties to `Node | bool`.
The `Node.draw` method can be used to reuse the drawing logic of a Node. This method assumes that the canvas' matrix has been already transformed to the node's local space (`[0, 0]` being the center of the node). This way we won't need to reparent anything.

strau0106:
Sounds good! 

I will be claiming this issue if that‘s okay for everyone!

strau0106:
The problem I'm seeing, is that the draw function on a Node is protected. Thus I can't call it from the drawShape function in Line for the respective ArrowHeads

PR #389, opened 2023-02-20
feat(2d): custom arrow heads for Lines (Curves) (#380)

strau0106:
Hi Everybody, 

This PR allows custom arrow heads for lines via signals, as requested in #380. 

Best, 

Strau

### DIFF
--- a/packages/2d/src/components/Curve.ts
+++ b/packages/2d/src/components/Curve.ts
@@ -216,26 +216,22 @@ export abstract class Curve extends Shape {
     context.save();
     context.beginPath();
     if (this.endArrow() instanceof Node) {
-      const endArrow = this.endArrow() as Node;
-      context.save();
-      context.translate(endPoint.x, endPoint.y);
-      context.rotate(endTangent.radians);
-
-      endArrow.render(context);
-
-      context.restore();
+      this.drawArrowNode(
+        context,
+        this.endArrow() as Node,
+        endPoint,
+        endTangent.radians,
+      );
     } else if (this.endArrow()) {
       this.drawArrow(context, endPoint, endTangent, arrowSize);
     }
     if (this.startArrow() instanceof Node) {
-      const startArrow = this.startArrow() as Node;
-      context.save();
-      context.translate(startPoint.x, startPoint.y);
-      context.rotate(startTangent.radians);
-
-      startArrow.render(context);
-
-      context.restore();
+      this.drawArrowNode(
+        context,
+        this.startArrow() as Node,
+        startPoint,
+        startTangent.radians,
+      );
     } else if (this.startArrow()) {
       this.drawArrow(context, startPoint, startTangent, arrowSize);
     }
@@ -259,5 +255,20 @@ export abstract class Curve extends Shape {
     lineTo(context, origin.add(tangent.sub(normal).scale(arrowSize)));
     lineTo(context, origin);
     context.closePath();
+  }
+
+  private drawArrowNode(
+    context: CanvasRenderingContext2D,
+    arrow: Node,
+    arrowPoint: Vector2,
+    arrowAngle: number,
+  ) {
+    context.save();
+    context.translate(arrowPoint.x, arrowPoint.y);
+    context.rotate(arrowAngle);
+
+    arrow.render(context);
+
+    context.restore();
   }
 }
--- a/packages/2d/src/components/Curve.ts
+++ b/packages/2d/src/components/Curve.ts
@@ -18,10 +18,10 @@ export interface CurveProps extends ShapeProps {
   closed?: SignalValue<boolean>;
   start?: SignalValue<number>;
   startOffset?: SignalValue<number>;
-  startArrow?: SignalValue<boolean>;
+  startArrow?: SignalValue<Node | boolean>;
   end?: SignalValue<number>;
   endOffset?: SignalValue<number>;
-  endArrow?: SignalValue<boolean>;
+  endArrow?: SignalValue<Node | boolean>;
   arrowSize?: SignalValue<number>;
 }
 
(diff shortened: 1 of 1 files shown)


packages/2d/src/components/Line.ts
```diff
@@ -251,10 +263,18 @@ export class Line extends Shape {
 
     context.save();
     context.beginPath();
-    if (this.endArrow()) {
+    if (this.endArrow() instanceof Node) {
+      const endArrow = this.endArrow() as Node;
+      endArrow.position(endPoint);
```
aarthificial, line 268: Editing signals inside of `render` / `draw` may cause unwanted behavior.
The `render` method is not guaranteed to be called each time the main generator progresses.
This means that accessing the position of this node from within a generator will return random values, depending on whether `render` was called or not.

Instead, you should transform the context to put the node in the correct location.
You should also rotate the matrix because right now arrows will be oriented the same way.

aarthificial (changes requested):

packages/2d/src/components/Line.ts
strau0106: yes makes sense. I forgot about the rotation sorry...

packages/2d/src/components/Line.ts
```diff
@@ -265,15 +265,25 @@ export class Line extends Shape {
     context.beginPath();
     if (this.endArrow() instanceof Node) {
       const endArrow = this.endArrow() as Node;
-      endArrow.position(endPoint);
+      context.translate(endPoint.x, endPoint.y);
+      context.rotate(Math.atan2(endTangent.y, endTangent.x));
```
aarthificial, line 269:
```suggestion
      context.rotate(endTangent.radians);
```

packages/2d/src/components/Line.ts
```diff
@@ -265,15 +265,25 @@ export class Line extends Shape {
...
+
       endArrow.render(context);
+
+      context.rotate(-Math.atan2(endTangent.y, endTangent.x));
```
aarthificial, line 273: We should use `save/restore` instead. Reverting the transform like that may be too imprecise due to floating-point errors.

packages/2d/src/components/Line.ts
```diff
@@ -265,15 +265,25 @@ export class Line extends Shape {
     context.beginPath();
     if (this.endArrow() instanceof Node) {
       const endArrow = this.endArrow() as Node;
-      endArrow.position(endPoint);
+      context.translate(endPoint.x, endPoint.y);
```
aarthificial, line 268: We also need to find a way to scale the arrows down when there's not enough space on the line.
Otherwise animating the `start/end` parameters will be cumbersome.

aarthificial (changes requested):

packages/2d/src/components/Line.ts
strau0106: could you elaborate on that?

aarthificial: Try running the following example:
```ts
import {Line} from '@motion-canvas/2d/lib/components';
import {all, waitFor} from '@motion-canvas/core/lib/flow';

  const line = createRef<Line>();

  view.add(
    <Line
      ref={line}
      points={[
        [-400, 0],
        [400, 0],
      ]}
      lineWidth={8}
      stroke={'white'}
      startArrow
      endArrow
      arrowSize={100}
    />,
  );

  yield* all(
    line().start(1, 3),
    line().end(0, 3),
  );

  yield* waitFor(5);
});
```
Description | Image
--- | ---
I exaggerated the size to better illustrate the effect. | image
Notice that when the line passes through the point of zero length the arrows get scaled down | image
Without that, it results in this ugly effect: | image

This is done automatically so that you don't need to animate `arrowSize` each time this happens.

strau0106: Should we do this by scaling or "cutting off" the node, if it overlaps?

aarthificial: I'm really sorry, I didn't notice your comment.
I think scaling down the node would be better because it would work even if the arrow exceeds the actual line (for example a circle whose center is placed at the end of the line).
You should be able to use `Node.cacheBBox()` to get the size of the node so that you know when to start scaling it down.

ksassnowski:
@strau0106 Are you still interested in finishing this PR? If not, I'd be happy to take it over from here.

strau0106:
Hi! 
I had a lot going on irl so I forgot about it completely.... Sorry! I‘ll get this done in a few hours/days

~strau

ksassnowski:
No pressure! I just wanted to make sure the work you've done so far doesn't go to waste.

Regarding the merge conflicts: we recently extracted a `Curve` base class that `Line` now extends from (#594). So these changes should also go into the `Curve` class instead of `Line`. You should be able just cut/paste them from `Line` to `Curve`, though.

strau0106:
I‘ll take a look at it! Thanks for the heads up :)

strau0106:
@ksassnowski If this can wait for the next two weeks, I'll finish this PR. If not I am happy to leave it to you

packages/2d/src/components/Curve.ts
```diff
@@ -37,10 +37,16 @@ export abstract class Curve extends Shape {
   @initial(0)
   @signal()
   public declare readonly startOffset: SimpleSignal<number, this>;
-
+  /**
+   * Specifies if or what arrow should be drawn at the start of the line.
+   *
+   * @remarks
+   * If `true`, a default arrow will be drawn. This can be used with the {@link arrowSize} signal to change the size of the arrow.
+   * If set to a {@link Node}, the node will be used as the arrow. This can be used to edraw a custom arrow. If a custom arrow is used, the {@link arrowSize} signal will be ignored.
```
gmile, line 45:
```suggestion
   * If set to a {@link Node}, the node will be used as the arrow. This can be used to draw a custom arrow. If a custom arrow is used, the {@link arrowSize} signal will be ignored.
```

packages/2d/src/components/Curve.ts
```diff
@@ -49,10 +55,16 @@ export abstract class Curve extends Shape {
   @initial(0)
   @signal()
   public declare readonly endOffset: SimpleSignal<number, this>;
-
+  /**
+   * Specifies if or what arrow should be drawn at the start of the line.
+   *
+   * @remarks
+   * If `true`, a default arrow will be drawn. This can be used with the {@link arrowSize} signal to change the size of the arrow.
+   * If set to a {@link Node}, the node will be used as the arrow. This can be used to edraw a custom arrow. If a custom arrow is used, the {@link arrowSize} signal will be ignored.
```
gmile, line 63:
```suggestion
   * If set to a {@link Node}, the node will be used as the arrow. This can be used to draw a custom arrow. If a custom arrow is used, the {@link arrowSize} signal will be ignored.
```

ksassnowski:
@strau0106 Fine with me. Take your time 👍

ksassnowski:
@strau0106 Any updates on this? My offer to take over still stands if you're too busy to continue working on this PR at the moment 👍

strau0106:
Feel free yes. I‘m sorry.

ksassnowski:
Then I'll take it from here. Nothing to be sorry about. Thanks for your work so far!

ksassnowski:
Closing this in favor of #629

Issue #281, opened 2023-02-09
Smooth corners for rounded rectangles


gustavneustadt:
**Description**
In design we often use continuous rounded corners instead of corners defined by a circle. Thats because circular rounded corners have an _sharp_ edge when the corner starts. As far as I know it started in product design and architecture to give the objects a more fluid look. I would really like to have the option to choose which kind of corners  a `Rect` has. I think this would really improve the aesthetics of animations.
As an example, SwiftUI has an `RoundedRectangle` with a `style` parameter to set either `.continuous` or `.circular`. 

**Proposed solution**
It might be just a boolean parameter to pass to a `Rect` like this:
```tsx
<Rect
   width={100}
   height={100}
   radius={10}
   smoothRadius={true}
/>
```

Additional we may provide a way to set the _sharpness_ of corners like this (I think 65% or so is the value Apple uses):
```tsx
<Rect
   width={100}
   height={100}
   radius={10}
   smoothRadius={true}
   cornerSharpness={.65}
/>
```

[I've build an plugin for svelte](https://github.com/gustavneustadt/svelte-smooth-corners) to draw smooth continuous rounded corners as SVG path, so it is not that difficult to implement it via `bezierCurveTo`.
But other than that; if this is a feature people are interested in, I would really like to give it a shot myself.

**Considered alternatives**
I don't think that there is an alternative here …

**Additional context**
Thats what it looks like. If you see it, unfortunately you will never be able to unsee it ;-)
<img width="400" alt="Example image of continous (smooth) and circular corners. " src="https://user-images.githubusercontent.com/35671734/217925506-00ec76c4-b0d3-49df-ab53-41a3cdb6ea32.png">

aarthificial:
Sounds interesting, the logic used for drawing rounded rectangles is here:
https://github.com/motion-canvas/motion-canvas/blob/fe34fa8ebfe66cd356fb1c3d85adedef11e03b45/packages/2d/src/utils/CanvasUtils.ts#L38-L78

One question though, can't we simply treat any `cornerSharpness` greater than zero as `smoothRadius={true}`?
Or does it have to be controlled by two separate settings?

gustavneustadt:
> One question though, can't we simply treat any cornerSharpness greater than zero as smoothRadius={true}?
Or does it have to be controlled by two separate settings?

Good point. CornerSharpness 0 acts like squared off edges. That might be something some one would want … so I am not sure.

aarthificial:
[...]

PR #394, opened 2023-02-21
feat(2d, docs): added documentation for smoothCorners, cornerSharpness

gustavneustadt:
Documents changes made in #281 
Added documentation to `smoothCorners` and `cornerSharpness`.

### DIFF
--- a/packages/2d/src/components/Rect.ts
+++ b/packages/2d/src/components/Rect.ts
@@ -11,7 +11,15 @@ import {SignalValue, SimpleSignal} from '@motion-canvas/core/lib/signals';
 
 export interface RectProps extends ShapeProps {
   radius?: SignalValue<PossibleSpacing>;
+
+  /**
+   * {@inheritDoc Rect.smoothCorners}
+   */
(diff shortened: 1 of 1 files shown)


packages/2d/src/components/Rect.ts
```diff
@@ -19,10 +19,48 @@ export class Rect extends Shape {
...
+   *
+   * @example
+   * ```tsx
+   * <Rect
+   *   width={300}
+   *   height={300}
+   *   smoothCorners={true}
+   * />
+   * ```
+   *
```
aarthificial, line 40: This new line is unnecessary

aarthificial (changes requested):
You should also add `@inheritDoc` to the interface:
```ts
  /**
   * {@inheritDoc Rect.smoothCorners}
   */
  smoothCorners?: SignalValue<boolean>;
```
This will make it inherit the docs from the class property.

gustavneustadt:
Thanks for reviewing! :-)

PR #398, opened 2023-02-21
feat(2d): add z-index property to nodes

ksassnowski:
Adds a `zIndex` signal to `Node` and sorts children by this value before drawing them.


BrettJay:
Oooh I was looking for such a property the other day, had to resort to changing the ordering nodes were added in the view 🤩

aarthificial (approved):

PR #401, opened 2023-02-22
feat: upgraded code-fns for new theme options and lazy loading

### LINKED ISSUE #322: CodeBlock component isn't exported
**Describe the bug**
The `CodeBlock` component isn't exported in [`@motion-canvas/2d/src/components/index.ts`](https://github.com/motion-canvas/motion-canvas/blob/main/packages/2d/src/components/index.ts). Is this intentional?

**To Reproduce**
Attempt to import the CodeBlock component:
`import { CodeBlock } from "@motion-canvas/2d/lib/components";`

**Expected behavior**
The component is successfully imported.

**Console errors**
image

Text form of above error:
`Uncaught SyntaxError: The requested module '/node_modules/.vite/deps/@motion-canvas_2d_lib_components.js?v=c5be1f43' does not provide an export named 'CodeBlock'`

**Package versions:**
- `@motion-canvas/2d: ^2.3.0`

### LINKED ISSUE #396: Upgrade CodeBlock to use Shiki and VSCode themes.
This would be a breaking change, so it would be nice to get it in before 3.0. I'm working on it today.

**Description**
Hunter mentioned Shiki on the discord and I'm now experimenting with it in CodeBlock. It's looking extremely promising. As in, I wish I had known about it before. It would allow for the use of VSCode themes, which would be far better than the theme hack that I did.

If I switch to it it will likely be a breaking change on "theme", since it would switch from an object to a name for a VSCode theme. There might be other changes.

**Proposed solution**
I would update `code-fns`, release a major version, then open a pull for MC.

**Considered alternatives**
I _might_ be able to keep the current theme with another hack and just deprecate it until the 4.0 release. That's what I would do if the 3.0 was years away, but it feels unnecessary at this stage.

In retrospect, I should have brought it in as "experimental", but I never suspected that I would discover a new code highlighter so soon. I searched for a long time before settling on Starry Night.

**Additional context**
If you want your changes in, feel free to release 3.0, but if I can squeeze it in, I'll try to.

Ross-Esmond:
* CodeBlock is now exported
* new stockTheme option with common theme options like github and nord
* new, more granular theme options
* uses the shiki library

closes #396, closes #322

### DIFF
--- a/packages/2d/src/components/CodeBlock.ts
+++ b/packages/2d/src/components/CodeBlock.ts
@@ -9,6 +9,8 @@ import {
   MorphToken,
   Token,
   CodeStyle,
+  Theme,
+  Lang,
 } from 'code-fns';
 import {
   clampRemap,
@@ -21,8 +23,8 @@ import {threadable} from '@motion-canvas/core/lib/decorators';
 import {DesiredLength} from '../partials';
 import {SerializedVector2, Vector2} from '@motion-canvas/core/lib/types';
 import {
-  createComputedAsync,
   createSignal,
+  DependencyContext,
   Signal,
   SignalValue,
   SimpleSignal,
@@ -34,11 +36,12 @@ type CodePoint = [number, number];
 type CodeRange = [CodePoint, CodePoint];
 
 export interface CodeProps extends ShapeProps {
-  language?: string;
+  language?: Lang;
   children?: CodeTree | string;
   code?: SignalValue<CodeTree | string>;
   selection?: CodeRange[];
   theme?: CodeStyle;
+  stockTheme?: Theme;
 }
 
 export interface CodeModification {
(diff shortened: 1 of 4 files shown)


packages/2d/src/components/CodeBlock.ts
```diff
@@ -47,14 +50,44 @@ export interface CodeModification {
...
   @signal()
-  public declare readonly language: SimpleSignal<string, this>;
+  public declare readonly language: SimpleSignal<Lang, this>;
+
+  @initial(undefined)
+  @signal()
+  public declare readonly stockTheme: SimpleSignal<Theme | undefined, this>;
+
+  protected static loadedLanguages = new Map<Lang, boolean>();
+  protected static loadedThemes = new Map<Theme, boolean>();
```
aarthificial, line 62: Looks like these could be simple Sets

packages/2d/src/components/CodeBlock.ts
```diff
@@ -47,14 +50,44 @@ export interface CodeModification {
...
+            CodeBlock.loadedThemes.set(theme, true);
+          }
+        }),
+      );
+      return false;
+    }
+
+    return (
+      CodeBlock.loadedLanguages.get(language) &&
+      (theme == null || CodeBlock.loadedThemes.get(theme))
```
aarthificial, line 88: Couldn't it be:
```suggestion
      CodeBlock.loadedLanguages.has(language) &&
      (theme == null || CodeBlock.loadedThemes.has(theme))
```
?

packages/2d/src/components/CodeBlock.ts
```diff
@@ -47,14 +50,44 @@ export interface CodeModification {
...
-  );
-
   @initial('tsx')
   @signal()
-  public declare readonly language: SimpleSignal<string, this>;
+  public declare readonly language: SimpleSignal<Lang, this>;
+
+  @initial(undefined)
+  @signal()
+  public declare readonly stockTheme: SimpleSignal<Theme | undefined, this>;
```
aarthificial, line 59: The default value should be `null` to be consistent with other Node properties:
```suggestion
  @initial(null)
  @signal()
  public declare readonly stockTheme: SimpleSignal<Theme | null, this>;
```

packages/2d/src/components/CodeBlock.ts
Ross-Esmond: I actually intended for these to use `false` as well. The false means that the language or theme is loading but isn't loaded yet.

packages/2d/src/components/CodeBlock.ts
Ross-Esmond: I updated the code.

aarthificial (approved):

PR #404, opened 2023-02-23
fix(2d): textDirection property for RTL/LTR text

mohsinhijazee:
Currently, the Text component defaults to LTR text direction and hence when setting the offset to -1,-1, we see a slight shift like for RTL texts like this:

https://user-images.githubusercontent.com/24268/220945677-29007823-f869-401c-a491-3c0fd486252b.mp4

With the new property `textDirection` set to `rtl`, we don't see such a shift and text stays steady just like for LTR:

https://user-images.githubusercontent.com/24268/220946030-8b0b7c6e-cd04-442f-bec4-ebfff4c13c83.mp4

Looking for feedback, happyt to make further adjustments. @aarthificial

### DIFF
--- a/packages/2d/src/components/Text.ts
+++ b/packages/2d/src/components/Text.ts
@@ -43,6 +43,7 @@ export class Text extends Shape {
   protected override draw(context: CanvasRenderingContext2D) {
     this.requestFontUpdate();
     this.applyStyle(context);
+    this.applyText(context);
     context.font = this.styles.font;
 
     const parentRect = this.element.getBoundingClientRect();
--- a/packages/2d/src/components/Layout.ts
+++ b/packages/2d/src/components/Layout.ts
@@ -94,6 +94,7 @@ export interface LayoutProps extends NodeProps {
   lineHeight?: SignalValue<Length>;
   letterSpacing?: SignalValue<number>;
   textWrap?: SignalValue<TextWrap>;
+  textDirection?: SignalValue<CanvasDirection>;
 
   size?: SignalValue<PossibleVector2<Length>>;
   offsetX?: SignalValue<number>;
@@ -196,6 +197,9 @@ export class Layout extends Node {
   })
   @signal()
   public declare readonly textWrap: SimpleSignal<TextWrap, this>;
+  @initial('inherit')
+  @signal()
+  public declare readonly textDirection: SimpleSignal<CanvasDirection, this>;
 
   @cloneable(false)
   @inspectable(false)
--- a/packages/2d/src/components/Shape.ts
+++ b/packages/2d/src/components/Shape.ts
@@ -65,6 +65,11 @@ export abstract class Shape extends Layout {
     super(props);
   }
 
+  protected applyText(context: CanvasRenderingContext2D) {
+    context.direction = this.textDirection();
+    this.element.dir = this.textDirection();
+  }
+
   protected applyStyle(context: CanvasRenderingContext2D) {
     context.fillStyle = resolveCanvasStyle(this.fill(), context);
     context.strokeStyle = resolveCanvasStyle(this.stroke(), context);


mohsinhijazee:
Okay, thank you for the feedback. Let me examine the situation.

mohsinhijazee:
I have investigated the issue further by directly experimenting with the canvas and turns out that the bounding box dimensions depend on the flow of text direction. Below are text strings rendered on exactly same x cooridnate (250) but the text flows (and hence the bounding box) in different direction. 

Please see the code below:

```javascript
function draw() {
  const ctx = document.getElementById("canvas").getContext("2d");

  let ltrText = "Hello world";
  let rtlText = "ہیلو ورلڈ";
  let x = 250;
  let y = 0;
  let lineGap = 50;
  let font = "36px serif";
  ctx.font = font;
  
  let ltrMeasure = ctx.measureText(ltrText);
  let rtlMeasure = ctx.measureText(rtlText);

  // Natraully LTR text in both LTR and RTL
  ctx.direction="ltr";
  ctx.fillText(ltrText, x, y + lineGap * 1);
  ctx.direction="rtl";
  ctx.fillText(ltrText, x  , y + lineGap * 2);
  
  // Natrually RTL text in both LTR and RTL
  ctx.direction="ltr";
  ctx.fillText(rtlText, x, y + lineGap  * 3);
  ctx.direction="rtl";
  ctx.fillText(rtlText, x, y + lineGap  * 4);

}
draw();
```
image

Keeping that in context after the `textDirection` property, if the text itself is really RTL, then the bounding box does the right thing (or maybe the wrong thing, because canvas would have flown it the other way around):
```typescript
import {Text} from '@motion-canvas/2d/lib/components';

  view.add(
    <>
    <Text position={[0, -300]} fill={'white'} textDirection={'ltr'} textWrap={true} width={640}>
      Lorem ipsum dolor sit amet, consectetur adipisicing elit. A ad aliquam
      aliquid aut autem consequuntur corporis cum, dolore enim eos facilis
      inventore magnam maxime non officiis quas quibusdam sunt
    </Text>,

<Text position={[0, 100]} fill={'white'} textDirection={'rtl'} textWrap={true} width={640} fontFamily={"Arial"}>
اس قسم کا متن تو پھر بہت مرتبہ آ سکتا ہے لیکن یہاں پر ہو یہ رہا ہے کہ ہمارے پاس جگہ ختم
 ہو رہی ہے لیکن یہاں پر لائن بریک ہی ٹھیک نہیں‌ہیں اور یہ نہیں ہونا چاہئے۔

</Text>
</>
,
  );

  yield* waitFor(5);
});
```
image

But one thing that's broken on the RTL text is - alignment. It is aligned on the left side (and setting textAlign to right breaks it). There are few more complications bascially around text wrapping (correctly considering text direction and text alignment together) so I am little unclear how to proceed with this because this requires multiline text rendering. Current implementation however, would work for shorter RTL strings and not paragraphs.

mohsinhijazee:
Have looked into it further and realised that we are using div tags in a shadow element. So the following code:

```typescript
  view.add(
    <>
      <Text
        position={[0, -300]}
        fill={'white'}
        textWrap={true}
        width={640}
      >
        Lorem ipsum dolor sit amet, consectetur adipisicing elit. A ad aliquam
        aliquid aut autem consequuntur corporis cum, dolore enim eos facilis
        inventore magnam maxime non officiis quas quibusdam sunt
      </Text>
      ,
      <Text
        position={[0, 100]}
        fill={'white'}
        textDirection={'rtl'}
        textWrap={true}
        width={640}
        fontFamily={'Arial'}
      >
        اس قسم کا متن تو پھر بہت مرتبہ آ سکتا ہے لیکن یہاں پر ہو یہ رہا ہے کہ
        ہمارے پاس جگہ ختم ہو رہی ہے لیکن یہاں پر لائن بریک ہی ٹھیک نہیں‌ہیں اور
        یہ نہیں ہونا چاہئے۔
      </Text>
    </>,
  );
```
will produce correct result:
image

The RTL paragraph is correctly right aligned as well. However, when setting the text direction opposite to the text direction (setting direction to RTL for Englis and LTR for Arabic/Urdu) of the script itself will result in broken layout:
image

I am not sure why but the fix is easy. We have to add or subtract the width considering both the _actual language direction of the language_ and the given _text direction_ like so:

```typescript
    if (lang != null && textDirection == 'ltr' && languageDirection == 'rtl') {
      x -= rect.width;
    }

    if (lang != null && textDirection == 'rtl' && languageDirection == 'ltr') {
      x += rect.width;
    }
```
which renders it correctly for each script (if we know the actual text direction of the string):
image

This is exactly how it wouldu appear in browser as well but here's the tricky part: How to detect the script/language? No a problem, can be done but it would cost a lot of computation because _for each chacter, we have to check all the unicode ranges for all the scripts_. And this is an unncessary tradeoff for a such an edge case. Ideally, there's usually no need to set a text direction opposite to the direction of the text but if you must, give us a hint about the language of the string and we make the adjustments and text appers all good:

image

For which, the corresponding JSX would be:

```typescript
    <>
      <Text
        textDirection={'rtl'}
        lang={"en"}
        textWrap={true}
        width={640}
      >
    LTR text here.... longer one...
      </Text>
      ,
      <Text
        textDirection={'ltr'}
        lang={"ur"}
        textWrap={true}
        width={640}
      >
        اس قسم کا متن تو پھر بہت مرتبہ آ سکتا ہے لیکن یہاں پر ہو یہ رہا ہے کہ
        ہمارے پاس جگہ ختم ہو رہی ہے لیکن یہاں پر لائن بریک ہی ٹھیک نہیں‌ہیں اور
        یہ نہیں ہونا چاہئے۔
      </Text>
    </>,
  );
```
I think it fixes the edge case with an addtional (but standard `lang` attribute being set) and if we don't do that, we will have to run the language detection ourselves to know if we have a string that has `textDirection` being supplied in opposite to make our adjustments. Maybe I am having a blindspot here, would appreciate any suggestions. @aarthificial

packages/2d/src/components/Text.ts
```diff
@@ -82,18 +92,37 @@ export class Text extends Shape {
...
+      ? 'rtl'
+      : 'ltr';
+
     const y = rect.y + rect.height / 2;
+    let x = rect.x;
+
+    // Adjustments for the odd cases where textDirection is set to RTL but language itself
+    // is a LTR or vice versa.
+
+    if (lang != null && textDirection == 'ltr' && languageDirection == 'rtl') {
```
mohsinhijazee, line 107: Problem here is that if we do not supply the language direction, then we will have to detect it ourselves which is a slight overhead IMHO for the common case.

aarthificial:
IMO, if the initial solution works for rtl languages then that's all that really matters. I used the latin alphabet for my quick test because I didn't know there's a difference. In practice, I don't see why anyone would like to do that. Especially since the result is the same as `textAlign="right"`.

So, if you're okay with that, I propse we keep things simple and go back to your orignal solution

mohsinhijazee:
Indeed, big proponent of keeping things simple. I have removed the unncessary (related to language and language codes) changes. Ran all the build, tests and e2e tests locally and all green. I forgot to do the `--fixup` commit so it all is squashed into one but next time, will do better.

And as this is my first PR, I'd like to thank you for creating this beautiful animation toolkit. I find the architectural/design choices to be very elegant and ergonomic with the combination of generators and a declarative paradigm (JSX) plus Typescript and then with a faster feedback loop.

aarthificial:
Awesome, thanks for the PR

PR #406, opened 2023-02-23
feat(2d): add save and restore methods to nodes

### LINKED ISSUE #403: Saving and restoring node states
## Description

Say I want to animate a point going across the screen by:

- going right 300px
- going down 200px
- going right 200px

A while later, I want to move the point back to it's original state by following the same path backwards. Currently, the only way for me to do this is by manually adding tweens for all these steps in reverse.

- go left 200px
- go up 200px
- go left 300px

Depending on the situation, reverting an object's transformations back to a specific state can be a non-trivial challenge. It's also a rather brittle solution as when you change the original transition you have to remember to update the "undo" transitions accordingly as well.

## Proposed solution

I propose to add two new methods `save` and `restore` to the `Node` class. These methods would allow you to save a snapshot of an object's current state and then later restore it.

States should be stored on a stack so you can save multiple states and then later undo them in order. This is similar to how the `save` and `restore` methods of the HTML canvas work.

I have a working proof-of-concept locally that required surprisingly few changes. Although I'm sure that there are many edge cases I haven't considered yet. My approach comes down to this:

- Add a new `saveable` property to the `PropertyMetaData` interface. This property would determine where a property gets saved when calling `node.save()`. This would allow us to exclude certain properties from the snapshot such as `children`. It would also allow users to explicitly exclude properties from snapshotting on their own components. By default, this would be set to `true`.
- Add a new `@saveable` decorator to configure if a property should get saved in a snapshot or not.
- Add a `save` method to `Node` that would iterate over all the node's saveable signals and save their current values in a new state object. These objects would then get pushed onto the node's stack.
- Add a `restore` method to `Node` that pops the latest state from the stack and tweens all the saved signal's values back to their previous state.

Here's the code of my current proof of concept:

```ts
export type NodeState = {
  signal: SimpleSignal<any>;
  value: any;
}[];

export class Node implements Promisable<Node> {
  // snip...

  protected stateStack: NodeState[] = [];

  public save(): void {
    const state: NodeState = [];

    for (const {meta, signal} of this) {
      if (!meta.saveable) {
        continue;
      }

      const value = signal();

      // Avoid "Cannot use 'in' operator to search for 'lerp' in false" error.
      if (typeof value === 'boolean') {
        continue;
      }

      state.push({
        signal,
        value,
      });
    }

    this.states.push(state);
  }

  public restore(duration: number, timing: TimingFunction = easeInOutCubic) {
    const state = this.states.pop();

    if (state === undefined) {
      useLogger().error('Trying to restore previous node state but no states have been saved.');
      return;
    }

    const tweens = state
      .filter(({signal, value}) => signal() !== value)
      .map(({signal, value}) => signal(value, duration, timing));

    return all(...tweens);
  }
}
```

The properties that I currently have marked as not saveable using the decorator are:

- `children`
- `spawner`
- `cache`
- `composite`

`absolutePosition` could probably also not be saveable since it gets derived from `position`.

We would also need to check the properties for all built-in components that extend from `Node`.

### Examples

```tsx
import {Circle} from '@motion-canvas/2d/lib/components';
import {all} from '@motion-canvas/core/lib/flow';

  view.add(<Circle ref={circle} width={150} height={150} fill={'slategray'} />);

  circle().save();
  yield* all(circle().position.x(300, 1), circle().fill('darksalmon', 1));

  circle().save();
  yield* all(circle().position.y(200, 1), circle().fill('hotpink', 1));

  circle().save();
  yield* all(circle().position.x(500, 1), circle().fill('bisque', 1));

  yield* circle().restore(1);
  yield* circle().restore(1);
  yield* circle().restore(1);
});
```

Would yield (hah!) the following animation.

https://user-images.githubusercontent.com/5139098/220866454-fcc2637f-aafd-4c41-9066-8fc125f8802a.mp4

## Considered alternatives

There was a brief discussion in the Discord about using the existing `clonable` meta property instead of adding a new one. This might be possible, although we would need to figure out if there are cases where a property should saveable and not cloneable and vice versa.

## Additional Context

This is the equivalent of Manim's [save_state()](https://docs.manim.community/en/stable/reference/manim.mobject.mobject.Mobject.html?highlight=save_state#manim.mobject.mobject.Mobject.save_state) method.

ksassnowski:
## Description

This PR introduces two new methods `save` and `restore` to the base `Node` class. These methods allow us to save a snapshot of a node's current state, so that we can then later revert the node to this state.

As part of implementing the save/restore feature, I have also extracted more general `getState()` and `applyState()` methods.

- `getState()` returns a snapshot of the node's current state. The implementation is almost identical to how `snapshotClone` works, however, it does not include snapshots of the node's children.
- `applyState` take a map of signal values and applies them to the current node. If the signal does not exist on the node, it gets ignore.

This also fixes an issue inside `deepLerp` so it deals with booleans properly.

Closes #403


packages/docs/docs/getting-started/tweening.mdx
```diff
@@ -199,6 +200,41 @@ yield *
   );
 ```
 
+## Saving and restoring states
```
aarthificial, line 203: This should be the last section on the page.

aarthificial (changes requested):
You should also clear the stack in the `dispose` method just to be safe.

PR #411, opened 2023-02-24
Update quickstart.mdx

jumangee:
Could not "serve" new project, created from scratch with nodejs 16 on windows 10x64
Only after updating to 18, with "npm init @motion-canvas" got "Need to install the following packages: @motion-canvas/create@2.5.0..."


aarthificial:
It looks like you just run `npm run serve` before actually installing packages with `npm install`.
We support Node 16 and this is the first time somebody is claiming it doesn't work.

PR #418, opened 2023-02-25
feat(core): add `loopUntil()` function

### LINKED ISSUE #362: Add "loopUntil" flow function
**Description**
Feature request: Add a "loopUntil" flow function
This will act like the "loop" function, replacing the "iterations" number with an event.
It would allow for more fine tuning in the editor as to when something will happen, and up until that point, something else will be going on.

**Proposed solution**
Here is an example of the way the "loopUntil" function would work:
```tsx
// Something vibrates until it explodes (example)
yield * loopUntil(
	'Explode', // "Explode" event
	circle().position.x(-10, 0.2).to(10, 0.2) // To repeatedly occur until event
);
```

**Considered alternatives**
I have considered simply allowing the "waitUntil" function as an "iterations" parameter, but I feel this would be confusing, so should be done as highlighted above, in the same way the "waitFor" and "waitUntil" functions are separated.

squigglesdev:
**Summary**
This PR closes #362 by implementing a `loopUntil()` function. It works in a way similar to `loop()` but uses a `TimeEvent` rather than `iterations` to control the loop.

**Functionality**
This functionality can be achieved by threads, however these are currently undocumented and the approach of adding a new `flow` function would make the process more intuitive for beginners. I followed the naming conventions, and would have renamed `loop()` to `loopFor()` to follow the names of `waitFor()` and `waitUntil()` however this would introduce breaking changes to projects currently using `loop()`.

**Example usage**
```tsx

view.add(
);

yield* loopUntil(  // Perform `circle().position.x(-10,0.1).to(10,0.1)` repeatedly until 'vibrate' event
	'vibrate', () =>
	circle().position.x(-10,0.1).to(10,0.1),
);
```


mkarodia (approved):

packages/core/src/flow/while.ts
```diff
@@ -0,0 +1,38 @@
...
+ * @param tasks - A list of tasks to run.
+ *
+ */
+export function* loopUntil(
+  event: string,
+  factory: LoopCallback,
+): ThreadGenerator {
+  const scene = useScene();
+  const frames = scene.timeEvents.register(event);
+  let iterator = 0;
```
aarthificial, line 28: It's technically `iteration` not an `iterator`

aarthificial (changes requested):

packages/core/src/flow/index.ts
```diff
@@ -13,3 +13,4 @@ export * from './noop';
 export * from './run';
 export * from './scheduling';
 export * from './sequence';
+export * from './while';
```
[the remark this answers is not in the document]
squigglesdev: Oh I see. I thought I couldn't use caps in the file names. Will change that.

packages/core/src/flow/while.ts
squigglesdev: its not a parameter but ok

squigglesdev:
Sorry that last commit took me some time, was in Paris for a week 🥐

packages/core/src/flow/loopUntil.ts
```diff
@@ -0,0 +1,37 @@
...
+  let iteration = 0;
+  while (scene.project.frame < frames) {
+    const generator = factory(iteration);
+    if (generator) {
+      yield* generator;
+    } else {
+      yield;
+    }
+    iteration += 1;
+  }
```
aarthificial, line 36: Some things have changed in version v3
You should implement this more like `waitFor`:
```suggestion
  const scene = useScene();
  const thread = useThread();
  const step = usePlayback().framesToSeconds(1);
  const targetTime = thread.time() + scene.timeEvents.register(event);
  
  let iteration = 0;
  while (targetTime - step > thread.fixed) {
    const generator = factory(iteration);
    if (generator) {
      yield* generator;
    } else {
      yield;
    }
    iteration += 1;
  }
  thread.time(targetTime);
```

aarthificial (changes requested):

aarthificial (approved):

packages/core/src/flow/loopUntil.ts
```diff
@@ -0,0 +1,41 @@
...
+ * @example
+ * ```ts
+ * yield* loopUntil(
+ *   'vibrate',
+ *   () => circle().position.x(-10, 0.1).to(10, 0.1)
+ * );
+ * ```
+ *
+ * @param event - The event.
+ * @param tasks - A list of tasks to run.
```
ksassnowski, line 19: This parameter description is incorrect. It parameter should be called `factory` instead of `tasks`. It's also not a list of tasks to run but a function creating the generator to run. This was probably incorrectly copy&pasted from the doc block of `all`.

I'd suggest using the description of `factory` from the `loop` function.

PR #444, opened 2023-02-28
feat(ui): Add hover shortcuts and global shortcuts

oacs:
# Shortcuts hovers

Now if you hover some of the components that have shortcut a small preview of the shortcuts will be displayed on the footer:

HoverGifExample

## Changes:

- Add footer component  to display the shortcuts here
- useHover hook added
- useShorcut hook added - to fetch the shortcut context
- ShortcutProvider to store all the shortcuts and change the current module that it is hover
- Added example use of the shortcut hook for the Timeline component
- Added shortcuts for Viewport

### DIFF
--- a/packages/ui/src/Editor.tsx
+++ b/packages/ui/src/Editor.tsx
@@ -25,8 +25,12 @@ export function Editor() {
           end={<Viewport />}
         />
       }
-      end={<Timeline />}
-      footer={<Footer />}
+      end={
+        <div>
+          <Timeline />
+          <Footer />
+        </div>
+      }
     />
   ) : (
     <PresentationMode />
--- a/packages/ui/src/Editor.tsx
+++ b/packages/ui/src/Editor.tsx
@@ -26,10 +26,10 @@ export function Editor() {
         />
       }
       end={
-        <div>
+        <>
           <Timeline />
           <Footer />
-        </div>
+        </>
       }
     />
   ) : (
--- a/packages/ui/src/Editor.tsx
+++ b/packages/ui/src/Editor.tsx
@@ -1,5 +1,6 @@
 import {Sidebar} from './components/sidebar';
 import {Timeline} from './components/timeline';
+import {Footer} from './components/footer';
 import {Viewport} from './components/viewport';
 import {ResizeableLayout} from './components/layout';
 import {useState} from 'preact/hooks';
@@ -25,6 +26,7 @@ export function Editor() {
         />
       }
       end={<Timeline />}
+      footer={<Footer />}
     />
   ) : (
     <PresentationMode />
--- a/packages/ui/src/components/footer/Footer.tsx
+++ b/packages/ui/src/components/footer/Footer.tsx
@@ -1,0 +1,21 @@
+import {useShortcuts} from '../../contexts/shortcuts';
+import styles from './Footer.module.scss';
+
+export function Footer() {
+  const {shortcuts, currentModule} = useShortcuts();
+  return (
+    <div className={styles.root}>
+      {shortcuts[currentModule].map(({key, action, isGlobal}, index, arr) =>
+        isGlobal ? null : (
+          <div className={styles.shortcut}>
+            <span className={styles.key}>{key}</span>
+            <span className={styles.action}>{action}</span>
+            {index < arr.length - 1 && (
+              <span className={styles.separator}>|</span>
+            )}
+          </div>
+        ),
+      )}
+    </div>
+  );
+}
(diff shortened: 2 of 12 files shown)


packages/ui/src/contexts/shortcuts.tsx
```diff
@@ -0,0 +1,51 @@
...
+type ShortcutsByModule = Record<ShortcutModules, Shortcut[]>;
+type ShortcutsState = {
+  currentModule: ShortcutModules;
+  shortcuts: ShortcutsByModule;
+  setCurrentModule?: (module: ShortcutModules) => void;
+};
+
+const initialshortcuts: ShortcutsByModule = {
+  global: [
+    {key: 'Space', action: 'Toggle playback', isGlobal: true},
```
aarthificial, line 20: Is `isGlobal` necessary?
Since all global keys by definition are in the `global` module, we could just assume that it will never be set as the current module.

packages/ui/src/contexts/shortcuts.tsx
```diff
@@ -0,0 +1,51 @@
...
+    {key: '←', action: 'Previous frame', isGlobal: true},
```
aarthificial, line 21: We should use ligatures since the editor font supports them:
```suggestion
    {key: '<-', action: 'Previous frame', isGlobal: true},
```
This will make sure that the arrow width matches the rest of the font.

packages/ui/src/contexts/shortcuts.tsx
```diff
@@ -0,0 +1,51 @@
...
+    {key: '→', action: 'Next frame', isGlobal: true},
+    {key: 'Shift + ←', action: 'Reset to first frame', isGlobal: true},
+    {key: 'Shift + →', action: 'Seek to last frame', isGlobal: true},
+    {key: 'm', action: 'Toggle audio', isGlobal: true},
```
aarthificial, line 25: Using uppercase letters would be more readable:
```suggestion
    {key: 'M', action: 'Toggle audio', isGlobal: true},
```

packages/ui/src/contexts/shortcuts.tsx
```diff
@@ -0,0 +1,51 @@
...
+    {key: 'l', action: 'Toggle loop', isGlobal: true},
+  ],
+  timeline: [{key: 'f', action: 'Focus Playhead'}],
+  none: [],
```
aarthificial, line 29: There are also shortcuts available in the viewport:
- `0` reset the zoom
- `+`/`-` zoom in/out

packages/ui/src/hooks/useHover.ts
```diff
@@ -0,0 +1,40 @@
...
+  const _handleMouseOut = () => {
+    setValue(false);
+    handleMouseOut?.();
+  };
+
+  useEffect(
+    () => {
+      const node = ref.current;
+      if (node) {
+        node.addEventListener('mouseover', _handleMouseOver);
```
aarthificial, line 27: Is `mouseover` necessary? Seems like `mouseenter` would be enough.

packages/ui/src/Editor.tsx
```diff
@@ -21,6 +22,7 @@ export function Editor() {
         />
       }
       end={<Timeline />}
+      footer={<Footer />}
```
aarthificial, line 25: `ResizeableLayout` is meant to have only two panes.
We should put the status bar under the timeline instead:
```suggestion
      end={
        <div className="add-necessary-css">
          <Timeline />
          <Footer />
        </div>
      }
```

packages/ui/src/hooks/useHover.ts
oacs: Yeah, the behavior from `mouseenter` is kinda of buggy at least for me

packages/ui/src/components/footer/Footer.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export function Footer() {
+  const {shortcuts, currentModule} = useShortcuts();
+  return (
+    <div className={styles.root}>
+      {shortcuts[currentModule].map(({key, action}, index, arr) => (
+        <div>
+          <span className={styles.key}>{key}</span>
+          <span className={styles.action}>{action}</span>
+          {index < arr.length - 1}
```
aarthificial, line 12: This is no longer necessary

packages/ui/src/Editor.tsx
```diff
@@ -20,7 +21,12 @@ export function Editor() {
           end={<Viewport />}
         />
       }
-      end={<Timeline />}
+      end={
+        <div>
```
aarthificial, line 25: I think some CSS will be needed here because right now the timeline doesn't expand correctly:
image

aarthificial (changes requested):
The viewport shortcuts are now declared but `useShortcut` is not hooked up in the `Viewport`

aarthificial (approved):

aarthificial:
Awesome, thanks for the PR!

PR #456, opened 2023-03-01
feat(docs): add Canvas Operations

WaldemarLehner:
This adds a new Page to the documentation explaining and showing the use of filters and compositeOperations to use as masks.

Screengrab for quick viewing attached below:
Screenshot 2023-03-01 at 18-04-48 Canvas Operations Motion Canvas


packages/docs/docs/advanced/canvas-operations.mdx
```diff
@@ -0,0 +1,99 @@
+---
+sidebar_position: 5
+slug: /canvas-operations
+---
+
+import AnimationPlayer from '@site/src/components/AnimationPlayer';
+
+# Canvas Operations
+
+Because Motion Canvas is build ontop of the Browser's 2D Rendering Context, one
```
nick-1666, line 10:
```suggestion
Because Motion Canvas is build ontop of the Browser's 2D Rendering Context, we
```
On behalf of Hunter, 
https://github.com/motion-canvas/motion-canvas/pull/295#discussion_r1103335169

packages/docs/docs/advanced/canvas-operations.mdx
```diff
@@ -0,0 +1,99 @@
...
+
+Composite Operations define how two layers are stacked ontop of each other. This
+way it is possible to define Masks. MDN has a
+[great visualisation of all Composite Operations](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/globalCompositeOperation#operations).
+
+If you are familiar with Adobe After Effects, you might know about
+[Track Mattes](https://helpx.adobe.com/after-effects/using/track-mattes-and-traveling-mattes.html).
+Track Mattes allow you to define a Background "Color" Layer, and a Foreground
+"Stencil" Layer. The Alpha Matte Layer for example will multiply the
+background's color by the stencil's alpha value.
```
nick-1666, line 46: What about those who aren't? wouldn't a more generalised description of masks be better? (although keep the link I think it may be helpful)

Don't get me wrong, this is a great analogy but wouldn't revolve my explanation around it. Relying on some prior knowledge from other software may be unintuitive, especially seeing as adobe is platform specific and quite pricey.

The best i can come up with is cutting shapes out of card and layering the cards but I doubt its the best.

nick-1666 (approved):
Overall, well worded and very descriptive. I like the mentioning of the nuance when adding multiple filters as it may be an undocumented source of frustration. 

I did write some very nit-picky, opinionated comments as a second opinion but its basically just polish from here on out. Nice job

packages/examples/src/scenes/canvas-operations-filters-stacking.tsx
```diff
@@ -0,0 +1,182 @@
...
+        x={-200}
+        y={250}
+        fontFamily={'monospace'}
+        fill={'white'}
+        fontSize={40}
+      />
+      <Txt
+        text={'1'}
+        y={200}
+        x={-250}
```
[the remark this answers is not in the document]
WaldemarLehner: Suggestions applied. Only place I left this out was the Circle as both dimensions are computed independently, so keeping them separate makes it easier to read.

packages/docs/docs/advanced/canvas-operations.mdx
WaldemarLehner: I would argue that L44-L46 define what a Track Matte is for people that never touched AfterFX. I could go into more detail though if you think that this is not the case.

packages/docs/docs/advanced/canvas-operations.mdx
WaldemarLehner: Applied

packages/docs/docs/advanced/canvas-operations.mdx
```diff
@@ -0,0 +1,99 @@
...
+example because you want to embed the Scene on a website, try to minimize the
+use of cached nodes as each cached node means a separate offscreen canvas. :::
+
+To turn a Node into a cached node, simply pass the `cache` property
+
+```tsx
+<Node cache>...</Node>
+```
+
+Note that all Components inherit from `Node`, so you can set the property on
```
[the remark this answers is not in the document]
WaldemarLehner: Done, but I did it like in the Rest of the Docs: 
```md
[`Node`](...)
```
so it is an inline code segment that is linked

packages/examples/src/scenes/canvas-operations-filters-stacking.tsx
```diff
@@ -0,0 +1,139 @@
...
+        fill={'#ffa'}
+      />
+      <Txt {...commonTxtValues} text={'contrast'} y={250} fill={'#aff'} />
+      <Txt {...commonTxtValues} text={'1'} position={[-200, 250]} />
+      <Txt {...commonTxtValues} text={'1'} position={[-250, 200]} />
+      <Txt {...commonTxtValues} text={'5'} position={[-250, -200]} />
+      <Txt {...commonTxtValues} text={'5'} position={[200, 250]} />
+      <Circle
+        x={() => -200 + (contrastValue() - 1) * 100}
```
aarthificial, line 126:
```suggestion
        x={() => map(-300, -200, contrastValue())}
```

packages/examples/src/scenes/canvas-operations-filters-stacking.tsx
```diff
@@ -0,0 +1,139 @@
...
+        y={() => 200 - (saturateValue() - 1) * 100}
```
aarthificial, line 127:
```suggestion
        y={() => map(100, 200, saturateValue())}
```

packages/examples/src/scenes/canvas-operations-filters-stacking.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const saturateValue = createSignal(1);
+  const contrastValue = createSignal(1);
+
+  scene.add(
+    <Layout position={[-600, -200]}>
+      <AnimatedMotionCanvasIcon
+        filters={[saturate(saturateValue), contrast(contrastValue)]}
+        timePassed={t}
+      />
+      <CodeBlock
```
aarthificial, line 58: This also looks off:
image
The font size shouldn't be different than the one in the CodeBlock

packages/examples/src/scenes/canvas-operations-filters.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+export default makeScene2D(function* (scene) {
+  const t = createSignal(0);
+  scene.add(<Rect size={5000} fill={'#111'} />);
+  const fontSize = 30;
+  const lineHeight = 40;
+  // See [Desmos](https://www.desmos.com/calculator/z9fafnej4k) for Mapping
+  const animationValue = createComputed(
+    () => 0.5 - Math.cos(t() * Math.PI) / 2,
+  );
```
aarthificial, line 17:
```suggestion
  const animationValue = createComputed(
    () => cos(t() * Math.PI, 1, 0),
  );
```

packages/examples/src/scenes/canvas-operations-filters.tsx
```diff
@@ -0,0 +1,115 @@
...
+      text={text}
+    />
+  );
+
+  // First Row
+  scene.add(
+    <Layout layout y={-100} scale={0.7} gap={100} direction={'row'}>
+      <Layout layout direction={'column'}>
+        <AnimatedMotionCanvasIcon
+          filters={() => [blur(animationValue() * 10)]}
```
aarthificial, line 53:
```suggestion
          filters={[blur(() => animationValue() * 10)]}
```

packages/examples/src/canvas-operations-alpha-matte.meta
```diff
@@ -0,0 +1,26 @@
...
+      "y": 1080
+    },
+    "audioOffset": 0
+  },
+  "preview": {
+    "fps": 30,
+    "resolutionScale": 1
+  },
+  "rendering": {
+    "fps": 30,
```
aarthificial, line 20: Shouldn't the default value be `60`?
Is there any reason for it to be `30` in these examples?

packages/docs/docs/advanced/canvas-operations.mdx
aarthificial: I'd like to avoid mentioning AE throughout the docs. 
This can give people a false idea that Motion Canvas is an alternative to it.
I also don't want to risk any sort of legal repercussions.

I agree with @nick-1666 that masks are a more general concept used in graphic software

packages/docs/docs/advanced/canvas-operations.mdx
```diff
@@ -0,0 +1,110 @@
...
+      compositeOperation={'source-in'}
+    />
+  </Node>
+);
+
+yield* yOffset(100, 2);
+yield* yOffset(-100, 2);
+```
+
+<AnimationPlayer name="canvas-operations-alpha-matte" />
```
aarthificial, line 73: `<AnimationPlayer/>` wasn't really meant to be used in full-width mode like this.
Right now it can be either a `banner` or a `small` player.
You'd need to add `margin-bottom: var(--ifm-leading);` to the styles to use it like this.

packages/docs/docs/advanced/canvas-operations.mdx
```diff
@@ -0,0 +1,110 @@
...
+[`Node`](/api/2d/components/Node). Filters can set it automatically, while
+Composite Operations require you to set it explicitly on the parent
+[`Node`](/api/2d/components/Node).
+
+A cached [`Node`](/api/2d/components/Node) and its children are rendered on an
+offscreen canvas first, before getting added to the main scene.  
+For filters this is needed because they are applied to the entire canvas. By
+creating a new canvas and moving the elements that should get affected by the
+filters over, applying filters to the entire "new" canvas, and then moving back
+the result, you effectively only apply the filters to the moved elements.
```
aarthificial, line 89: > you effectively only apply the filters to the moved elements

That's not the reason we render stuff onto a separate canvas, though.
The filters are always applied *only* to whatever you are drawing.
But there's a difference between drawing two blurred nodes separately, one after the other. And drawing them first and then blurring the result.
This is especially noticeable when it comes to opacity. Which will also force the node to be `cached`.

packages/docs/docs/advanced/canvas-operations.mdx
```diff
@@ -0,0 +1,110 @@
...
+
+<!-- Prettier merges header with text -->
+<!-- prettier-ignore -->
+:::caution Performance Implications 
+If you care about playback performance, for
+example because you want to embed the Scene on a website, try to minimize the
+use of cached Nodes as each cached [`Node`](/api/2d/components/Node) means a separate offscreen canvas.
+
+<!-- prettier-ignore -->
+:::
```
aarthificial, line 99: Admonitions should be formatted with new lines inbetween:
```suggestion
:::caution Performance Implications

If you care about playback performance, for example because you want to embed
the Scene on a website, try to minimize the use of cached Nodes as each cached
[`Node`](/api/2d/components/Node) means a separate offscreen canvas.

:::
```

aarthificial: By the way, do we have any tests for this? 
How do additional canvases slow the playback?

packages/docs/docs/advanced/canvas-operations.mdx
```diff
@@ -0,0 +1,110 @@
...
+<AnimationPlayer name="canvas-operations-alpha-matte" />
+An example of how composite operations are used to create a mask. The Motion Canvas
+Icon acts as the stencil, the image in the back as the value.
+
+## Cached Nodes
+
+Both Filters and Composite Operations require a cached
+[`Node`](/api/2d/components/Node). Filters can set it automatically, while
+Composite Operations require you to set it explicitly on the parent
+[`Node`](/api/2d/components/Node).
```
aarthificial, line 82: > Composite Operations require you to set it explicitly on the parent

That's not necessarily true, any of the ancestors can be cached to create all sorts of different effects.

packages/docs/docs/advanced/canvas-operations.mdx
```diff
@@ -0,0 +1,110 @@
...
+sidebar_position: 5
+slug: /canvas-operations
+---
+
+import AnimationPlayer from '@site/src/components/AnimationPlayer';
+
+# Canvas Operations
+
+Because Motion Canvas is build ontop of the Browser's 2D Rendering Context, we
+can make use several canvas operations that are provided by the Browser.
```
aarthificial, line 11: I'm not sure if this section should be called "Canvas Operations" to be honest.
Any method of `CanvasRenderingContext2D` could be called a canvas operation.
This guide is more about Filters and Effects.

packages/docs/docs/advanced/canvas-operations.mdx
```diff
@@ -0,0 +1,110 @@
...
+`@motion-canvas/2d/lib/partials`
+([API Documentation](/api/2d/partials#Functions)). Note that some filters, like
+`opacity`, can be set directly on the [`Node`](/api/2d/components/Node#opacity),
+and as a result are not exported from [partials](/api/2d/partials#Functions).
+
+`filters` accepts an array. This means that you can stack multiple filters onto
+a single Node. The order in which you add them does matter, as can be seen
+below:
+
+<AnimationPlayer name="canvas-operations-filters-stacking" />
```
aarthificial, line 31: I know that the animation includes code snippets but it's impossible to copy them.
Having an example in a code block together with imports would be nice.

packages/docs/docs/advanced/canvas-operations.mdx
```diff
@@ -0,0 +1,110 @@
...
+the Browser can understand on
+[MDN](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/filter).
+
+All filters available in Motion Canvas can be found under
+`@motion-canvas/2d/lib/partials`
+([API Documentation](/api/2d/partials#Functions)). Note that some filters, like
+`opacity`, can be set directly on the [`Node`](/api/2d/components/Node#opacity),
+and as a result are not exported from [partials](/api/2d/partials#Functions).
+
+`filters` accepts an array. This means that you can stack multiple filters onto
```
aarthificial, line 27: Signals for all filters can also be accessed using properties:
```ts
yield* node().filters.blur(20, 1);
```
In this case, `filters` will try to find the first `blur` in the array, and if it can't do that it will create and add a new one to the list.

aarthificial (changes requested):

packages/examples/src/canvas-operations-alpha-matte.meta
WaldemarLehner: Same reason I set the default rendering resolution from 2x to back to 1.

Trying to run 2 or more animations made them extremely unresponsive and stuttery.

aarthificial: Even after removing the codeblock tweening?

WaldemarLehner: Yep. This was after changing to a static code block.

Animations would play back extremely slow until I reduced the render size (which was at 2x by default for some reason) and framerate.

You can still notice that it slows down a little bit when playing multiple animations, but it's no longer that big of a change.

aarthificial: For me, on Chrome, in 4k 60fps, none of the examples cause trouble. Maybe it's a hardware difference though.

packages/examples/src/scenes/canvas-operations-alpha-matte.tsx
```diff
@@ -0,0 +1,105 @@
+import {makeScene2D} from '@motion-canvas/2d';
+import {AnimatedMotionCanvasIcon} from '../components/AnimatedMotionCanvasIcon';
+import {createSignal} from '@motion-canvas/core/lib/signals';
+import {linear} from '@motion-canvas/core/lib/tweening';
+import {Gradient} from '@motion-canvas/2d/lib/partials';
+import {Img, Layout, Rect, Txt} from '@motion-canvas/2d/lib/components';
+
+import valueImage from '../../assets/federico-bottos-Z3NceSeZqgI-unsplash.jpg';
```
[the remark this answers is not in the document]
WaldemarLehner: Could do. My rationale was to keep it local in case the file path changes or the picture is taken offline. Also means that the Docs can be used offline.

packages/docs/docs/advanced/canvas-operations.mdx
WaldemarLehner: The relevant code bit is just above it. So I would argue that you could just copy that instead. 
The comment above `compositeOperation` tells you what changing between source-in and source-out does.

I could add the relevant imports.
image

packages/docs/docs/advanced/canvas-operations.mdx
WaldemarLehner: So something like 
```ts
yield* node().filters.blur(20, 1);
yield* node().filters.saturation(0, 1);
```
would be equivalent to 
```tsx
const blurValue = createSignal(0);
const saturationValue = createSignal(1);

yield scene.add(<Node filters={[blur(blurValue), saturation(saturationValue)]} />)
yield* blurValue(20, 1);
yield* saturationValue(0, 1);
```
?

That might actually be a more preferable syntax :D

packages/docs/docs/advanced/canvas-operations.mdx
WaldemarLehner: My main thought was that excessive use of filters would cause slowdown — having experienced this when working on some animations (this has been a week after the public release, so caching might have been different back then).

I tried to replicate but failed. I actually got the opposite result!

I tried this scene:
```tsx
const cacheNodes = true

export default makeScene2D(function* (scene) {
  scene.fill("#aaa")
  yield scene.add(
    <Node cache={cacheNodes}>
      {
        ([...new Array(1000).keys()]).map( (i) => {
          return <Icon filters={[blur(10)]} icon={"material-symbols:360"} cache={cacheNodes} x={useRandom().nextInt(-500, 500)} y={useRandom().nextInt(-500, 500)} fill={useRandom().nextInt(0, 0xFFFFFF)} size={40} />
        })
      } 
    </Node>
  )
  yield* waitFor(5);
});
```
This will hang for 2 or so seconds at the start of the scene, then render everything at 30 (target) FPS.
When I tried to set `cacheNodes` to `false`, I got 2 FPS. 

So I guess I will just remove the admonition.

packages/examples/src/components/AnimatedMotionCanvasIcon.tsx
```diff
@@ -0,0 +1,141 @@
...
+    const GREEN = '#99C47A';
+    const star = createRef<Node>();
+    const trail1 = createRef<Layout>();
+    const trail2 = createRef<Layout>();
+    const trail3 = createRef<Layout>();
+    const dot = createRef<Rect>();
+
+    this.add(
+      <>
+        <Layout layout={false}>
```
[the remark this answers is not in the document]
WaldemarLehner: You are correct. I believe it was relevant before I turned everything into cached nodes :)

packages/examples/src/components/AnimatedMotionCanvasIcon.tsx
```diff
@@ -0,0 +1,141 @@
...
+      <Layout layout direction={'column'} gap={30} offsetY={-1} {...props} />
+    );
+    const YELLOW = '#FFC66D';
+    const RED = '#FF6470';
+    const GREEN = '#99C47A';
+    const star = createRef<Node>();
+    const trail1 = createRef<Layout>();
+    const trail2 = createRef<Layout>();
+    const trail3 = createRef<Layout>();
+    const dot = createRef<Rect>();
```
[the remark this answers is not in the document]
WaldemarLehner: Nope. Have been removed

packages/docs/docs/advanced/canvas-operations.mdx
aarthificial: Yes

packages/docs/docs/advanced/canvas-operations.mdx
aarthificial: I must have missed that code block.
Btw, we now have a fiddle component that displays an editor with a preview which could be better in this case:

https://user-images.githubusercontent.com/64662184/236676320-6753910d-d506-4c0f-8593-56d3b8f233a2.mp4

You could make the reader edit the operations themselves and toggle between a few presets.
It's also really easy to write in markdown:
https://github.com/motion-canvas/motion-canvas/blob/fae72b3729f2a7b5b1291a92fbf2ff8e767e0d43/packages/docs/docs/components/spline.mdx?plain=1#L362-L418

packages/docs/docs/advanced/canvas-operations.mdx
aarthificial: Yes, exactly

aarthificial:
I modified the PR a bit to update the imports in the examples and switch to some of the newer features where applicable.
I also rewrote the Filters section a bit to get rid of the repeating "Note that" phrase.

Sadly, I decided to remove the gradients from the examples to keep them consistent with the rest of the docs.
The gradient declaration was taking a significant portion of each code snippet and it didn't seem to make the visuals clearer and so I think it's for the best.

aarthificial (approved):

WaldemarLehner:
Perfect. Yeah. The Idea behind the Gradients was to signify that masking actually causes the value layer to become transparent, and not just get filled in a solid color :) 

An alternative could be to add a grid, but I don't thing it is really needed for the example.

aarthificial:
I'll think about adding some fiddle-level background control, something like a simple CSS class.
So that the background doesn't have to be included in the snippet code itself.

Anyway, thanks so much for the PR! The page should be up later today.

PR #463, opened 2023-03-02
feat(2d): add Polygon component (#455)

### LINKED ISSUE #455: Create a `Polygon` component
**Description**

It would be helpful to have a component to draw various regular polygons, e.g. hexagons, octagons, etc.

**Proposed solution**

I propose this can be implemented by extending `Rect` and directly drawing to canvas by overriding the `getPath()` method.

You would pass these parameters to `PolygonProps`:

- `nSides` : number of sides of the polygon
- `angle`: the starting angle offset, in degrees

The `angle` parameter is necessary to control the orientation of the polygon.  For example, changing the angle of a hexagon can change the appearance so the hexagon is standing either on a point, or on a side.

This would be different from changing the `rotation` of the polygon.  I anticipate the need to have, for example, text inside a polygon (as a child node) but the text needs to remain horizontal, even if the angle of the polygon changes.

Note: the implementation draws the polygon inside a bounding circle (or ellipse). Thus it may appear that not all the vertices touch the bounding rectangle.

**Considered alternatives**

I tried to implement this by extending `Line` but the points of a `Line` don't animate well in canvas.

**Additional context**

I have working code and can prepare a PR, if this is useful.

Here is an example:

motion-canvas-polygon

```
import {Polygon} from '../components/Polygon';

  const myPoly = createRef<Polygon>();
  const radius = createSignal(50);
  const nsides = createSignal(3);

  view.add(
    <>
      <Polygon 
        ref={myPoly}
        width={() => radius()}
        height={() => radius()}
        nSides={() => nsides()} 
        angle={() => 0.5 * 360 / nsides()}
        fill={'#ff0000'} 
        stroke={'#fff'}
        lineWidth={3} 
      />
    </>
  )

  yield * all (
    radius(100, 2) ,
    nsides(8, 2),
    myPoly().fill('#004400', 2)
  );
});
```

andrie:
Implements Polygon class by extending Shape

Closes #455


packages/2d/src/components/Polygon.ts
```diff
@@ -0,0 +1,104 @@
...
+export interface PolygonProps extends ShapeProps {
+  /**
+   * {@inheritDoc Polygon.nSides}
+   */
+  nSides?: SignalValue<number>;
+
+  /**
+   * {@inheritDoc Polygon.angle}
+   */
+  angle?: SignalValue<number>;
```
ksassnowski, line 14: I might be missing something here, but doesn't this end up doing basically the same as `rotation` already does? I understand that it draws the polygon at a different rotation without actually changing the `rotation` property of the element. But I'm not sure if that doesn't end up making things more confusing as there are now two ways of (visually) rotating a polygon. `Rect` also doesn't have an `angle` property.

Unless there's some use case I am overlooking?

packages/2d/src/components/Polygon.ts
```diff
@@ -0,0 +1,104 @@
...
+      const y = -height * Math.cos(theta);
+      if (i == 0) {
+        path.moveTo(x, y);
+        x0 = x;
+        y0 = y;
+      } else {
+        path.lineTo(x, y);
+      }
+    }
+    path.lineTo(x0, y0);
```
ksassnowski, line 101: There's a potential issue with this approach of "closing" the polygon by manually drawing the final line segment back to the start location: the `lineJoin` property won't get applied to the last and first point of the polygon which–depending on the angle–will produce an ugly gap or overlap.

For what it's worth, this is also a problem with the `Line` component.

I think it might be better to override the `drawShape` method instead of `getPath`. This would allow us to call `context.closePath()` instead of manually drawing the final line segment. This would ensure that the first and last line segments would get joined properly.

ksassnowski: CleanShot 2023-03-03 at 10 42 40@2x
Here's the issue I'm talking about

ksassnowski: Sorry, I was overcomplicating the proposed solution a little bit. All that's required is to add a `path.closePath()` call inside the `getPath` method. This also allows us to simplify the method itself a little bit because we no longer have to draw the final line segment ourselves:
```diff
protected override getPath(): Path2D {
    const path = new Path2D();
    const angle = (this.angle() / 360) * 2 * Math.PI;
    const width = this.width() / 2;
    const height = this.height() / 2;
    const nSides = this.nSides();
-   let x0 = 0;
-   let y0 = 0;

-    for (let i = 0; i <= nSides; i += 1) {
+   for (let i = 0; i <= nSides - 1; i += 1) {
      const theta = (i * 2 * Math.PI) / nSides - angle;
      const x = width * Math.sin(theta);
      const y = -height * Math.cos(theta);
      if (i == 0) {
        path.moveTo(x, y);
-       x0 = x;
-       y0 = y;
      } else {
        path.lineTo(x, y);
      }
    }
+   path.closePath();
    return path;
  }
```

CleanShot 2023-03-03 at 10 49 06@2x

packages/2d/src/components/Polygon.ts
andrie: The use case I have in mind where this might matter is where you have children of the polygon.  Imagine you want to put text inside the polygon.  If you *rotate* the polygon to get a certain alignment, then the children will also rotate.  So then do you have to anti-rotate the children to get the text to align properly?
Similarly, the use case where you want to draw a connecting line from the polygon to another element, let's say from the right edge.  If you change the rotation of the polygon, then I think the right edge will also rotate, so the connecting point may no longer be where you expect.
I may well be wrong in my understanding of how child components work though, so open to illumination.

packages/2d/src/components/Polygon.ts
andrie: This is great, thanks.  I observed that gap, but didn't know what caused it or how to fix.  Let me investigate.

andrie: This works great, thank you!

packages/2d/src/components/Polygon.ts
andrie: Here is an example of the use case I have in mind.

```
  view.add(
    <>
      <Polygon 
        ref={hex} x={0} y={0} width={320} height={320} nSides={3} 
        stroke={'#fff'} lineWidth={8}
        angle={1 * 360/3}
        // rotation={1 * 360/3}
        fill={'orange'}
        children={<Txt ref={label} text="Yield" fill="white" />}
      />
    </>
  );

  yield* waitFor(0.5);

  yield* all(
    hex().nSides(8, 5),
    hex().fill("red", 5),
    hex().angle(0.5 * 360 / 8, 5),
    // hex().rotation(0.5 * 360 / 8, 5),
    label().text("Stop", 5)
  );
  
  yield* waitFor(0.5);
```

Using `angle`:

motion-canvas-polygon-angle

Using `rotation`:

motion-canvas-polygon-rotation

andrie: Would it help if I add this to the documentation?

```
   * @remarks
   * Use `angle` to change the apparent rotation of the polygon.
   * One reason you would use `angle` instead of `rotation` is when you
   * use child components, e.g. `Txt` inside the polygon.  In this case you
   * may want to change the angle of the polygon, rather than rotation of the 
   * polygon as well as it's children.
```

aarthificial: I think I agree with @ksassnowski here
You can achieve the same thing by doing:
```tsx
<Node>
  <Polygon rotation={30} />
  <Txt>Some text here</Txt>
</Node>
```
This touches on a more general problem: transforming the content rendered by a node separately from its children. And I believe it should be addressed in the base `Node` class in a separate issue. This way other nodes like `Rect`s and `Img`s would also benefit from this.

When it comes to drawing lines between the polygon and other things, it could be more useful to add a `getVertexPosition(n)` method that would return the position of the vertex in local space.

packages/2d/src/components/Polygon.ts
```diff
@@ -0,0 +1,104 @@
+import {Shape, ShapeProps} from './Shape';
+import {initial, signal} from '../decorators';
+import {SignalValue, SimpleSignal} from '@motion-canvas/core/lib/signals';
+
+export interface PolygonProps extends ShapeProps {
+  /**
+   * {@inheritDoc Polygon.nSides}
+   */
+  nSides?: SignalValue<number>;
```
aarthificial, line 9: What would you say about calling this prop simply `sides`?

packages/2d/src/components/Polygon.ts
```diff
@@ -0,0 +1,104 @@
...
+   */
+  @initial(0)
+  @signal()
+  public declare readonly angle: SimpleSignal<number, this>;
+
+  public constructor(props: PolygonProps) {
+    super(props);
+  }
+
+  protected override getPath(): Path2D {
```
aarthificial, line 80: `Shape`s should also implement a `getRipplePath()` method that generates the same path but expanded by `this.rippleSize()`. This is used for ripple effects:
```ts
yield* shape().ripple();
```

aarthificial (changes requested):

packages/2d/src/components/Polygon.ts
andrie: Sounds good to me.

andrie: And thanks for the hint on using `<Node>` instead.  I think this works.

packages/2d/src/components/Polygon.ts
andrie: Let me investigate.

andrie: I have `ripple()` working.  Will prepare a fresh commit.

packages/2d/src/components/Polygon.ts
```diff
@@ -0,0 +1,97 @@
...
+    const theta = (i * 2 * Math.PI) / sides;
+    const x = width * Math.sin(theta);
+    const y = -height * Math.cos(theta);
+    if (i == 0) {
+      path.moveTo(x, y);
+    } else {
+      path.lineTo(x, y);
+    }
+  }
+  path.closePath();
```
aarthificial, line 29: This is an optional suggestion, but I think the code would be a bit more readable when using vectors:
```suggestion
  const size = rect.size.scale(0.5);
  for (let i = 0; i <= sides; i++) {
    const theta = (i * 2 * Math.PI) / sides;
    const direction = Vector2.fromRadians(theta).perpendicular;
    const vertex = direction.mul(size);
    if (i == 0) {
      moveTo(path, vertex);
    } else {
      lineTo(path, vertex);
    }
  }
  path.closePath();
  ```
  since they better illustrate the math that's happening

File: packages/2d/src/components/Polygon.ts [hunk 6]
```diff
@@ -0,0 +1,97 @@
+import {
+  // PossibleSpacing,
```
aarthificial, line 2: There's some leftover commented code here and there

packages/2d/src/components/Polygon.ts
```diff
@@ -0,0 +1,97 @@
...
+  BBox,
+  // SpacingSignal,
+} from '@motion-canvas/core/lib/types';
+import {Shape, ShapeProps} from './Shape';
+import {initial, signal} from '../decorators';
+import {SignalValue, SimpleSignal} from '@motion-canvas/core/lib/signals';
+
+function drawPolygon(
```
aarthificial, line 10: This function could be moved to `src/utils/CanvasUtils.ts`.

packages/2d/src/components/Polygon.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+export interface PolygonProps extends ShapeProps {
+  /**
+   * {@inheritDoc Polygon.sides}
+   */
+  sides?: SignalValue<number>;
+}
+
+export class Polygon extends Shape {
```
aarthificial, line 39: Can we also add a short description for the node with an example or two?
Something like the `Circle` node:
https://github.com/motion-canvas/motion-canvas/blob/1c259d0d574bb56dbc8bc448300d9b94ee4d0bc4/packages/2d/src/components/Circle.ts#L20-L63

aarthificial (changes requested):

packages/2d/src/components/Polygon.ts
andrie: Thank you.  This is helpful.

I had to make a small modification to get this to run:

``` js
    if (i == 0) {
      path.moveTo(vertex.x, vertex.y);
    } else {
      path.lineTo(vertex.x, vertex.y);
    }
```

packages/2d/src/components/Polygon.ts
andrie: Thanks. Fixed now.

packages/2d/src/components/Polygon.ts
ksassnowski: The `moveTo` and `lineTo` functions are helper functions that can be imported from `@motion-canvas/2d/lib/utils`. These functions actually accept vectors as inputs instead of expecting each point individually.

packages/2d/src/components/Polygon.ts
andrie: Done.

packages/2d/src/components/Polygon.ts
andrie: Working on this.  Thanks for pointing me to the location where `Circle` is described.  This is helpful.

andrie:
@aarthificial  and @ksassnowski 

I propose to modify the API for polygon to have `size` instead of `height` and `width`.

I think the most common use case is going to be regular polygons, and using `size` that implies equal height and width is simpler to use.

The only reason I used `height` and `width` in the first place was because someone suggested on Discord that I inherit from `Rect`.  But given we're now inheriting from `Node` directly, using the design pattern of `Circle` with size makes more sense to me.

andrie:
Please ignore my previous comment about `size`.  It turns out that it just works already.

packages/2d/src/components/Polygon.ts
andrie: Done

packages/2d/src/components/Polygon.ts
andrie: I tried that and couldn't get it to work.  I'll try again.

ksassnowski: Can you provide more details on what exactly wasn't working?

andrie: OK, I figured out why.  I haven't pulled the latest changes from `main`, and now I see there are changes in v3.1 that allows this new syntax.
Thank you.

andrie:
For funzies I made this.

Another `--fixup` is inbound soon.

motion-canvas-hex-grid

packages/2d/src/components/Polygon.ts
andrie: Hard to say now.  Probably user error.

(I thought something changed in v3.1, so I did a `git pull` and tried again. This time it worked. So if there were no changes in `utils` in v3.1, then this was certainly user error.)

aarthificial (approved):

Issue #464, opened 2023-03-02
CodeBlock not exported as per documentation


alpharahl:
**Describe the bug**

The documentation for the code block section [Docs](https://motioncanvas.io/docs/code-block) mentions importing CodeBlock from the 2d lib directory:

`import {CodeBlock} from '@motion-canvas/2d/lib/components';`

However, it looks like this export was removed.

**To Reproduce**

Following the getting started guide:

`npm init @motion-canvas`

run npm install

Within example.tsx attempt to import CodeBlock

```ts
import {Circle, CodeBlock} from '@motion-canvas/2d/lib/components';
```

Editor will throw error indicating CodeBlock is not exported from components file.

**Expected behavior**
CodeBlock is exported as described in the CodeBlock docs

**Package versions (please complete the following information):**

```
"dependencies": {
        "@motion-canvas/2d": "^3.0.2",
        "@motion-canvas/core": "^3.0.0"
      },
      "devDependencies": {
        "@motion-canvas/ui": "^3.0.1",
        "@motion-canvas/vite-plugin": "^3.0.0",
        "typescript": "^4.9.5",
        "vite": "^3.0.5"
      }
```

PR #468, opened 2023-03-03
docs: correct import for CodeBlock (#464)

alpharahl:
Due to #401 the codeblock component is no longer exported from the components index. Docs should be updated to reflect this change

### DIFF
--- a/packages/docs/docs/components/code-block.mdx
+++ b/packages/docs/docs/components/code-block.mdx
@@ -22,7 +22,7 @@ important snippets.
 To display code, set the `code` and `language` property.
 
 ```ts
-import {CodeBlock} from '@motion-canvas/2d/lib/components';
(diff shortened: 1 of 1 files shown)


packages/docs/docs/components/code-block.mdx
```diff
@@ -165,7 +165,8 @@ for future edits. Then call `edit` with an embedded `insert` call to add the new
 code.
 
 ```tsx
-import {CodeBlock, insert} from '@motion-canvas/2d/lib/components';
+import {insert} from '@motion-canvas/2d/lib/components';
```
aarthificial, line 168: `insert` is also currently not exported from `'@motion-canvas/2d/lib/components`

aarthificial (changes requested):

aarthificial:
I'll apply the changes in #504 because I need this to be ready for the next release.

Issue #462, opened 2023-03-02
feat: Individual scene export option 


nick-1666:
Add an option to render all scenes into individual folders so we can render the videos with a script using ffmpeg instead of counting frames or splitting a video.

aarthificial:
I think it's a good idea.
It could be added as an additional option in the Video Settings.

misclicl:
hi there!

I'd like to work on this if you don't mind.

aarthificial:
Closed by #477

PR #477, opened 2023-03-05
feat(core,ui,vite-plugin): add option to group output by scenes

misclicl:
This update adds an option to the Video Settings tab that allows users to organise their project output by scenes. The options is disabled by default.


misclicl:
damn, sorry for the mess, but I can't figure how I'm supposed to link the issue (https://github.com/motion-canvas/motion-canvas/issues/462) to this. i haven't used Github in a while. does the name have to be identical?

packages/ui/src/components/settings/RenderingSettings.tsx
```diff
@@ -69,6 +69,16 @@ export function RenderingSettings() {
           />
         </Group>
       )}
+      <Group>
+        <Label>{meta.rendering.groupByScene.name}</Label>
```
misclicl, line 73: i'm not sure if this meant to be pulled from the `meta` to be honest. `MetaField` documentation mentions something like this: https://github.com/motion-canvas/motion-canvas/blob/1c259d0d574bb56dbc8bc448300d9b94ee4d0bc4/packages/core/src/meta/MetaField.ts#L25-L31
while no other fields use it. depending on your answer i can either change it all to read from `MetaField` value, or use string literal here instead.

misclicl:
hmm, now thinking about it i realised that i probably want to reset the frame number in between scenes.

packages/ui/src/components/settings/RenderingSettings.tsx
```diff
@@ -69,6 +69,16 @@ export function RenderingSettings() {
...
+        <Input
+          type="checkbox"
+          checked={Boolean(state.groupByScene)}
```
aarthificial, line 76: The cast is not necessary:
```suggestion
          checked={state.groupByScene}
```
The type system prevents anything other than a boolean to be set as the value of this field.
If you want to make sure that the value read from the meta file is correct, you should create a new `BoolMetaField` class and override the `parse` method.

packages/core/src/vite-env.d.ts
```diff
@@ -7,11 +7,13 @@ declare module 'vite/types/customEvent' {
...
       data: string;
-      mimeType: string;
-      project: string;
-      isStill: boolean;
+      meta: {
+        name: string;
+        subDirectories: string[];
+        mimeType: string;
+        frameNumber: number;
+      };
```
aarthificial, line 16: I'm not sure what's the point of wrapping these things in an additional object.

aarthificial (changes requested):

packages/ui/src/components/settings/RenderingSettings.tsx
misclicl: you're right, not sure why i added `Boolean` here. will remove. although idk about your suggestion with `BoolMetaField`. does it make sense to create it now? I initialise `groupByScene` with a boolean value in `createProjectMetadata`. Unless someone decides to manually modify meta files this seems useless.

packages/core/src/app/FrameSaver.ts
```diff
@@ -0,0 +1,71 @@
+export class FrameSaver {
```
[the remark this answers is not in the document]
misclicl: I definitely overcomplicated lot of things here. thanks for the review. `FrameSaver` class is removed.

packages/ui/src/components/settings/RenderingSettings.tsx
misclicl: alright, I'll just use string literal here then.

packages/core/src/app/ImageExporter.ts
```diff
@@ -56,13 +60,19 @@ export class ImageExporter implements Exporter {
...
+        {
+          frame,
+          fileType: this.fileType,
+          projectName: this.projectName,
+          groupByScene: this.groupByScene,
+          sceneName,
+        },
+      );
+
+      return Promise.resolve();
```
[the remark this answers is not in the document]
misclicl: removed.

packages/core/src/vite-env.d.ts
misclicl: changed it to a flat structure now.

misclicl:
@aarthificial I addressed everything.

aarthificial:
Thanks!
To address your comments that I missed before:
> does the name have to be identical?

If the pr closes an issue, the commit should contain this in its body:
```
Closes: #ISSUE_ID
```
But I can add this when merging so no worries.

> hmm, now thinking about it i realised that i probably want to reset the frame number in between scenes.

I think it's a good idea too. You can calculate the frame number relative to the scene in the `Renderer`:
```ts
const frame = this.playback.frame - this.playback.currentScene.firstFrame;
```

nick-1666:
Thanks for working on this. I wanted this for the longest time :)

misclicl:
> I think it's a good idea too. You can calculate the frame number relative to the scene in the Renderer:
> ```
> const frame = this.playback.frame - this.playback.currentScene.firstFrame;
> ```
i tried this, but `this.playback.currentScene.firstFrame` seems to return `0` regardless of the current scene. i can look into it a bit later.

aarthificial:
image
image

misclicl:
@aarthificial, the issue with `0` is caused by the `PlaybackManager` instance used in `Renderer` never updating the `firstFrame` variable because the `recalculate` function is never called. I tried calling `recalculate` during the `render`, but it caused the rendering to become stuck at the very end. While I can investigate further, do you have any tips for me?

aarthificial:
Rendering will always get stuck when using `ui:serve` because there's no backend in place to handle the export.
`playback.recalculate()` should be called in the `run` method, after `reloadScenes`.

misclicl:
I was running all core, UI, and plugins in watch mode with template:serve for this issue, so that wasn't the problem. The real reason was that I messed up ImageRenderer 🤦 Anyway, everything is working properly now.

aarthificial (changes requested):
Requesting changes because there are merge conflicts

aarthificial (approved):

aarthificial:
Awesome, thanks for the PR!

PR #478, opened 2023-03-05
Add `XY` types of methods to Vector2

magx2:
I propose to add methods with suffix `XY` that accepts 2 numbers: x and y`.

WDYT?

PS
Of course this PR is just POC. It requires to add more methods with this suffix

### DIFF
--- a/packages/core/src/types/Vector.ts
+++ b/packages/core/src/types/Vector.ts
@@ -229,23 +229,28 @@ export class Vector2 implements Type {
     );
   }
 
-  public mul(vector: Vector2) {
+  public mul(possibleVector: PossibleVector2) {
+    const vector = new Vector2(possibleVector);
     return new Vector2(this.x * vector.x, this.y * vector.y);
   }
 
-  public div(vector: Vector2) {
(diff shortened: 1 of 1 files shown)


aarthificial:
Proposals like this should be done via issues and not PRs:
https://github.com/motion-canvas/motion-canvas/blob/main/CONTRIBUTING.md#proposing-a-change

Instead of introducing a set of new methods, it would be better to extend the existing `add`, `sub`, `mul`, etc. methods to accept a `PossibleVector2` instead of just `Vector2`.
This would allow us to do things like:
```ts
vectorA.add(vectorB);
vectorA.add([100, 200]);
vectorA.add(30);
```

magx2:
Nice! I was missing this. So you think it's a good idea to go with? Should I've adjust the PR?

aarthificial:
Yes, please go ahead

magx2:
Is it OK? I'm not a JS programmer so feel free to leave me a comment

aarthificial (approved):

aarthificial:
Awesome, thanks for the PR!

Issue #499, opened 2023-03-09
Animation example has invalid source code link


Zombee8424:
**Description**
On the documentation page for [transitions](https://motioncanvas.io/docs/transitions/#pre-made-transitions), the animation example for "Pre-made transitions" links to an invalid page for the source code. I am not sure what page it should link to though, because it is showcasing multiple scenes.

PR #502, opened 2023-03-09
fix(docs): transitions example has invalid source code link (#499)

AshishBarvaliya:
An optional Link prop has been introduced to AnimationPlayer, to specify a separate source link in cases where the name of the example and scene file are not the same.

An alternative solution would be to ensure consistency by updating the file name "transitions-second" to simply "transitions".


aarthificial:
Thanks for the fix!
For the future, issues closed by a PR should be linked in the body of the commit, not in the summary:
```
Closes #<pr number>
```

PR #512, opened 2023-03-12
feat(ui): add `?present` url parameter

squigglesdev:
This PR adds a `?present` URL parameter that takes you directly to presentation mode.

### DIFF
--- a/packages/ui/src/main.tsx
+++ b/packages/ui/src/main.tsx
@@ -39,7 +39,6 @@ export function editor(project: Project) {
   const meta = project.meta;
 
   const queryString = window.location.search;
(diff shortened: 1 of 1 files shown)


packages/ui/src/main.tsx
```diff
@@ -31,10 +32,26 @@ export function editor(project: Project) {
...
   });
 
+
   const renderer = new Renderer(project);
   const presenter = new Presenter(project);
 
   const meta = project.meta;
+
+  const queryString = window.location.search;
+  console.log(queryString);
```
ksassnowski, line 42: 🤔

PR #514, opened 2023-03-13
feat(2d): add spline node

ksassnowski:
> **Note**
> This is technically still a draft request since I'm working on finishing up the last few loose ends. Most of the code is ready to be reviewed, however, and I want to get feedback on this as soon as possible. I expect that there will be quite a few change requests before this is ready to be merged. So it's better to get started now and not wait for me to finish writing the documentation, for instance.

_inhales_ Here we go.

This PR adds a `Spline` node which allows users to draw a smooth curve through a series of control points called **knots**.

## Usage

There are two ways to define the knots of a spline:

**1. Providing them via the `points` property**

```tsx
<Spline
  lineWidth={12}
  stroke={"lightseagreen"}
  points={[
    [-400, 0],
    [-200, -300],
    [0, 0],
    [200, -300],
    [400, 0],
  ]}
/>
```

**2. Providing `Knot` nodes as children**

```tsx
<Spline lineWidth={12} stroke={"lightseagreen"}>
  <Knot position={[-400, 0]} />
  <Knot position={[-200, -300]} />
  <Knot position={[0, 0]} />
  <Knot position={[200, -300]} />
  <Knot position={[400, 0]} />
</Spline>
```

Both examples would result in the same spline being drawn:

CleanShot 2023-03-13 at 10 54 19@2x

Providing the knots by using the `Knot` node also allows the user to specify the positions of the start and end handles of the knot via the `startHandle` and `endHandle` properties, respectively. The position of these handles are relative to the knot's position.

```tsx
<Spline lineWidth={12} stroke={'lightseagreen'}>
  <Knot position={[-400, 0]} />
  <Knot position={[-200, -300]} />
  <Knot
    position={[0, 0]}
    startHandle={[-100, 200]}
    endHandle={[100, 200]}
  />
  <Knot position={[200, -300]} />
  <Knot position={[400, 0]} />
</Spline>
```

CleanShot 2023-03-13 at 10 55 58@2x

### Mirrored handles and broken knots

If only one handle is provided, the other handle is mirrored by default. For example, when we specify a knot like this:

```tsx
<Knot startHandle={[-100, 100]} />
```

the position of `endHandle` will be `[100, -100]`. This works the same way when only an `endHandle` is provided.

```tsx
<Spline ref={spline} lineWidth={12} stroke={'lightseagreen'}>
  <Knot position={[-400, 0]} />
  <Knot position={[-200, -300]} />
  <Knot position={[0, 0]} startHandle={[100, 0]} />
  <Knot position={[200, -300]} />
  <Knot position={[400, 0]} />
</Spline>
```

CleanShot 2023-03-13 at 10 50 47@2x

When both handles are provided, the positions will be used as provided. This is called a _broken_ knot and can be used to create sharp corners in a spline.

```tsx
<Spline>
  <Knot position={[-400, 0]} />
  <Knot position={[-200, -300]} />
  <Knot
    position={[0, 0]}
    startHandle={[-100, 200]}
    endHandle={[100, 200]}
  />
  <Knot position={[200, -300]} />
  <Knot position={[400, 0]} />
</Spline>
```

CleanShot 2023-03-13 at 10 52 02@2x

### Auto-calculated handles

When no handles are provided for a knot, the handle positions get calculated automatically such that the spline passes smoothly through the knot. Check out [this link][auto-handles] for an explanation of the geometry behind this.

The "smoothness" of the resulting curve can be controlled with the `smoothness` property on the `Spline` node. This essentially controls the distance of the calculated handles from the knot itself.

```tsx
<Spline smoothness={0.2}>
  <Knot position={[-400, 0]} />
  <Knot position={[-200, -300]} />
  <Knot position={[0, 0]} />
  <Knot position={[200, -300]} />
  <Knot position={[400, 0]} />
</Spline>
```

CleanShot 2023-03-13 at 11 29 20@2x

It is also possible to blend between user-provided handles and auto-calculated handles by using the `auto` property on the `Knot` class. `auto` is a value between `0` and `1` and essentially serves as the `t` parameter to `Vector2.lerp`. `1` means we're using the auto handles and `0` means we're using the provided handles.

```tsx
<Knot
  startHandle={[-100, 200]}
  endHandle={[100, 200]}
  auto={0.5}
/>
```

CleanShot 2023-03-13 at 10 52 45@2x

`auto` is a compound signal, so it's also possible to control the `startHandleAuto` and `endHandleAuto` values individually.

```tsx
<Knot
  startHandle={[-100, 200]}
  endHandle={[100, 200]}
  startHandleAuto={0.5}
  endHandleAuto={0.2}
/>
```

CleanShot 2023-03-13 at 10 53 23@2x

Note that when `auto` is set to `1`, any transformation to `Knot` itself, such as rotation or scale, won't affect the handle position. This is because the auto handles get calculated purely based on the knot's position and its two neighboring knots' positions. This is something I'm not entirely sure about, but changing this behavior would require a significant change to the way the auto handles are calculated.

## Implementation

Wherever possible, I wanted to avoid reinventing solutions to problems that had already been solved in the `Line` class. A `Spline` should feel very similar to use than a `Line`, so things like arrow heads, `lineJoin`, and animating a spline should work the same way.

For this reason, `Spline` extends the `Line` node. The main task is then to calculate a `CurveProfile` for the spline that would allow us to reuse the `Line` node's existing drawing logic to do all the heavy lifting.

### Polynomials

The main architectural goals of this implementation was to both be compatible with the current `Line` implementation as well as be reusable enough to build other kinds of curves on top of it (like a standalone `CubicBezier` curve, for example). For this reason, this PR adds a few more data types to represent generic polynomials up to degree 3. All the math for these classes was basically taken verbatim from Freya Holmér's [Mathfs][mathfs] library and translated to Typescript.

- `Polynomial`
- `Polynomial2D`

Having these generic types allows to more easily create different kinds of spline segments down the line by specifying different coefficients. The Bézier curve specific implementations can be found in these classes: 

- `BezierCubic2D` for cubic Bézier curves
- `BezierQuad2D` for quadratic Bézier curves

The polynomial classes also allow us to reuse the math that stays identical across all types of polynomials such as:

- evaluating the polynomial for a given `x`
- calculating the derivative(s) at a given `x`
- splitting the polynomial into two separate polynomials that form the same overall shape

### Animating splines

To keep in line (hah!) with how the `Line` node works, the `start` and `end` properties of the spline are relative to the spline's arclength. This makes animating a spline at constant speed fairly tricky. There's no closed form solution for retrieving the point at given distance along a cubic Bézier curve's arclength. Instead, we approximate the points position using the `UniformPointSampler` class.

This class will sample a number of points along a spline segment by continuously evaluating the curve's polynomial for different `t` values. It then records the distance along the curve's arclength that the sampled point is at.

When looking up a point on the curve by distance, we find the two sampled points that are closed to the desired distance. We can then interpolate their original `t` values across the remaining distance to get an approximation of the "real" points `t` value. We can then simply evaluate the curve at this `t` value to get the final point.

The precision of this can obviously be increased by sampling more points, but I couldn't see any (visual) improvements by going beyond 20 sampled points.

## Open Tasks

- [ ] Write documentation
- [x] Calculate correct bounding box for spline

[mathfs]: https://github.com/FreyaHolmer/Mathfs
[auto-handles]: http://scaledinnovation.com/analytics/splines/aboutSplines.html


packages/2d/src/components/Line.ts
```diff
@@ -230,6 +228,16 @@ export class Line extends Shape {
     return box;
   }
 
+  protected validateProps(props: LineProps) {
```
ksassnowski, line 231: Extracted to a method so that `Spline` can override this as both the condition as well as the error message changes.

packages/2d/src/curves/BezierSegment.ts
```diff
@@ -0,0 +1,69 @@
...
+  ): [Vector2, Vector2, Vector2, Vector2] {
+    const startDistance = this.length * start;
+    const endDistance = this.length * end;
+    const startPoint = this.pointSampler.pointAtDistance(startDistance);
+    const endPoint = this.pointSampler.pointAtDistance(endDistance);
+    const endPointT = (endPoint.t - startPoint.t) / (1 - startPoint.t);
+
+    const [, startSegment] = this.bezier.split(startPoint.t);
+    const [segment] = startSegment.split(endPointT);
+    const points = segment.points;
```
aarthificial, line 49: I think you could skip a lot of work by checking if `start === 0 && end === 1`
Seems like that would be the case most of the time, and no splitting is necessary in that case.

packages/2d/src/curves/SerializedKnot.ts
```diff
@@ -0,0 +1,10 @@
+import {Vector2} from '@motion-canvas/core/lib/types';
+
+export type KnotAutoHandles = {start: number; end: number};
+
+export interface SerializedKnot {
```
aarthificial, line 5: I don't think `SerializedKnot` is a good name since this representation contains instances of other classes.
I'd go for `ParsedKnot` or `KnotInfo`

packages/2d/src/curves/CurveSampler.ts
```diff
@@ -0,0 +1,28 @@
...
+export interface SampledPoint {
+  t: number;
+  position: Vector2;
+  distance: number;
+  tangent: Vector2;
+  previous?: SampledPoint;
+  next?: SampledPoint;
+}
+
+export interface CurveSampler {
```
aarthificial, line 12: Interfaces for classes should be used only when there are two or more actual classes implementing them.
The fact that the only class implementing this is called `Uniform` makes me think that this interface is not necessary.

aarthificial (commented):
Both `Segment` and `Curve` seem to represent the same abstract entity.
Is there any reason for the current separation?
Right now we have two duplicated class hierarchies:

Curve | Segment
--- | ---
BezierCurve | BezierSegment
BezierCubic2D | CubicBezierSegment
BezierQuad2D | QuadBezierSegment

Where a lot of functionality in Segments is simply proxied from Curves.
For example `arcLength`, `points`, and `getPoint` (via `CurveSampler` but proxied nonetheless)

ksassnowski:
> Both `Segment` and `Curve` seem to represent the same abstract entity. Is there any reason for the current separation? Right now we have two duplicated class hierarchies:

The main idea was that the `Curve` are really just ways to deal with the mathematics of curves. Whereas `Segment` is more Motion Canvas specific as it includes things like drawing to a canvas. Stuffing it all into the `Segment` made it quite bloated and a bit all over the place. But I don't feel super strongly about this.

packages/2d/src/components/Knot.ts
```diff
@@ -0,0 +1,183 @@
...
+    // auto-calculated handles.
+    if (!hasStartHandle && !hasEndHandle) {
+      const hasStartHandleAuto = !this.startHandleAuto.isInitial();
+      const hasEndHandleAuto = !this.endHandleAuto.isInitial();
+
+      // If an `auto` value was provided, we interpret this to mean that the
+      // user wants to blend between the auto-calculated handles and a handle
+      // that sits at the origin of the knot.
+      // If no `auto` value is provided either, we simply set the `auto` value
+      // to 1 to use position of the auto-calculated handles.
```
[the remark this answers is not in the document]
ksassnowski: Wouldn't that mean that the user would always have to explicitly pass the `auto` parameter? This was meant as a shortcut for `auto = 1` when no handles are provided. 

We could just ignore the `auto` parameter if no handles are provided though, instead of implicitly treating the handles as `0` in that case.

packages/2d/src/components/Knot.ts
```diff
@@ -0,0 +1,183 @@
...
+        auto: {
+          start: autoStartHandle,
+          end: autoEndHandle,
+        },
+      };
+    }
+
+    const startHandle = hasStartHandle
+      ? this.startHandle()
+      : this.getDefaultStartHandle();
```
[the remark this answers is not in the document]
ksassnowski: You're right 🤦

packages/2d/src/components/Spline.ts
```diff
@@ -0,0 +1,262 @@
...
+  @initial(0.4)
+  @signal()
+  public declare readonly smoothness: SimpleSignal<number>;
+
+  public constructor(props: SplineProps) {
+    super(props);
+  }
+
+  @computed()
+  public override profile(): CurveProfile {
```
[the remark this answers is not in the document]
ksassnowski: Good to know! I also remove it for `childrenBBox` although that's probably gonna change anyways.

packages/2d/src/curves/BezierSegment.ts
ksassnowski: Good point. Changed.

packages/2d/src/curves/SerializedKnot.ts
ksassnowski: Agreed. I renamed it to `KnotInfo`

packages/2d/src/curves/CurveSampler.ts
ksassnowski: Fair enough. I removed the interface.

packages/2d/src/components/Knot.ts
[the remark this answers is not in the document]
aarthificial: >This was meant as a shortcut for auto = 1 when no handles are provided.

Maybe a better idea would be to only modify the property during initialization?
Something like:
```tsx
public constructor(props: KnotProps) {
  super(
    props.startHandle === undefined && props.endHandle === undefined
      ? {auto: 1, ...props}
      : props,
  );
}
```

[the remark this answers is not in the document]
ksassnowski: Yeah that seems to work fine. It also simplifies the `points` method quite a bit 👍

packages/2d/src/components/Line.ts
```diff
@@ -199,6 +193,10 @@ export class Line extends Shape {
       }
     }
 
+    if (this.closed()) {
```
ksassnowski, line 196: I just figured out that this won't work like this as it it's going to _always_ close the path even though that segment should not even be shown (because of `start`, `end`, or the offset). I'll have to figure something out there.

aarthificial:
> Stuffing it all into the Segment made it quite bloated and a bit all over the place.

I have a hard time believing this. The only functionality `Segment`s provide right now is the aforementioned `draw` method. 
Everything else is already in `Curve`s. So that's about 10 lines of code for `CubicBezierSegment` and 5 for `QuadBezierSegment`.
`BezierSegment` is a bit bigger but its counterpart, `BezierCurve`, implements practically no methods.

I get the whole separation between Math and drawing but I don't really see the practical reason for it.
At the end of the day, the `draw` method is just a converter that turns a segment into a Path2D representation. It doesn't even touch the canvas api.

ksassnowski:
Actually, looking at it now, the `BezierSegment` doesn't really have anything to do with Bézier curves anymore since it just takes a `Polynomial2D` as input. Maybe we should rename this to something like `CurveSegment` so it can be reused for any kind of segment that can be expressed as a polynomial?

packages/2d/src/components/Line.ts
ksassnowski: Nevermind, I just need to add `this.end() === 1` to the condition.

aarthificial:
> Maybe we should rename this to something like PolynomialSegment so it can be reused for any kind of segment that can be expressed as a polynomial?

I agree with you

packages/2d/src/components/Knot.ts
```diff
@@ -0,0 +1,183 @@
...
+   * the handles gets calculated automatically to create smooth curve through
+   * the knot. The smoothness of the resulting curve can be controlled via the
+   * {@link Spline.smoothness} property.
+   *
+   * It is also possible to blend between a user-defined position and the
+   * auto-calculated position by using the {@link auto} property.
+   *
+   * @defaultValue Mirrored position of the endHandle.
+   */
+  @initial(() => ((<unknown>this) as Knot).getDefaultStartHandle)
```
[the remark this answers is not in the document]
aarthificial: I'm 90% sure this won't work.
After compilation, this decorator will be turned into a call under the class:
```ts
__decorate([
    initial(() => this.getDefaultStartHandle),
    wrapper(Vector2)
    signal()
], Knot.prototype, "startHandle", void 0);
```
How would `this` be available in this context?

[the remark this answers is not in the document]
ksassnowski: Seems like you're right, this is working the exact same way even if I remove the `@initial` decorator. It looks like the `getDefaultStartHandle` method was working all this time.

ksassnowski:
Renamed `BezierSegment` to `PolynomialSegment`. As we talked about, I also updated `Segment.getPoint` to return a `CurvePoint` instead.

ksassnowski:
Bounding boxes not get calculated correctly, so the code side if this PR is complete and ready for review. I'm getting started on the documentation later today or tomorrow.

packages/2d/src/components/Spline.ts
```diff
@@ -0,0 +1,260 @@
...
+        inspect: this.key,
+      });
+    }
+  }
+
+  protected override desiredSize(): SerializedVector2<DesiredLength> {
+    const bounds = (this.profile().segments as PolynomialSegment[]).map(
+      segment => segment.getBBox(),
+    );
+    return BBox.fromBBoxes(...bounds);
```
aarthificial, line 143:
```suggestion
    return BBox.fromBBoxes(...bounds).size;
```
`desiredSize` should return a vector, if you return a bbox, it will use its position instead of its size.

packages/2d/src/components/Spline.ts
```diff
@@ -0,0 +1,260 @@
...
+      for (const point of [startHandle, endHandle]) {
+        if (point.magnitude > 0) {
+          moveTo(context, point);
+          context.beginPath();
+          arc(context, point, 4);
+          context.closePath();
+          context.fill();
+          context.stroke();
+        }
+      }
```
aarthificial, line 238: Maybe we could differentiate visually mirrored handles from normal ones?

aarthificial (changes requested):
You'll also need to override the `getComputedLayout` because right now the spline is not placed correctly within a layout:
image
Take a look at how `Line` overrides it. It offsets it's position so that `childrenBBox` is always in the center. Here you'll have to do something similar using `BBox.fromBBoxes(...bounds)`

Repro:
```tsx
import {Knot, Rect, Spline} from '@motion-canvas/2d/lib/components';

  const ref = createRef<Spline>();

  view.add(
    <Rect ref={rect} fill={'#242424'} layout>
      <Spline ref={ref} lineWidth={12} stroke={'lightseagreen'}>
        <Knot position={[-400, 0]} />
        <Knot position={[-200, -200]} endHandle={[100, -500]} />
        <Knot position={[0, 0]} />
        <Knot position={[200, -300]} />
        <Knot position={[400, 0]} />
      </Spline>
    </Rect>,
  );

  yield* waitFor(5);
});
```

packages/2d/src/components/Spline.ts
ksassnowski: As discussed privately, this is something we might add at A Later Date™️

ksassnowski:
The layout should not get computed properly.

aarthificial (approved):

Issue #256, opened 2023-02-09
Consistent script names


hhenrichsen:
**Description**
This is super silly, but it'd be nice to have consistent script names between the packages. Right now some use `start`, some use `serve`, 
and some use `watch` for their dev mode scripts, and I have trouble remembering which is which.

**Proposed solution**
Change the scripts to consistently use one of the above options. `build` is consistent already; is there a reason that the other shared script intent should not be?

aarthificial:
docs use `start` because [in docusaurus `serve` means something different](https://docusaurus.io/docs/cli#docusaurus-serve-sitedir) - serving a preview build of the website. 
Compiled packages, such as `core`, `2d`, and `vite-plugin` use `watch` because they run the typescript compiler in watch mode.
Vite-based packages use `serve` because they boot up a development server.
The main reason `build` is consistent is because Lerna requires that.

TLDLR, the current naming convention reflects what is actually run in the given package. 
We can change all development scripts to something generic like `dev`, I don't really mind.

aarthificial:
Closed by #748

PR #516, opened 2023-03-13
Add consistency for the script names

guifeliper:
Resolving issue #256, I have added the script called "dev" for each package that runs on watch mode.


aarthificial:
#256 states:
> We can change all development scripts to something generic like `dev`

I don't think introducing duplicated scripts helps. If we want to use `dev` we should replace the current ones together with the root `package.json`

aarthificial (changes requested):

aarthificial:
Closing in favor of #748

Issue #519, opened 2023-03-13
Fix version comparison


aarthificial:
**Describe the bug**
Version comparison that decides when to display the "UPDATE AVAILABLE" hint uses string comparison:
https://github.com/motion-canvas/motion-canvas/blob/a23d06cbf6e29f37a9259e50fe71c482640b83fb/packages/ui/src/components/footer/Versions.tsx#L10
This means that `3.13.0` is less than `3.3.0` which is not valid.

**Expected behavior**
Versions should be correctly parsed and compared according to the semver standard.

**Package versions:**
- ui: 3.2.1

PR #520, opened 2023-03-13
Fix(ui): version comparison issue

AshishBarvaliya:
closes (#519)

Added compareVersions function.


aarthificial (changes requested):
This doesn't properly handle [prereleases](https://semver.org/#spec-item-9):
```
compareVersions('3.1.0-alpha.1', '3.1.0') // returns 1, should be -1
```

aarthificial (approved):

PR #525, opened 2023-03-14
fix(2d): clip Txt by default

### LINKED ISSUE #465: Txt height slices overflow inconsistently
**Describe the bug**
If `Txt` element has height, which is lower than required to display whole text, it slices overflow inconsistently:
- If text is animated, overflow is cut.
- If text is not being animated, overflow is displayed.

https://user-images.githubusercontent.com/822287/222658306-19f624b9-679b-42db-b8e5-40cecd98344c.mp4

**To Reproduce**
Steps to reproduce the behavior
or a code snippet containing the scene that causes the problem:

```ts
import { makeScene2D } from "@motion-canvas/2d/lib/scenes";
import { Line, Node, Txt } from "@motion-canvas/2d/lib/components";
import { waitFor } from "@motion-canvas/core/lib/flow";
import { createRef } from "@motion-canvas/core/lib/utils";

  const textRef = createRef<Txt>();

  view.add(
    <Node>
      <Txt
        ref={textRef}
        text="Test / Text"
        fontFamily="Tilt Neon"
        fontSize={128}
        height={128}
        fill="#000"
        opacity={0}
      />
      <Line
        endArrow
        lineWidth={8}
        stroke="red"
        opacity={0.7}
        points={[
          [200, 200],
          [-15, 80],
        ]}
      />
    </Node>
  );

  // Animate text
  yield* textRef().opacity(1, 2);

  // Wait
  yield* waitFor(2);
});
```

Look at bottom of '/'. It is rendered outside of text height.

Font used in example is "Tilt Neon" from Google Fonts.

- core: 3.0.0
- ui: 3.0.1

ribru17:
Clip Txt by default to prevent confusing behavior where text clips only while animating. I only enacted this change for text height as I didn't know whether it would potentially affect text wrapping in the future.

Fixes #465

### DIFF
--- a/packages/2d/src/components/Txt.ts
+++ b/packages/2d/src/components/Txt.ts
@@ -92,6 +92,9 @@ export class Txt extends Shape {
     const y = box.y + box.height / 2;
     context.save();
     context.textBaseline = 'middle';
+    const clipPath = new Path2D();
+    clipPath.rect(box.x, box.y, box.width, this.getHeight());
+    context.clip(clipPath);
 
     if (this.lineWidth() <= 0) {
       context.fillText(text, box.x, y);


packages/2d/src/components/Txt.ts
```diff
@@ -92,6 +92,9 @@ export class Txt extends Shape {
     const y = box.y + box.height / 2;
     context.save();
     context.textBaseline = 'middle';
+    const clipPath = new Path2D();
+    clipPath.rect(box.x, box.y, box.width, this.getHeight());
+    context.clip(clipPath);
```
aarthificial, line 97: Clipping should be done in the `draw` method. `drawText` is called for each text segment.

`this.computedSize()` should be used to acquire the bounding box, analogically to how `Layout` does it.

aarthificial (changes requested):
Text still isn't clipped in the same way as when using cache.

Example with text with limited width:
image
```ts
view.add(
  <>
    <Txt fill={'white'} x={-200} width={100} cache>
      Lorem ipsum
    </Txt>
    <Txt fill={'white'} x={200} width={100}>
      Lorem ipsum
    </Txt>
  </>,
);
````

Example with wrapped text:
image
```ts
view.add(
  <>
    <Txt fill={'white'} x={-200} width={320} height={200} textWrap cache>
      Lorem ipsum dolor sit amet, consectetur adipisicing elit. Est harum
      illum, ipsam iure laborum modi molestiae mollitia provident quod
      sapiente! Aperiam asperiores beatae fuga iusto nostrum numquam odit
      totam, vitae!
    </Txt>
    <Txt fill={'white'} x={200} width={320} height={200} textWrap>
      Lorem ipsum dolor sit amet, consectetur adipisicing elit. Est harum
      illum, ipsam iure laborum modi molestiae mollitia provident quod
      sapiente! Aperiam asperiores beatae fuga iusto nostrum numquam odit
      totam, vitae!
    </Txt>
  </>,
);
```

aarthificial:
This PR only partially resolves #465, as the `clip` property is still not being respected by the `Txt` node

aarthificial:
> to prevent confusing behavior

I think this is still a confusing behavior. Ideally, the cache size should be increased so that the text never gets clipped, even when lowering the opacity. The `clip` property would then enable clipping like in all other Shape nodes.

ribru17:
Thanks for the feedback. I'll start working on changes

ksassnowski:
@ribru17 It's been a while since the last review. Are you still planning to continue with this PR?

ribru17:
@ksassnowski Sorry I've been really busy I don't know if I will be able to continue any time soon, I'm open to someone else taking a look. I might not have time to dedicate to this for a couple months.

nishchay7pixels:
Picking this one up if everyone agree.

rmhartog:
> > to prevent confusing behavior
> 
> I think this is still a confusing behavior. Ideally, the cache size should be increased so that the text never gets clipped, even when lowering the opacity. The `clip` property would then enable clipping like in all other Shape nodes.

Based on this comment, I've created a PR for the cache bounding box of the `Txt` component in #836.

PR #530, opened 2023-03-15
feat(cli): add automatic project creation

Thomsr:
This pull request adds a new package: `cli`. This package contains all the command line interfaces that Motion Canvas has.

With this pull request comes the ability to type `mc` into your terminal and get prompted with several options which consist of the following:
```
mc
├── add
     ├── scene
     ├── project
     ├── component
```
This allows the user to create a new scene, project or component with ease.

### DIFF
--- a/packages/cli/add/commands/project.js
+++ b/packages/cli/add/commands/project.js
@@ -1,0 +1,87 @@
+import prompts from 'prompts';
+import path from 'path';
+import fs from 'fs';
+import ts from 'typescript';
+
+import {projectTemplate} from '../templates.js';
+import {createDir} from '../../utils/createDir.js';
+import {createFile} from '../../utils/createFile.js';
+
+export async function addProject() {
+  const response = await prompts({
+    type: 'text',
+    name: 'projectName',
+    message: 'Project Name',
+    initial: 'myProject',
+    validate: value =>
+      isValidProjectName(value) ? 'Project already exists.' : true,
+  });
+  await createProject(response.projectName);
+}
+
+/* Is it OK to check if the project exists using this? */
+function isValidProjectName(projectName) {
+  return fs.existsSync(path.resolve('src', `${projectName}.ts`));
+}
+
+function isProjectArray(node) {
+  if (!ts.isArrayLiteralExpression(node)) return false;
+  const projectPropertyAssignment = node.parent;
+  if (
+    !(
+      projectPropertyAssignment &&
+      ts.isPropertyAssignment(projectPropertyAssignment) &&
+      ts.isIdentifier(projectPropertyAssignment.name) &&
+      projectPropertyAssignment.name.escapedText === 'project'
+    )
+  )
+    return false;
+  return true;
+}
+
+async function createProject(projectName) {
+  await createFile(path.resolve('src', `${projectName}.ts`), projectTemplate());
+  await createDir(path.resolve('src', 'scenes', projectName));
+
+  const transformer = context => node => {
+    const visitor = node => {
+      if (isProjectArray(node)) {
+        const newElement = ts.factory.createStringLiteral(
+          `./src/${projectName}.ts`,
+        );
+        return ts.factory.updateArrayLiteralExpression(node, [
+          ...node.elements,
+          newElement,
+        ]);
+      }
+      return ts.visitEachChild(node, child => visitor(child), context);
+    };
+
+    const rootVisitor = node => {
+      node = ts.visitNode(node, visitor);
+      return node;
+    };
+
+    return ts.visitNode(node, rootVisitor);
+  };
+
+  // Is this a correct way of finding the right extension?
+  let language = 'ts';
+  if (fs.existsSync('vite.config.js')) language = 'js';
+
+  const src = ts.createSourceFile(
+    '',
+    fs.readFileSync(`vite.config.${language}`, 'utf8'),
+    ts.ScriptTarget.Latest,
+    true,
+  );
+  const transformed = ts.transform(src, [transformer]);
+  const newSrc = transformed.transformed[0];
+
+  const printer = ts.createPrinter({newLine: ts.NewLineKind.LineFeed});
+  const result = printer.printNode(ts.EmitHint.Unspecified, newSrc, newSrc);
+
+  fs.writeFile(`vite.config.${language}`, result, e => {
+    if (e) console.error(e);
+  });
+}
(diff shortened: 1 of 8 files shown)


Thomsr:
My current testing setup is as follows:
```
test/
└──src/
     ├── scenes/
     └── vite.config.ts
```

And then call `node ..\packages\cli\index.js` from inside the `test/` directory

I am open to any feedback you might have, please dont hesitate.

packages/cli/utils/createDir.js
```diff
@@ -0,0 +1,10 @@
+import fs from 'fs';
+import kleur from 'kleur';
+
+export async function createDir(path) {
+  await fs.mkdir(path, e => {
+    if (e) {
+      console.error(kleur.red(`Could not create file ${path}\n${e}`));
+    }
```
aarthificial, line 8: I'd recommend throwing errors with the messages to display and catching them in one place.

packages/cli/package.json
```diff
@@ -0,0 +1,33 @@
...
+    ],
+    "repository": {
+      "type": "git",
+      "url": "https://github.com/motion-canvas/motion-canvas.git"
+    },
+    "devDependencies": {
+      "@motion-canvas/2d": "^3.2.0",
+      "@motion-canvas/core": "^3.2.0",
+      "@motion-canvas/ui": "^3.2.1",
+      "@motion-canvas/vite-plugin": "^3.2.0"
```
aarthificial, line 27: Are these required?

packages/cli/add/commands/project.js
```diff
@@ -0,0 +1,87 @@
...
+      node = ts.visitNode(node, visitor);
+      return node;
+    };
+
+    return ts.visitNode(node, rootVisitor);
+  };
+
+  // Is this a correct way of finding the right extension?
+  let language = 'ts';
+  if (fs.existsSync('vite.config.js')) language = 'js';
```
aarthificial, line 70: Let's make an array of "possible config file names", do `readdir` in the current directory, and try to find the first match. Users can use custom names for their configs, this way it will be easier to support that in the future.

packages/cli/add/commands/project.js
```diff
@@ -0,0 +1,87 @@
...
+    '',
+    fs.readFileSync(`vite.config.${language}`, 'utf8'),
+    ts.ScriptTarget.Latest,
+    true,
+  );
+  const transformed = ts.transform(src, [transformer]);
+  const newSrc = transformed.transformed[0];
+
+  const printer = ts.createPrinter({newLine: ts.NewLineKind.LineFeed});
+  const result = printer.printNode(ts.EmitHint.Unspecified, newSrc, newSrc);
```
aarthificial, line 82: All of these steps may throw errors, would be nice to handle them in some way.

packages/cli/add/commands/project.js
```diff
@@ -0,0 +1,87 @@
...
+  if (
+    !(
+      projectPropertyAssignment &&
+      ts.isPropertyAssignment(projectPropertyAssignment) &&
+      ts.isIdentifier(projectPropertyAssignment.name) &&
+      projectPropertyAssignment.name.escapedText === 'project'
+    )
+  )
+    return false;
+  return true;
```
aarthificial, line 39: This could be simplified to:
```suggestion
  return (
    ts.isArrayLiteralExpression(node) &&
    node.parent &&
    ts.isPropertyAssignment(node.parent) &&
    ts.isIdentifier(node.parent.name) &&
    node.parent.name.escapedText === 'project'
  );
```

packages/cli/add/commands/project.js
```diff
@@ -0,0 +1,87 @@
...
+    name: 'projectName',
+    message: 'Project Name',
+    initial: 'myProject',
+    validate: value =>
+      isValidProjectName(value) ? 'Project already exists.' : true,
+  });
+  await createProject(response.projectName);
+}
+
+/* Is it OK to check if the project exists using this? */
```
aarthificial, line 22: Yes

aarthificial (changes requested):
I think we should use typescript here.
The `create` package is kept extremely simple with as little overhead as possible because it needs to be snappy - it's downloaded on demand when `npm init` is used.
Here, the package will be actually installed globally by the user, so we can have more freedom.

packages/cli/package.json
aarthificial: They're not required
`create` uses them to know what are the newest versions but here they're not necessary here.

packages/cli/add/commands/project.js
aarthificial: Yes but these loops can be written as:
```ts
const match = possibleConfigFileNames.find(name => files.includes(name));
```
`match` will then be `string` or `undefined` if no match was found

Thomsr: I knew there would be something more elegant, just could not find the right way of doing it. Thanks :)

packages/cli/index.js
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env node
+
+import prompts from 'prompts';
+
+import {Add} from './add/index.js';
+
+(async () => {
```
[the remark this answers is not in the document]
aarthificial: You can try to load `package.json` and see if `@motion-canvas/core` is defined in the dependencies

packages/cli/add/commands/project.js
```diff
@@ -0,0 +1,87 @@
...
+      ts.isIdentifier(projectPropertyAssignment.name) &&
+      projectPropertyAssignment.name.escapedText === 'project'
+    )
+  )
+    return false;
+  return true;
+}
+
+async function createProject(projectName) {
+  await createFile(path.resolve('src', `${projectName}.ts`), projectTemplate());
```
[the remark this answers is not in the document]
aarthificial: Yes, but I think this check should be performed at the very beginning when testing if the project is a motion canvas one.
Generally, at the beginning we should figure out these three things:
1. is this a motion canvas project? (abort otherwise)
2. can we locate the config file? (store the path for later)
3. what language does this project use? (store that for later too)

packages/cli/index.js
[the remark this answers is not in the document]
Thomsr: This is what I came up with

```ts
function isMotionCanvasProject(): boolean{
  const dependencies = JSON.stringify(JSON.parse(fs.readFileSync(path.resolve('package.json')).toString()).dependencies);
  if(dependencies == undefined) throw new Error(kleur.red('Unable to locate dependencies in package.json'));
  return dependencies.includes("@motion-canvas/core");
}
```
I dont know if this is the most efficient way of handling this, but it seems to work.

Thomsr:
I will add the `component` command later, I think that just `scene` and `project` will suffice for now.

PR #538, opened 2023-03-16
fix(vite-plugin): can't assign port

### LINKED ISSUE #527: Can't assign port
**Describe the bug**

Normally, you can overwrite the port in the `vite.config.ts` file with

```
export default defineConfig({
  plugins: [motionCanvas()],
  server: {
    port: 3000
  }
});
```

Or just run `vite --port 3000`

But that doesn't work and always starts on port 9000

**To Reproduce**

Just follow the instructions

**Expected behavior**

The application runs on port 3000

- core: ^3.2.0
- ui: ^3.2.1

AshishBarvaliya:
closes #527


packages/vite-plugin/src/main.ts
```diff
@@ -384,7 +384,7 @@ export default ({
           },
         },
         server: {
-          port: 9000,
+          port: config?.server?.port || 9000,
```
aarthificial, line 387: We should use nullish coalescing whenever applicable:
```suggestion
          port: config?.server?.port ?? 9000,
```
I know the port can't be `0` in this specific case but `??` better conveys the meaning of the code (if `port` is not defined, use `9000`)

aarthificial (changes requested):

aarthificial (approved):

aarthificial:
Awesome, thanks!

PR #539, opened 2023-03-16
feat(2d): support Latex tweening

### LINKED ISSUE #482: Support LaTex Tweening
**Description**
Math video need LaTex Tweening like in Manim.

**Proposed solution**
Parse MathJax rendered SVG and render each shape from SVG manually. Then we can animate each shape.

To tween LaTex we can use this function:

```js
yield * tex().tweenTex(target tex, duration)
```

**Considered alternatives**
Use external library for doing it.

**Additional context**

I have been working on this issue but PR are closed.

levirs565:
This PR add tweening support for Latex component. This PR will  fix #482 

## How it Work?

Latex component now inherit SVG component. SVG component can tween between two SVG. Latex component just tweening between two rendered SVG from two Latex.

SVG component parse SVG to extract all shapes. All SVG shapes then converted into Motion Canvas Shapes and added into SVG component as children. All shapes is not directly inserted into SVG component instead there a Node component that wrap all SVG shapes. This Node component is first children of SVG component. This Node is used to make Layout component respect SVG scaling.

## Changed Behavior

Old behavior: Latex component's size is depend on custom width or custom height
New behavior: Latex component's size is depend on font size. With this, With this, we can synchronize Txt size and Latex size more easily.

## Latex Tweening

Latex can be splitted into subtex. To make subtex we just enclose part of tex with `{{}}`. Why?. The reason is to make tweening more flexible. User can choose where transformed, inserted or deleted subtex.  

Latex `svg` propery can be string or array of string. When it string, it will be parsed to list of substring.

## How to Use?

To use Latex tweening we can use tween `tex` property as normal.

## Examples and Result

This is example usage of Latex tweening:

```js
import {Latex, Rect} from '@motion-canvas/2d/lib/components';
import {makeScene2D} from '@motion-canvas/2d/lib/scenes';

  const tex = createRef<Latex>();
  view.add(
    <Rect layout>
      <Latex
        ref={tex}
        tex={['x', '=', '{{\\sin \\left(}} \\pi {{\\over}} 2 {{\\right)}}']}
        fill="white"
      />
    </Rect>,
  );

  yield* waitFor(2);
  yield* tex().opacity(0, 1);
  yield* tex().opacity(1, 1);
  yield* waitFor(2);
  yield* tex().tex(
    ['x', '=', '{{\\sin \\left(}} \\pi {{\\over}} 2 {{\\right)}}', '+', '3'],
    1,
  );
  yield* tex().tex(
    [
      'x',
      '=',
      '1',
      '+',
      '{{\\sin \\left(}} \\pi {{\\over}} 2 {{\\right)}}',
      '+',
      '3',
    ],
    1.2,
  );
  yield* tex().tex(
    [
      'x',
      '=',
      '5',
      '+',
      '9',
      '+',
      '{{\\sin \\left(}} \\pi {{\\over}} 2 {{\\right)}}',
      '+',
      '7',
      '+',
      '8',
    ],
    1.2,
  );
  yield* tex().tex(
    [
      'x',
      '=',
      '5',
      '+',
      '2',
      '+',
      '9',
      '+',
      '{{\\sin \\left(}} \\pi {{\\over}} 2 {{\\right)}}',
      '+',
      '7',
      '+',
      '1',
      '+',
      '8',
    ],
    1.2,
  );
  yield* tex().tex(
    [
      'x',
      '=',
      '7',
      '+',
      '1',
      '+',
      '8',
      '+',
      '{{\\sin \\left(}} \\pi {{\\over}} 2 {{\\right)}}',
      '+',
      '5',
      '+',
      '2',
      '+',
      '9',
    ],
    1.2,
  );
  yield* tex().tex(
    [
      'x',
      '=',
      '3',
      '+',
      '{{\\sin \\left(}} \\pi + 3 {{\\over}} 2 {{\\right)}}',
      '+',
      '1',
    ],
    1.2,
  );
  yield* tex().tex(
    [
      'x',
      '=',
      '{{\\sin \\left(}} \\pi + 3 {{\\over}} 2 {{\\right)}}',
      '+',
      '3',
      '+',
      '1',
    ],
    1.2,
  );
  yield* tex().tex(
    ['x', '=', '{{\\sin \\left(}} \\pi {{\\over}} 2 {{\\right)}}'],
    1.2,
  );
  yield* waitFor(2);
});

```

Here is the result:

https://user-images.githubusercontent.com/42236775/228152134-34936737-5058-4957-94c4-c1c672653221.mp4


packages/2d/src/components/SVG.ts
```diff
@@ -0,0 +1,387 @@
...
+    const size = new Vector2(
+      svgRoot.width.baseVal.value,
+      svgRoot.height.baseVal.value,
+    );
+    const scale = size.div(viewBox.size);
+    const center = viewBox.center;
+
+    const rootTransform = new DOMMatrix()
+      .scale(scale.x, scale.y)
+      .translate(-center.x, -center.y);
```
aarthificial, line 85: Could be:
```suggestion
      .scaleSelf(scale.x, scale.y)
      .translateSelf(-center.x, -center.y);
```
to not create multiple instances of the matrix.

packages/2d/src/components/SVG.ts
```diff
@@ -0,0 +1,387 @@
...
+    const props: ShapeProps = {
+      position: node.position(),
+      scale: node.scale(),
+      rotation: node.rotation(),
+    };
+    if (node instanceof Layout) {
+      props.size = node.size();
+    }
+    return node.clone(props);
+  }
```
aarthificial, line 315: Couldn't we use `snapshotClone()` here?

packages/2d/src/components/SVG.ts
```diff
@@ -0,0 +1,387 @@
...
+            Vector2.lerp(node.from.position(), node.to.position(), eased),
+          );
+          node.current.scale(
+            Vector2.lerp(node.from.scale(), node.to.scale(), eased),
+          );
+          node.current.rotation(
+            easeInOutSine(remapped, node.from.rotation(), node.to.rotation()),
+          );
+
+          if (node.current instanceof Rect) {
```
aarthificial, line 351: Shouldn't it be `Layout`?
```suggestion
          if (node.current instanceof Layout) {
```

packages/2d/src/components/SVG.ts
```diff
@@ -0,0 +1,387 @@
...
+          );
+
+        const deletedOpacity = clampRemap(0, beginning + overlap, 1, 0, value);
+        for (const node of diff.deleted) node.opacity(deletedOpacity);
+
+        const insertedOpacity = clampRemap(ending - overlap, 1, 0, 1, value);
+        for (const node of diff.inserted) node.opacity(insertedOpacity);
+      },
+      () => {
+        this.wrapper.children(this.parsedNodes);
```
aarthificial, line 380: Ideally, we should also call `dispose` on the removed nodes and the clones used for transformed nodes

packages/2d/src/components/SVG.ts
```diff
@@ -0,0 +1,387 @@
...
+        data,
+        ...style,
+      });
+      const center = path.getPathBBox().center;
+      this.applyTransformToShape(
+        path,
+        transformMatrix.translate(center.x, center.y),
+      );
+      yield path;
+    } else if (child.tagName == 'rect') {
```
aarthificial, line 211: What about `circle`, `ellipse` and `image`?

packages/2d/src/components/SVG.ts
```diff
@@ -0,0 +1,387 @@
...
+  public *tweenSVG(svg: string, time: number) {
+    const newSVG = this.parseSVG(svg);
+    const diff = this.diffSVG(this.parsed(), newSVG);
+    const transformed = diff.transformed.map(({from, to}) => ({
+      from: this.cloneNodeExact(from),
+      current: from,
+      to,
+    }));
+
+    for (const node of diff.inserted) this.wrapper.add(node);
```
aarthificial, line 326: Will this insert them in the correct order?
I know that we reset the wrapper children at the end of the tween but if there are overlapping elements in the svg couldn't that cause visible glitches during tweening?

packages/2d/src/components/SVG.ts
```diff
@@ -0,0 +1,387 @@
...
+        for (const node of transformed) {
+          node.current.position(
+            Vector2.lerp(node.from.position(), node.to.position(), eased),
+          );
+          node.current.scale(
+            Vector2.lerp(node.from.scale(), node.to.scale(), eased),
+          );
+          node.current.rotation(
+            easeInOutSine(remapped, node.from.rotation(), node.to.rotation()),
+          );
```
aarthificial, line 349: Shouldn't we also tween `fill` and other style properties?

aarthificial (changes requested):
I left some initial comments just by looking at the code.
Unfortunately I wasn't able to run the project locally as it throws an error:
```
Failed to read the 'value' property from 'SVGLength': Could not resolve relative length.
```
It seems that the svg generated by latex uses `ex` units, and since the container is not added to the document it cannot resolve them.

levirs565:
> Failed to read the 'value' property from 'SVGLength': Could not resolve relative length.

I have been fixed this issue. This issue is only occured in Chrome.

levirs565:
> Would it be possible to do manim-style [write-in animations](https://youtu.be/iydG-e1dQGA?t=46)?

This is possible. For example to create write in animation for Path we can use these two solution.

First solution is to use this : https://css-tricks.com/svg-line-animation-works/

Second solution is to parse path, convert path into Segment and draw all Segment. Then we can animate start and end position of path to draw.

packages/2d/src/components/Latex.ts
```diff
@@ -42,45 +35,36 @@ export class Latex extends Img {
...
+    const svg = Adaptor.innerHTML(JaxDocument.convert(tex, this.options));
+    Latex.svgContentsPool[src] = svg;
+    return svg;
+  }
 
-    return image;
+  public *tweenTex(tex: string, time: number) {
+    const newSVG = this.latexToSVG(tex);
+    yield* this.tweenSVG(newSVG, time);
+    this.tex(tex);
```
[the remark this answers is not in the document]
levirs565: I have rewrite the function. However, the tweening look faster than before.

packages/2d/src/components/SVG.ts
levirs565: No couldn't. `snapshotClone()` does not keep position, scale and rotation of node.

packages/2d/src/components/SVG.ts
levirs565: It must be `Shape` not `Rect` or `Layout` because all node is actually `Shape`.

levirs565: The if statement actually usable. I have been rewrite this.

packages/2d/src/components/SVG.ts
levirs565: We should. However, i don't know how to tween `PossibleCanvasStyle`. How to tween `PossibleCanvasStyle`

packages/2d/src/components/SVG.ts
levirs565: This PR only focus on Latex tweening. Latex SVG does not contain `cirlce`, `ellipse` or `image`. In my opinion, implementing shape other than used in Latex must be in another PR.

packages/2d/src/components/SVG.ts
```diff
@@ -0,0 +1,387 @@
...
+    return this.parseSVG(this.svg());
+  }
+
+  @computed()
+  private parsedNodes() {
+    return this.parsed().nodes;
+  }
+
+  private parseSVG(svg: string): ParsedSVG {
+    const container = document.createElement('div');
```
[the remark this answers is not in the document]
levirs565: Yes, can. However, `parseSVG` depend on `parseColor` that depend on `this`.  To make this function static we must rewrite the function.

packages/2d/src/components/Latex.ts
```diff
@@ -42,45 +35,36 @@ export class Latex extends Img {
   public declare readonly tex: SimpleSignal<string, this>;
 
   public constructor(props: LatexProps) {
-    super(props);
+    super({
+      fontSize: 48,
```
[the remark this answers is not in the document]
levirs565: When `fontSize` not set this error will occur:

```
A circular dependency occurred between signals.
```

packages/2d/src/components/SVG.ts
```diff
@@ -0,0 +1,387 @@
...
+      y: custom.y ?? y,
+    };
+  }
+
+  @computed()
+  private parsed() {
+    return this.parseSVG(this.svg());
+  }
+
+  @computed()
```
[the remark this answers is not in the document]
levirs565: The `computed()` decorator is used to bind this function to SVG component. Are bounding this function into SVG component manually are better than using this decorator?

levirs565:
I have changed SVG tweening process and diff algorithm. Now SVG tweening use animation flow to tween children. Diff algorithm mimic Manim `TransformMatchingShape` and `TransformMatchingTex`. The algorithm use id to check two node equality.

I also introduced subtex in Latex component. With subtex, user can choose where subtex to be deleted, inserted or transformed.

aarthificial:
I converted the PR to a draft since you keep making changes to it.
Let me know once the PR is ready to be reviewed.

levirs565:
My PR is ready. Please review it again. I will not create changes again.

levirs565:
SVG node now has support for `circle`, `ellipse`, `line`, `polyline`, `polygon` and `image` element.
When SVG Tweening happen, SVG will inserted new Shape at correct order. 
My PR is ready. Please review it again.

ksassnowski:
@levirs565 First of all, thanks a lot for all the effort you put into this, we really appreciate it!

The problem is that because of the scope of this PR—in LoC, new concepts added, as well as its general complexity—it is very difficult to review in its current shape. So ideally, it should be split into two separate PRs:

1. One PR that just adds the `SVG` component and related functionality
2. A follow-up PR once the `SVG` component got merged that builds on top of it and adds support for Latex tweening

This would allow us to review each part individually and make the whole process a lot more manageable.

So it would be great if you could extract all code related to _just_ the `SVG` component into a separate branch and open a PR for that one first. Thanks!

levirs565:
In my opinion this PR ideally splitted into 3 PR

- First, Path component
- SVG component
- Last, SVG component

ksassnowski:
Sure, if you think it makes more sense to split it into three parts then we can also do that 👍

Issue #297, opened 2023-02-10
Multiple projects - use the name from `makeProject` instead of the filename


DanielBreiner:
Currently, the filename from which the project is being loaded is used as the project name. I propose using the `name` field in `makeProject`. Also, consider showing a warning when two projects have the same name.

![image](https://user-images.githubusercontent.com/43221175/218133084-30c1abc6-7888-4d49-acba-b3860d47a64e.png)

Repro:
```ts
// vite.config.ts

import { defineConfig } from "vite";
import motionCanvas from "@motion-canvas/vite-plugin";

export default defineConfig({
  plugins: [
    motionCanvas({
      project: [
        "./src/projects/project1/project.ts",
        "./src/projects/project2/project.ts",
      ],
    }),
  ],
});
```
```ts
// src/projects/project1/project.ts

import { makeProject } from "@motion-canvas/core";
import "../../global.css";
import main from "./scenes/main?scene";

export default makeProject({
  name: "project1",
  scenes: [main],
  background: "#141414",
});
```
```ts
// src/projects/project2/project.ts

import { makeProject } from "@motion-canvas/core";
import "../../global.css";
import main from "./scenes/main?scene";

export default makeProject({
  name: "project2",
  scenes: [main],
  background: "#141414",
});
```

aarthificial:
We can't really use the name in this way for two reasons:
1. The project list is prepared on Node.js by our Vite plugin. The project files are not interpreted there, Vite only serves them to the browser. There's no way to access their names at that time.
2. The idea behind the selection screen is that it doesn't require loading all the projects in order to be displayed. This would be really slow for larger packages, like our examples.

After #166, we may be able to retrieve the name of the project from its metafile and display it instead, but the URL at which a given project is served will still depend on its filename.

To avoid the collision from your repro, we could use the entire path of the project as its URL, and not just the filename.

DanielBreiner:
Thank you for the clarification, makes a lot of sense. What you propose is reasonable and would certainly fix all the problems I opened this issue for. I will rename the issue so a linked PR may be created.

guifeliper:
I am working on this one!

aarthificial:
After #552 project names are displayed in the selection screen but there's still this part left to implement:
> To avoid the collision from your repro, we could use the entire path of the project as its URL, and not just the filename.

PR #552, opened 2023-03-18
feat(ui,vite-plugin): get name from meta file

guifeliper:
This PR is related to issue #297 

I have performed the test and all runs fine.


guifeliper:
@aarthificial I believe I would need to add some info in the docs for this implementation

packages/ui/src/Index.tsx
```diff
@@ -15,8 +16,10 @@ export function Index({projects}: IndexProps) {
       <div className={styles.header}>Projects</div>
       <div className={styles.list}>
         {projects.map(project => (
-          <a className={styles.element} href={`./${project.name}`}>
-            <div className={styles.title}>{project.name}</div>
+          <a className={styles.element} href={`./${project.fileName}`}>
+            <div className={styles.title}>
+              {project.name || project.fileName}
```
aarthificial, line 21: It's not possible for `name` to not be defined here

packages/vite-plugin/src/main.ts
```diff
@@ -111,7 +111,8 @@ export default ({
   const projectLookup: Record<string, ProjectData> = {};
   for (const url of typeof project === 'string' ? [project] : project) {
     const {name} = path.parse(url);
-    const data = {name, url};
+    const metaData = getMeta(`./src/${name}.meta`);
```
aarthificial, line 114: Projects can be located anywhere in the `src` directory.
Take a look at how the project loader creates the meta path:
https://github.com/motion-canvas/motion-canvas/blob/cdd3df1bff25b04b905e289264831d8d328caaab/packages/vite-plugin/src/main.ts#L206-L207

aarthificial (changes requested):

packages/ui/src/Index.tsx
guifeliper: Sorry, I don't understand. Why do you say it is not possible? 
Do You mean not technically, but for other reasons?
Because it is working like that:
image

packages/vite-plugin/src/main.ts
guifeliper: Ok, got it!

guifeliper: @aarthificial please have a look now, I have used the path library as the load function uses it.

aarthificial (approved):

PR #558, opened 2023-03-19
feat(core): add wait signal to signal context

### LINKED ISSUE #480: Method to hold an animation state for a specified time in compact form
**Description**
There are cases where we might want to do some animation, hold it for a certain amount of time, and then move to another.

I have found that it is quite a common usecase in my workflow.

For example, fade some text in, hold it for 2 seconds, and then fade it out again.

Right now, we have to do it like this
```js
yield* headerText().opacity(1, 1, easeInOutCubic);
yield* waitFor(2);
yield* headerText().opacity(0, 1, easeInOutCubic);

// or we can do it like this
// note how the .to method is the same value, and so it keeps the text opacity at 1 for an additional second
yield* headerText().opacity(1, 1, easeInOutCubic).to(1, 2).to(0, 1, easeInOutCubic);
```

This is very verbose, and it means that if you want to change max opacity from 1 to 0.8, then you would have to make
the change in two places.

**Proposed solution**
I think one solutions that would be nice is to have a `hold` method.
```js
yield* headerText().opacity(1, 1, easeInOutCubic).hold(2).to(0, 1, easeInOutCubic);
```

**Additional context**
I would like to make the contribution myself, but right now, I don't know where the change should be,
My best guess is in `signals/types`, in the `SignalGenerator`.

I would appreciate if you could point me in the correct direction.

samhattangady:
This change now allows user to use the `wait` signal while chaining commands.
For example, if you want the animation to grow the circle, stay that size, and then shrink again,

```typescript
yield* circle().scale(2, 2).wait(1).to(1, 2);
```

I tried to follow the implementation of `SignalTween` to the best of my understanding.
Please let me know if you think there is a better way to implement this.
Thanks.

Closes #480


aarthificial:
This works great but I think there's some room for improvement.
Right now there's a lot of code duplication in `makeStay`.
Perhaps we could simplify this by using a queue:

```ts
protected makeAnimate(
  defaultTimingFunction: TimingFunction,
  defaultInterpolationFunction: InterpolationFunction<TValue>,
) {
  // instead of passing `before` between generators, 
  // let's store all generators in a queue:
  const queue: ThreadGenerator[] = [];
  
  // make one generator that goes through the queue:
  const task = run(function* animate() {
    while (queue.length > 0) {
      yield* queue.shift()!;
    }
  }) as SignalGenerator<TSetterValue, TValue>;

  // `to` can now be implement by simply pushing a generator to the queue:
  task.to = (
    value: SignalValue<TSetterValue> | typeof DEFAULT,
    duration: number,
    timingFunction = defaultTimingFunction,
    interpolationFunction = defaultInterpolationFunction,
  ) => {
    queue.push(
      this.tween(value, duration, timingFunction, interpolationFunction),
    );
    return task;
  };

  // other methods such as `wait` could be implemented here in the same manner:
  task.wait = (duration: number) => {
    queue.push(waitFor(duration));
    return task;
  };

  // simply return `to`:
  return task.to;
}
```

aarthificial (approved):

ksassnowski:
Hey @samhattangady. Are you still interested in completing this PR? Did you have a look at @aarthificial's comments? I think this would be super useful to get into the library and it would be a shame for this PR to be abandoned.

If you don't plan to continue working on it, that's also fine. Someone else can take over in that case. Let us know what you think 👍

samhattangady:
Hey, I'm sorry but I don't think I will be able to continue working on this.

If anyone is interested, please feel free to take it up.

ksassnowski:
Alright, in that case I will take over. Thanks for your work so far!

aarthificial:
Closing this in favor of #651

Issue #289, opened 2023-02-10
Viewport Information in the UI


hhenrichsen:
**Description**
It'd be nice to display a couple pieces of info in the editor UI:
- [x] The current viewport center, and buttons to return it to default as well as copy the current value as a Vector2 (and their associated hotkeys)
- [x] The current mouse hover location (if it's over the viewport), and a button that displays the hotkey to copy that as a Vector2
- [x] The current zoom level, and buttons to zoom in / out / reset zoom (and their associated hotkeys) 
  #531
  #444
- [x] The current version of Motion Canvas 
  #501

**Proposed solution**
Perhaps these would be good to include in a bar at the top of the viewport?

aarthificial:
I think we could put it on a status bar.
As mentioned in https://github.com/motion-canvas/motion-canvas/issues/19#issuecomment-1419484450

I'm claiming this issue for myself, though. 
I'd like to keep absolute control over the UI until we have a proper design system/guide.

nrthbound:
Would definitely like to see, at the very least, the current zoom level so I know when I'm at 50%, 100%, or 150%. That would be quite handy for me in terms of knowing what my render could potentially look like.

Logon27:
> I think we could put it on a status bar. As mentioned in [#19 (comment)](https://github.com/motion-canvas/motion-canvas/issues/19#issuecomment-1419484450)
> 
> I'm claiming this issue for myself, though. I'd like to keep absolute control over the UI until we have a proper design system/guide.

@aarthificial It was discussed on discord that UI contributions are possible now. I would like to try taking on this feature if that is alright with you.

aarthificial:
Closed by #737

PR #569, opened 2023-03-21
feat(ui): added position text in ui

ShivamRawat0l:
Added a position indicator which updates on click. 

Attached Video:

https://user-images.githubusercontent.com/27425384/226644566-50390327-9e0f-41bd-b5f7-71d6f6c0ceab.mp4

Future Upgrade: A pointer pointing to the current position.

### DIFF
--- a/packages/ui/src/components/viewport/EditorPreview.tsx
+++ b/packages/ui/src/components/viewport/EditorPreview.tsx
@@ -1,4 +1,4 @@
-import {useCallback, useMemo, useRef} from 'preact/hooks';
+import {useCallback, useMemo, useRef, useState} from 'preact/hooks';
 
 import {
   useCurrentScene,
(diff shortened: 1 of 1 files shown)


ShivamRawat0l:
> Can you elaborate a bit on what this feature aims to achieve? A similar feature is requested in #289 but there the position updates immediately as the mouse moves and it's possible to copy it using a hotkey.

So, I wanted the location of the cursor to place elements on the canvas easily. With this we know the exact position. 

I can look into how we can copy those coordinates using a hotkey. Let me know which hotkey I should assign it to. 

My initial thought was updating the position with the cursor but there were 2 issues. 
1. There was some performance issue if I have a onHover on an element and constantly update on it. 
2. onHover does not capture the event on the canvas element.

aarthificial:
Closing this since the feature has been implemented in the context of #737

PR #575, opened 2023-03-24
docs: add spline documentation

ksassnowski:
This PR adds documentation for the new `Spline` node merged in #514. The section `A gentle introduction to splines` is still a work in progress, but the rest should be ready to be reviewed.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -33,7 +33,7 @@
         "leadingUnderscore": "forbid",
         "trailingUnderscore": "forbid",
         "filter": {
-          "regex": "^(__html|_jsx|_jsxs|_Fragment)$",
+          "regex": "^(__html)$",
           "match": false
         }
       },
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -33,7 +33,7 @@
         "leadingUnderscore": "forbid",
         "trailingUnderscore": "forbid",
         "filter": {
-          "regex": "^(__html|_jsx)$",
+          "regex": "^(__html|_jsx|_jsxs|_Fragment)$",
           "match": false
         }
       },
--- a/packages/docs/docs/components/spline.mdx
+++ b/packages/docs/docs/components/spline.mdx
@@ -1,5 +1,5 @@
 ---
-sidebar_position: 4
+sidebar_position: 3
 slug: /spline
 ---
 
@@ -34,8 +34,8 @@ export default makeScene2D(function* (view) {
 });
 ```
 
-The `Spline` component allows us to draw and animate smooth curves through a
-series of control points.
+The [`Spline`][spline] component allows us to draw and animate smooth curves
+through a series of control points.
 
 ## Defining control points
 
@@ -220,9 +220,8 @@ property.
 ```
 
 `auto` should be a value between `0` and `1` and represents the percentage of
-how much to blend between the user-provided handles and auto-calculated handles.
-A value of `1` means to use the auto handles while a value of `0` means using
-the user handles.
+how much to blend between the user-provided handles (`0`) and auto-calculated
+handles (`1`).
 
 :::tip
 
@@ -248,7 +247,7 @@ export default makeScene2D(function* (view) {
   const knots: Knot[] = [];
 
   view.add(
-    <Spline lineWidth={16} stroke={'lightseagreen'} closed>
+    <Spline lineWidth={16} stroke={'lightseagreen'} lineJoin={'round'} closed>
       <Knot
         ref={makeRef(knots, 0)}
         position={[-50, -80]}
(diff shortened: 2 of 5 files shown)


packages/docs/docs/components/spline.mdx
```diff
@@ -0,0 +1,501 @@
...
+`auto` should be a value between `0` and `1` and represents the percentage of
+how much to blend between the user-provided handles and auto-calculated handles.
+A value of `1` means to use the auto handles while a value of `0` means using
+the user handles.
+
+:::tip
+
+`auto` is a compound signal which means you can also specify `startHandleAuto`
+and `endHandleAuto` respectively to control the blend of each handle
+individually.
```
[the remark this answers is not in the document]
hhenrichsen: `auto` is a compound signal, which means you can specify `startHandleAuto` and `endHandleAuto` to individually control the blend of each handle.

packages/docs/docs/components/spline.mdx
```diff
@@ -0,0 +1,493 @@
...
+:::
+
+Since `auto` is a signal, it can also be animated.
+
+```tsx editor
+export default makeScene2D(function* (view) {
+  const knots: Knot[] = [];
+
+  view.add(
+    <Spline lineWidth={16} stroke={'lightseagreen'} closed>
```
aarthificial, line 251: Should also have `lineJoin={'round'}` to prevent the quick transition from a broken handle to auto

packages/docs/docs/components/spline.mdx
```diff
@@ -0,0 +1,493 @@
...
+
+### Animating the knots of a spline
+
+Since the `Knot` node is a regular Motion Canvas component, it can be animated
+in much the same way as other components.
+
+:::note
+
+Animating a spline's knots is only possible when using the
+[Knot](#using-knot-nodes) component, not when using the `points` property.
```
aarthificial, line 344:
```suggestion
[`Knot`](#using-knot-nodes) component, not when using the `points` property.
```

packages/docs/docs/components/spline.mdx
```diff
@@ -0,0 +1,493 @@
...
+    endHandle={[0, -20]}
+    auto={0.5}
+  />
+</Spline>
+```
+
+`auto` should be a value between `0` and `1` and represents the percentage of
+how much to blend between the user-provided handles and auto-calculated handles.
+A value of `1` means to use the auto handles while a value of `0` means using
+the user handles.
```
aarthificial, line 225: I think this could be simplified to:
```suggestion
`auto` should be a value between `0` and `1` and represents the percentage of
how much to blend between the user-provided handles (`0`) and auto-calculated handles (`1`).
```

packages/docs/docs/components/spline.mdx
```diff
@@ -0,0 +1,493 @@
...
+  yield* spline().start(1, 1.5).to(0.5, 1);
+  yield* spline().end(0.5, 1);
+  yield* all(spline().start(0, 1.5), spline().end(1, 1.5));
+});
+```
+
+### Animating the knots of a spline
+
+Since the `Knot` node is a regular Motion Canvas component, it can be animated
+in much the same way as other components.
```
aarthificial, line 339: Could be just:
```suggestion
`Knot`s can be animated in the same way as other nodes in Motion Canvas.
```

packages/docs/docs/components/spline.mdx
```diff
@@ -0,0 +1,493 @@
...
+
+Since `startHandle` and `endHandle` are relative to the knot, rotating the knot
+will also change the handle positions.
+
+:::caution
+
+Note that animating a knot's `rotation` is only possible if its `auto` value is
+less than `1`.
+
+:::
```
aarthificial, line 393: This information should be hoisted outside of the tabs since it applies to transform in general.
You could say:
```
You can think of `startHandle` and `endHandle` as being the children of the Knot -
changing the knot's position, rotation and scale will also transform the handles.
The only exceptions are `auto` handles which are unaffected by these transformations.
```

packages/docs/docs/components/spline.mdx
```diff
@@ -0,0 +1,493 @@
...
+
+  </TabItem>
+
+</Tabs>
+
+### Animating objects along a spline
+
+Splines are often used to animate objects along a path. The same is possible in
+Motion Canvas by using the `getPointAtPercentage` method of the `Spline`
+component.
```
aarthificial, line 449: Just a suggestion but I'd reword this as:
```suggestion
Splines can be useful to model the path that an object should follow.
You can use the `getPointAtPercentage` method to achieve this.
```
I think it's a bit clearer since Splines don't do the animating and the current sentence kinda reads like that: "Splines are often used to animate objects [...]"

.eslintrc.json
```diff
@@ -33,7 +33,7 @@
         "leadingUnderscore": "forbid",
         "trailingUnderscore": "forbid",
         "filter": {
-          "regex": "^(__html|_jsx)$",
+          "regex": "^(__html|_jsx|_jsxs|_Fragment)$",
```
aarthificial, line 36: At this point I'd just remove this whole thing, leave only `__html` and add this:
```ts
/* eslint-disable @typescript-eslint/naming-convention */
```
to `runtime.ts`

packages/docs/docs/components/spline.mdx
```diff
@@ -0,0 +1,493 @@
+---
+sidebar_position: 4
```
aarthificial, line 2: I think custom components should be the last thing on the menu

packages/docs/docs/components/spline.mdx
```diff
@@ -0,0 +1,493 @@
...
+      <Knot position={[120, -30]} startHandle={[0, -70]} />
+      <Knot position={[0, 100]} startHandle={[5, 0]} />
+    </Spline>,
+  );
+
+  yield* spline().scale(0.9, 0.6).to(1, 0.4);
+});
+```
+
+The `Spline` component allows us to draw and animate smooth curves through a
```
aarthificial, line 37: There should be a link to the corresponding API docs at least in the first mention of the node.

aarthificial (changes requested):

aarthificial (approved):

PR #579, opened 2023-03-27
feat(core): extend Vector2 type

ksassnowski:
This PR adds a few more convenience methods to the `Vector2` type.

- `equals` allows comparing vectors with an additional error threshold to account for floating point inaccuracies.
- `degrees` returns the angle between the vector and the positive x-axis in degrees. A `radians` getter already existed.
- `squaredMagnitude` returns the squared magnitude of the vector. This one is really just for completeness' sake but you gotta save those frames!

I also added the corresponding static methods for `radians`, `degrees`, and `squaredMagnitude` to work similarly to `magnitude`.

### DIFF
--- a/packages/core/src/types/Vector.ts
+++ b/packages/core/src/types/Vector.ts
@@ -117,8 +117,37 @@ export class Vector2 implements Type {
     return new Vector2(Math.cos(radians), Math.sin(radians));
   }
 
+  /**
+   * Return the angle in radians between the vector described by x and y and the
+   * positive x-axis.
+   *
+   * @param x - The x component of the vector.
+   * @param y - The y component of the vector.
+   */
+  public static radians(x: number, y: number) {
+    return Math.atan2(y, x);
+  }
+
+  /**
+   * Return the angle in degrees between the vector described by x and y and the
+   * positive x-axis.
+   *
+   * @param x - The x component of the vector.
+   * @param y - The y component of the vector.
+   *
+   * @remarks
+   * The returned angle will be between 0 and 360 degrees.
+   */
+  public static degrees(x: number, y: number) {
+    return (360 + (Vector2.radians(x, y) * 180) / Math.PI) % 360;
+  }
+
   public static magnitude(x: number, y: number) {
     return Math.sqrt(x * x + y * y);
+  }
+
+  public static squaredMagnitude(x: number, y: number) {
+    return x * x + y * y;
   }
 
   public get width(): number {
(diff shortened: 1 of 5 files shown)


packages/core/src/types/Vector.ts
```diff
@@ -270,7 +316,34 @@ export class Vector2 implements Type {
...
+   *
+   * @param other - The vector to compare.
+   * @param epsilon - The allowed error threshold when comparing the vectors.
+   */
+  public equals(other: Vector2, epsilon = 0.000001): boolean {
+    return (
+      Math.abs(this.x - other.x) <= epsilon + Number.EPSILON &&
+      Math.abs(this.y - other.y) <= epsilon + Number.EPSILON
+    );
+  }
```
aarthificial, line 348: Why is it implemented differently than `Matrix2D.equals()`?

packages/core/src/types/Vector.ts
```diff
@@ -117,10 +117,39 @@ export class Vector2 implements Type {
...
+
+  /**
+   * Return the angle in degrees between the vector described by x and y and the
+   * positive x-axis.
+   *
+   * @param x - The x component of the vector.
+   * @param y - The y component of the vector.
+   *
+   * @remarks
+   * The returned angle will be between 0 and 360 degrees.
```
aarthificial, line 139: Idk about this chief, `Vector2.radians()` returns the range `[-PI, PI]`
Shouldn't this return `[-180, 180]` too?

ksassnowski: I don't feel super strongly about this since this wouldn't make any difference when using it for rotating a node. I just found `[0, 360]` a bit more intuitive but I can change it if you'd prefer it to be consistent with `radians`

packages/core/src/types/Vector.ts
ksassnowski: If anything, `Matrix2D.equals()` should be updated to work this way as well. The reason I'm adding `Number.EPSILON` is to account for cases like this

```ts
const a = new Vector2(2.5, 2.5);
const b = new Vector2(2.4, 2.4);

a.equals(b, 0.1);
// => false
```

because in magical floating point land

```
2.5 - 2.4 = 0.10000000000000009
```

I can make the same change to `Matrix2D.equals()` in this PR as well. Or I can open a new PR with the change if you'd prefer that.

aarthificial: You can update it in this PR
Also make sure the arguments are called the same thing (`threshold` vs `epsilon`) and that the default value uses the same constant.

packages/core/src/types/Vector.ts
aarthificial: I'd go for consistency

packages/core/src/types/Vector.ts
ksassnowski: Done 👍

ksassnowski:
I implemented the requested changes and also made `Matrix2D` implement `Type` as we talked about.

PR #582, opened 2023-03-28
fix(2d): support full-width CJK characters

### LINKED ISSUE #476: Text overlap when using Chinese in codeblock
**Describe the bug**
Text overlap when using Chinese in codeblock

image

**To Reproduce**
Steps to reproduce the behavior
or a code snippet containing the scene that causes the problem:

```ts
  const codeRef = createRef<CodeBlock>();
  yield view.add(
    // note that the ` bracket is followed by a new line
    <CodeBlock
      ref={codeRef}
      language="js"
      code={`
        console.log('Hello World!')
        // more indented
        // less indented`}
      // selection={range(0, 12, 0, 26)}
    />
  );
  yield* waitFor(2);

  yield* codeRef().edit(2, false)`
  console.log('Hello ${edit("World", "kohai柯灰~~~")}!')
  // more indented1柯灰柯灰柯灰
  // less indented`;
  yield* codeRef().selection(lines(1), 1);
  yield* codeRef().selection(lines(2), 1);
  yield* codeRef().selection(lines(0, Infinity), 1);
  yield* waitFor(2);
});
```

**Expected behavior**
If possible, I would like to support other languages, thank you!

**Console errors**
N/A

- core: ^3.0.0
- ui: ^3.0.1

**Additional context**
N/A

liraymond04:
Closes #476


aarthificial:
Feels like most of this logic should be implemented in [code-fns](https://github.com/Ross-Esmond/code-fns) and not here.
CC: @Ross-Esmond

liraymond04:
Trying to get around the incorrectly spaced 'to' and 'from' results with offsets is admittedly very hack-y. Most of the logic could be removed if it was handled by diff directly.

liraymond04:
@aarthificial ^0.10.1 of code-fns now handles full-width characters for tokens, the draw fixes here now similarly use `wcwidth` to generally handle full-width characters.

Ross-Esmond:
Note that this would be a breaking change, since it removes the stock theme option. I brought back starry night.

We haven't figured out if we want that though, so this might get pushed back.

Ross-Esmond:
You don't have to install wcwidth here. Code-fns has it as a dependency.

aappaapp:
> Nevermind. 0.8 was the last starry night version.

Does this mean the PR can be merged now?

aarthificial:
I'm closing this PR because the next version includes a brand new [`Code`](https://github.com/orgs/motion-canvas/discussions/688) node that correctly supports CJK characters, emojis, and other non-mono-width characters.
It should be released before the end of the week.

PR #592, opened 2023-03-28
fixed typo in getting started guid

mabdullahadeel:


aarthificial:
`?scene` is a parameter for our Vite loader, it's not a typo.

PR #595, opened 2023-03-29
feat(renderer): renderer package

### LINKED ISSUE #415: Render projects headlessly
**Description**
It would be useful to be able to render a project from a shell script or in CI, without a human to operate a web browser. I'm not sure if this feature is already present; if it is, I can't find it in the documentation.

**Proposed solution**
A `render` script in package.json would be ideal, with a configuration file to set the parameters for the render. I don't know anything about the project's inner workings, so I don't think I can be more detailed than that.

**Considered alternatives**
Making a program that launches a headless browser to start the render is possible, but probably not ideal.

therealtimhawkins:
PR for the renderer package to headlessly render projects.

Thought I would make a PR for this to get some feedback and make sure I've not been barking up the wrong tree 😅

I thought it would be a better idea to make the changes to the **create** package in a different PR.

Also, not sure how package deployments work in lerna but happy to update after feedback 👍

Changes: 

1. The vite-plugin package is updated to load the renderer package when **/project/renderer** is visited.
2. The renderer package loads the renderer in puppeteer then renders to /output.

Closes #415


aarthificial (changes requested):
Since the `renderer` package provides its own editor, I think we can get away without modifying `vite-plugin`. Simply change `renderer.html` to `editor.html` and from `main.ts` export an `editor` function instead of `render`. Puppeteer can then simply visit `/projectName`.

packages/renderer/src/render.ts
```diff
@@ -0,0 +1,35 @@
...
+  if (!fs.existsSync(metaPath))
+    return console.log(`Project couldn't be found.`);
+
+  const [browser, server] = await Promise.all([
+    puppeteer.launch({headless: true}),
+    createServer({
+      root: Root,
+      configFile: path.resolve(Root, '../vite.config.ts'),
+      server: {
+        port: 9000,
```
gmile, line 23: Should this port number be configurable?

Not exactly sure if this is related, but I noticed that running `npm run serve` offers the following convenience: will start on port `9000`, and if it's not available, use the next available e.g. `9001`.

Perhaps the port number here could follow a similar approach?

aarthificial: Choosing the next available port is a Vite functionality. It will do it automatically.
Tho it may be a good idea to actually check what port Vite picked using:
```ts
server.httpServer.address().port
```
and redirect Puppeteer there.

Right now we're using a hardcoded `9000` which will work most of the time but will fail if something is already running on `9000`.

dcsan:
since this library is rendering to a canvas surface, is it not possible to just dump every frame to a file and then use something like ffmpeg to generate a video?

aarthificial:
@dcsan This is a draft for #415 which aims to give people the ability to
>  render a project from a shell script or in CI, without a human to operate a web browser

We already have an [ffmpeg exporter](https://motioncanvas.io/docs/rendering/video) (It even comes installed by default when bootstrapping a new project)
CLI rendering is not related to the exporter logic.

neilkinnish:
Hi, really like the look of motion canvas. Just wondering what stage this PR is at? is it usable at all or has main branch diverged off too much. Thanks

aarthificial:
@neilkinnish You can find a newer implementation in #1055
I'm closing this PR in favor of it.

Tomiborges (approved):

Tomiborges (approved):

PR #597, opened 2023-03-30
fix(2d): calculate arrow orientations for curves correctly

ksassnowski:
This PR does three things:

1. Add a `normal` property to `CurvePoint` that contains the normal of the point
2. Deprecate the `tangent` property of `CurvePoint` as it actually contains the normal of the point in a lot of cases.
3. Use the `normal` to calculate the actual tangent internally for now, e.g. when drawing arrow heads for lines and curves.

Tangents are kept as-is for now as changing them would be a breaking change. They will be fixed in the next major version.

This also fixes an inconsistent behavior in how arrow heads were drawn for linear and non-linear curves.

| Before | After |
| -------|------|
| CleanShot 2023-03-30 at 08 20 18@2x | CleanShot 2023-03-30 at 08 20 29@2x |

### DIFF
--- a/packages/2d/src/curves/CircleSegment.ts
+++ b/packages/2d/src/curves/CircleSegment.ts
@@ -43,19 +43,19 @@ export class CircleSegment extends Segment {
       );
     }
 
-    const startTangent = Vector2.fromRadians(startAngle);
-    const endTangent = Vector2.fromRadians(endAngle);
+    const startNormal = Vector2.fromRadians(startAngle);
+    const endNormal = Vector2.fromRadians(endAngle);
 
     return [
       {
-        position: this.center.add(startTangent.scale(this.radius)),
-        tangent: this.counter ? startTangent : startTangent.flipped,
-        normal: this.counter ? startTangent : startTangent.flipped,
+        position: this.center.add(startNormal.scale(this.radius)),
+        tangent: this.counter ? startNormal : startNormal.flipped,
+        normal: this.counter ? startNormal : startNormal.flipped,
       },
       {
-        position: this.center.add(endTangent.scale(this.radius)),
-        tangent: this.counter ? endTangent.flipped : endTangent,
-        normal: this.counter ? endTangent.flipped : endTangent,
+        position: this.center.add(endNormal.scale(this.radius)),
+        tangent: this.counter ? endNormal.flipped : endNormal,
+        normal: this.counter ? endNormal.flipped : endNormal,
       },
     ];
   }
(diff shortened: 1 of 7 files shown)


packages/2d/src/curves/CurvePoint.ts
```diff
@@ -2,5 +2,13 @@ import type {Vector2} from '@motion-canvas/core/lib/types';
 
 export interface CurvePoint {
   position: Vector2;
+  /**
+   * @deprecated The tangent is currently inconsistent for different types of
```
aarthificial, line 6: When used as a block tag, `@deprecated` should be formatted with a new line:
```suggestion
   * @deprecated 
   * The tangent is currently inconsistent for different types of
```

packages/2d/src/curves/CircleSegment.ts
```diff
@@ -47,10 +47,16 @@ export class CircleSegment extends Segment {
...
-      this.counter ? endTangent.flipped : endTangent,
+      {
+        position: this.center.add(startTangent.scale(this.radius)),
+        tangent: this.counter ? startTangent : startTangent.flipped,
+        normal: this.counter ? startTangent : startTangent.flipped,
+      },
+      {
+        position: this.center.add(endTangent.scale(this.radius)),
+        tangent: this.counter ? endTangent.flipped : endTangent,
+        normal: this.counter ? endTangent.flipped : endTangent,
```
aarthificial, line 58: I think we should also rename the variables to better reflect what they are,
```suggestion
        position: this.center.add(startTangent.scale(this.radius)),
        tangent: this.counter ? startNormal : startNormal.flipped,
        normal: this.counter ? startNormal : startNormal.flipped,
      },
      {
        position: this.center.add(endTangent.scale(this.radius)),
        tangent: this.counter ? endNormal.flipped : endNormal,
        normal: this.counter ? endNormal.flipped : endNormal,
```

aarthificial (changes requested):

packages/2d/src/curves/LineSegment.ts
```diff
@@ -24,18 +24,34 @@ export class LineSegment extends Segment {
...
       moveTo(context, from);
     }
     lineTo(context, to);
-    return [from, this.tangent.flipped, to, this.tangent];
+
+    return [
+      {
+        position: from,
+        tangent: this.tangent.flipped,
+        normal: this.tangent,
```
aarthificial, line 39: `this.tangent` should be renamed to `this.normal`

PR #603, opened 2023-04-01
Add Bézier curve components

ksassnowski:
This PR adds two new components: `CubicBezier` and `QuadBezier` for drawing cubic and quadratic Bézier curves, respectively.

These components don't really add any new math or logic since I was able to completely reuse the functionality added by the recent splines PR (#514).

### DIFF
--- a/packages/2d/src/components/CubicBezier.ts
+++ b/packages/2d/src/components/CubicBezier.ts
@@ -1,0 +1,109 @@
+import {SignalValue} from '@motion-canvas/core/lib/signals';
+import {PossibleVector2, Vector2Signal} from '@motion-canvas/core/lib/types';
+import {CurveProps} from './Curve';
+import {CubicBezierSegment} from '../curves';
+import {PolynomialSegment} from '../curves/PolynomialSegment';
+import {computed, vector2Signal} from '../decorators';
+import {bezierCurveTo, lineTo, moveTo} from '../utils';
+import {Bezier, BezierOverlayInfo} from './Bezier';
+
+export interface CubicBezierProps extends CurveProps {
+  p0: SignalValue<PossibleVector2>;
+  p0X: SignalValue<number>;
+  p0Y: SignalValue<number>;
+
+  p1: SignalValue<PossibleVector2>;
+  p1X: SignalValue<number>;
+  p1Y: SignalValue<number>;
+
+  p2: SignalValue<PossibleVector2>;
+  p2X: SignalValue<number>;
+  p2Y: SignalValue<number>;
+
+  p3: SignalValue<PossibleVector2>;
+  p3X: SignalValue<number>;
+  p3Y: SignalValue<number>;
+}
+
+/**
+ * A node for drawing a cubic Bézier curve.
+ *
+ * @example
+ * Defining a cubic Bézier curve using `points` property.
+ *
+ * ```tsx editor
+ * export default makeScene2D(function* (view) {
+ *   const bezier = createRef<CubicBezier>();
+ *
+ *   <CubicBezier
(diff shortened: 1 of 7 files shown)


packages/2d/src/components/QuadBezier.ts
```diff
@@ -0,0 +1,98 @@
...
+    super(props);
+  }
+
+  @computed()
+  protected segment(): PolynomialSegment {
+    return new QuadBezierSegment(this.p0(), this.p1(), this.p2());
+  }
+
+  @computed()
+  protected overlayInfo(matrix: DOMMatrix): BezierOverlayInfo {
```
aarthificial, line 78: Although it would be an interesting feature, right now computed methods don't get updated when called with a different set of arguments. This means that the value returned by `overlayInfo` will not be updated when the `matrix` change.

Repro: select the node and zoom out the preview:
image

packages/docs/docs/components/bezier.mdx
```diff
@@ -0,0 +1,295 @@
...
+## Using the components
+
+To draw Bézier curves, we can use the [`QuadBezier`][quad-bezier] component for
+quadratic Bézier curves and [`CubicBezier`][cubic-bezier] component for cubic
+Bézier curves, respectively.
+
+:::info
+
+Most of the examples on this page can be switched between cubic and quadratic
+Bézier curves by using the dropdown at the right side of the animation player.
```
aarthificial, line 80:
```suggestion
Bézier curves by using the dropdown on the right side of the animation player.
```

packages/docs/docs/components/bezier.mdx
```diff
@@ -0,0 +1,295 @@
...
+
+:::
```
aarthificial, line 82: I think this could be shortened to:
```suggestion
Each example below is applicable to both the [`QuadBezier`][quad-bezier] and
[`CubicBezier`][cubic-bezier] nodes. You can switch between the two types of
curves using the dropdown on the right side of the animation player.
```

packages/docs/docs/components/bezier.mdx
```diff
@@ -0,0 +1,295 @@
...
+
+### Defining control points
+
+Bézier curves are defined by a start and end point, as well as several control
+points. The exact number of control points is different for different kinds of
+Bézier curves. A quadratic Bézier curve only has a single control point whereas
```
aarthificial, line 88:
```suggestion
Bézier curves. A quadratic Bézier curve has only a single control point whereas
```

packages/docs/docs/components/bezier.mdx
```diff
@@ -0,0 +1,295 @@
...
+a cubic Bézier curve has two.
+
+To define a Bézier curve, we need to define its start and point, as well as its
+control points.
```
aarthificial, line 92: This is not necessary
```suggestion
```
You already said that in the previous paragraph:
```
Bézier curves are defined by a start and end point, as well as several control
points.
```

aarthificial (changes requested):

packages/2d/src/components/CubicBezier.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  p3: SignalValue<PossibleVector2>;
+  p3X: SignalValue<number>;
+  p3Y: SignalValue<number>;
+}
+
+/**
+ * A node for drawing a cubic Bézier curve.
+ *
+ * @example
```
aarthificial, line 31: The main example should use a `@preview` tag:
```suggestion
 * @preview
```
Unlike `@example`, it will be always expanded at the top of the documentation.

aarthificial (approved):

PR #613, opened 2023-04-07
docs: document setting layout=false on a node

### LINKED ISSUE #449: Document `layout={false}`
**Description**
Modify the [*Layouts* guide](https://motioncanvas.io/docs/layouts) to include a section about opting out of the layout using `layout={false}`.

gmile:
Document change of node's properties after setting its `layout` to `false`.

Fixes #449

### DIFF
--- a/packages/docs/docs/getting-started/layouts.mdx
+++ b/packages/docs/docs/getting-started/layouts.mdx
@@ -107,6 +107,65 @@ useful properties are listed below:
 <ApiSnippet url={'/api/2d/components/Layout#justifyContent'} />
 <hr />
 
+### Disabling `layout` on a child node
+
+Sometimes it can be useful to remove the node from the node tree. This way the
+node can be styled independently of its ancestor layout, as well as moved freely
+on the canvas.
+
+Removal of node from the node tree is done by setting a value of `layout`
+property to `false`. The node can be brought back to its original position
+inside the node tree by setting the value of `layout` property to `true`.
+
+After disabling `layout` on a node, the following things change:
+
+1. the node is moved to the place of origin, e.g. center of the scene,
+2. the node can now be position on canvas freely,
+3. the node no longer inherits properties from its parent, such as text styling,
+4. the node's former parent layout will no longer considers the node's width or
+   height when calculating its own bounding box. Disabling a layout on a node
+   can be roughly thought of as an equivalent to setting a `position` CSS
+   property of an HTML node to `absolute` value.
+
+:::info
+
+The node retains its rendering order (`z-index`, in terms of CSS), e.g. it will
+potentially overlap any other nodes that were rendered previously, as well as
+get potentially covered by nodes that appear later.
+
+:::
+
+Consider the following example:
+
+1. a top-level `<Rect>` node has a `layout` property set, meaning that all its
+   children nodes follow `<Rect>`'s styling and direction,
+2. the yellow `<Circle>` is a child of `<Rect>`,
+3. commenting out line `yellowCircle().layout = false;` will make two green
+   circles get close to each other, as if the yellow circle is not there; the
+   yellow circle now stands alone now, can be moved freely and has it's own text
+   styling.
+
+```tsx editor
+import {makeScene2D} from '@motion-canvas/2d/lib/scenes';
+import {Circle, Rect} from '@motion-canvas/2d/lib/components';
+import {createRef} from '@motion-canvas/core/lib/utils';
+
+export default makeScene2D(function* (view) {
+  const yellowCircle = createRef<Circle>();
+
+  view.add(
+    <Rect layout alignItems="center" direction="row">
+      <Circle width={100} height={100} fill="forestgreen" />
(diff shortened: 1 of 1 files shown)


packages/docs/docs/getting-started/layouts.mdx
```diff
@@ -107,6 +107,65 @@ useful properties are listed below:
 <ApiSnippet url={'/api/2d/components/Layout#justifyContent'} />
 <hr />
 
+### Disabling `layout` on a child node
+
+Sometimes it can be useful to remove the node from the node tree. This way the
```
aarthificial, line 112: > remove the node from the node tree

That's not what disabling `layout` does. The node will still be the child of its parent, even when `layout` is disabled.

packages/docs/docs/getting-started/layouts.mdx
```diff
@@ -107,6 +107,65 @@ useful properties are listed below:
...
+node can be styled independently of its ancestor layout, as well as moved freely
```
aarthificial, line 113: > node can be styled independently of its ancestor layout

A node can always be styled independently. The goal of disabling `layout` is to regain control over the node's position and size. Or, in other words, to exclude it from the parent's layout.

packages/docs/docs/getting-started/layouts.mdx
```diff
@@ -107,6 +107,65 @@ useful properties are listed below:
...
+on the canvas.
+
+Removal of node from the node tree is done by setting a value of `layout`
+property to `false`. The node can be brought back to its original position
+inside the node tree by setting the value of `layout` property to `true`.
+
+After disabling `layout` on a node, the following things change:
+
+1. the node is moved to the place of origin, e.g. center of the scene,
```
aarthificial, line 122: That's not true. The node's position will be whatever the `position` property is set to. Enabling the layout does not change this property - it overrides it but the values are still kept in there.

packages/docs/docs/getting-started/layouts.mdx
```diff
@@ -107,6 +107,65 @@ useful properties are listed below:
...
+2. the node can now be position on canvas freely,
+3. the node no longer inherits properties from its parent, such as text styling,
```
aarthificial, line 124: text styling are the only properties inherited by a node. "such as" implies that there are more which is not true.

packages/docs/docs/getting-started/layouts.mdx
```diff
@@ -107,6 +107,65 @@ useful properties are listed below:
...
+4. the node's former parent layout will no longer considers the node's width or
+   height when calculating its own bounding box. Disabling a layout on a node
+   can be roughly thought of as an equivalent to setting a `position` CSS
+   property of an HTML node to `absolute` value.
```
aarthificial, line 128: I don't think mentioning `absolute` positioning from CSS here makes this clearer.
Our default positioning system is already described in the previous section.

packages/docs/docs/getting-started/layouts.mdx
```diff
@@ -107,6 +107,65 @@ useful properties are listed below:
...
+property to `false`. The node can be brought back to its original position
+inside the node tree by setting the value of `layout` property to `true`.
+
+After disabling `layout` on a node, the following things change:
+
+1. the node is moved to the place of origin, e.g. center of the scene,
+2. the node can now be position on canvas freely,
+3. the node no longer inherits properties from its parent, such as text styling,
+4. the node's former parent layout will no longer considers the node's width or
+   height when calculating its own bounding box. Disabling a layout on a node
```
aarthificial, line 126: The layout system calculates the sizes and positions of nodes, not their bounding boxes. 
"Bounding box" has a specific meaning in Motion Canvas which is different from the size of a node.

packages/docs/docs/getting-started/layouts.mdx
```diff
@@ -107,6 +107,65 @@ useful properties are listed below:
...
+The node retains its rendering order (`z-index`, in terms of CSS), e.g. it will
+potentially overlap any other nodes that were rendered previously, as well as
+get potentially covered by nodes that appear later.
+
+:::
+
+Consider the following example:
+
+1. a top-level `<Rect>` node has a `layout` property set, meaning that all its
+   children nodes follow `<Rect>`'s styling and direction,
```
aarthificial, line 141: I don't like the emphasis on "styling and direction" here.
The point of the `layout` property is to arrange the children into a layout by changing their sizes and postions.
Inheriting the text style is a secondary thing that's not yet described in other sections.
Additionally, this example doesn't even use text.

aarthificial (changes requested):

ksassnowski:
@gmile It's been a while since the last review. Do you still plan on continuing with this PR? Have you had time to address @aarthificial's comments yet?

gmile:
@ksassnowski I haven't had a chance to go over them, unfortunately. I intent to complete this, but I am also open to closing the issue for the time being. Closing for now to keep the tracker tidy 🙇

ksassnowski:
Alright, thanks! Feel free to reopen the PR if you ever get around to working on it again 👍

PR #618, opened 2023-04-09
docs: convert <AnimationPlayer name="layout" />

gmile:
This commit upgrades <AnimationPlayer name="layout" /> to a runnable
example

For discussion, see:
https://discord.com/channels/1071029581009657896/1076166997513093272/1094241272098725959

There appears to be 16 more occurences of player left:

  $ rg --count-matches --fixed-strings '<AnimationPlayer' | awk -F ':' '{sum+=$2;}END{print sum;}'
  16


gmile:
I've converted a few more banner AnimationPlayer-s into fiddles, intend to look into implementing hiding of the fiddle UI / editor for them.

aarthificial:
I converted this PR to a draft instead of labeling it as a "work in progress"
As I've mentioned on Discord, I think the banners should still be implemented using the player package.
All of the other examples however should be converted into fiddles.

gmile:
ℹ️ this needs a rebase now. Here's what needs to be done for this PR to be considered complete:

1. reset all changes on the fiddle,
2. bring back all banners to be video players again - e.g. reset all changes done to top-level banners in this PR,
3. convert all other examples across documentation to fiddles.

gmile:
@aarthificial I've reset all changes previously made to `<AnimationEditor banner...>` and updated all examples that aren't banners. There's a bit more work necessary here, so before I do it I appreciate your review of the existing conversions first.

Additionally, I'd love your input on some issues that I've encountered:

1. all converted examples use a `small` attribute. Now that all these examples are gone, would you say it's necessary to keep supporting the `small` attribute in `AnimationEditor` or could it be cleaned up?
2. Additionally, some animations are huge in size now, see example below. However, modifying the absolute values of the shapes of the animation makes it fit, but it also becomes blurry. I'm not sure if blurriness is OK. Wtdy?

    <details>
    <summary>Preserving size example</summary>
    image
    </details>

    <details>
    <summary>Blurriness example</summary>
    image
    image
    </details>

3. some more advanced code examples, for instance "tweening visualizer", are quite long:

    <details>
    <summary>Long code example</summary>
    image
    </details>

    It looks a bit impractical to have such a long example. What do you think about it?  wonder if adding something like a "show/hide source" button in the editor toolbar wouldn't help here, for example (not necessarily in this PR).

gmile:
Following up on my previous comment, another issue worth pointing our is that highlighting of individual rows is gone. One example is this https://motioncanvas.io/docs/signals#complex-example: after converting it to AnimationEditor, the lines will no longer be highlighted.

gmile:
@aarthificial some tiny progress update on this:

> Yes, I believe we can get rid of the small attribute

I've cleaned up the `small` attribute.

> This should be solved by adding an additional property for Fiddles that controls the ratio of the preview. That specific example could then use 16/9 as its ratio.

After enabling `16/9` value as aspect ratio (thanks for adding it, btw!), here's how the example looks like - there's a bit more room now, but the circle is still too big. Probably it'd make sense to adjust the example values now, wtdy? I intent to play with the absolute values in the example a bit, but feel free to express your vision for how this should be handled ideally.

<details>
<summary>Screenshot 1</summary>
image
</details>

There's something about proportions in other examples too, see another screenshot below (though maybe even this happens to all example - I still intend to check all of them). So I think they will need adjustments too.

<details>
<summary>Screenshot 1</summary>
image
</details>

> specific tweening visualizer example could be replaced with a much simpler one

Make sense, I'll work on a new one.

> a select box that lets us choose the timing function

The description (e.g. "_The effects of a particular easing function can be visualised..._") to the example in question, as well as the example itself demonstrates only the `easeInOutBounce` function.

Now that you mention a select box to choose the timing function (via `snippet`, like some other examples already do), I have to ask: are you saying you would like more functions to be demonstrated within this particular instance of AnimationPlayer? If so, do you have a specific timing functions in mind you want to see in the select box, or all of them?

By all I am thinking of specifically this list (got it by looking at functions that return a value of `TimingFunction` type):

* [easeInBack](https://motioncanvas.io/api/core/tweening/#easeInBack)
* [easeInBounce](https://motioncanvas.io/api/core/tweening/#easeInBounce)
* [easeInElastic](https://motioncanvas.io/api/core/tweening/#easeInElastic)
* [easeInOutBack](https://motioncanvas.io/api/core/tweening/#easeInOutBack)
* [easeInOutBounce](https://motioncanvas.io/api/core/tweening/#easeInOutBounce)
* [easeInOutElastic](https://motioncanvas.io/api/core/tweening/#easeInOutElastic)
* [easeOutBack](https://motioncanvas.io/api/core/tweening/#easeOutBack)
* [easeOutBounce](https://motioncanvas.io/api/core/tweening/#easeOutBounce)
* [easeOutElastic](https://motioncanvas.io/api/core/tweening/#easeOutElastic)

Maybe I am jumping too far ahead, but in case we want to have _all_ functions added to given AnimationPlayer instance of Tweening page, then there's another question/suggestion:
1. on https://motioncanvas.io/docs/tweening/, for example in question (e.g. for easeInOutQuad), update the text description + add a select with 1-2 more functions,
2. on https://motioncanvas.io/api/core/tweening/, add an animation example (via AnimationPlayer) to each timing function.

I think doing this could help improve the API reference rich with examples of usage of individual functions, while the "Docs" section could focus more on the aspects of discovery and education of basic principles. Wdyt?

aarthificial:
> After enabling 16/9 value as aspect ratio [...] the circle is still too big

The examples were made using a 1920x1080 canvas.
The fiddle uses a fixed width of 960 with the height controlled by ratio.
So with a 16/9 ratio, it would be 960x540. This means that all lengths in the examples should be divided by two to make it look the same.

>are you saying you would like more functions to be demonstrated within this particular instance of AnimationPlayer

I'm assuming that `AnimationPlayer`s on this page are being replaced by `Fiddle`s. The snippets for timing functions could then be added to the fiddle above:

<details><summary>Photo</summary>

image

</details>

> If so, do you have a specific timing functions in mind you want to see in the select box, or all of them?

I think the list you've mentioned is ok.

> Maybe I am jumping too far ahead, but in case we want to have all functions added to given AnimationPlayer instance of Tweening page, then there's another question/suggestion:

I think we should start by including them all in the handwritten docs. We don't include fiddles in the `@example` section of the API docs at the moment because they require the whole boilerplate with `export default makeScene2D` which IMO makes them less readable in an IDE.

In the future, I think it would be ideal to add some sort of `template` prop to Fiddle that would include the boilerplate automatically, for example, something like this:
````md
```ts editor template=circle
yield* circle().scale(1, 1);
```
````
could result in:
```ts
  const circle= createRef<Circle>();
  view.add(<Circle ref={circle} size={140} fill="#e13238" />);
  
  yield* circle().scale(1, 1);  
});
```
But this may be out of the scope of this PR.

PR #620, opened 2023-04-10
chore(ui): fix typo in ID of the element

gmile:
Just a small typo in the code. Quickly checked, it didn't looked like this element is being referenced anywhere though 🤔


aarthificial:
Thanks for the PR!
A note for the future: the `chore` scope has been removed from the Angular Commit Conventions in favor of `ci` and `build`.
Though I'd say this is neither of those - it's just a `fix`.

PR #624, opened 2023-04-27
feat(core): add `loopUntil` function for V3

### LINKED ISSUE #362: Add "loopUntil" flow function

squigglesdev:
**Summary**
This PR closes #362 by implementing a `loopUntil()` function. It works in a way similar to `loop()` but uses a `TimeEvent` rather than `iterations` to control the loop. It is now fully updated to work with Motion Canvas version 3 or above.

**Functionality**
This functionality can be achieved by threads, however these are currently undocumented and the approach of adding a new `flow` function would make the process more intuitive for beginners. I followed the naming conventions, and would have renamed `loop()` to `loopFor()` to follow the names of `waitFor()` and `waitUntil()` however this would introduce breaking changes to projects currently using `loop()`.

**Example usage**
```tsx

view.add(
);

yield* loopUntil(  // Perform `circle().position.x(-10,0.1).to(10,0.1)` repeatedly until 'endVibrate' event
	'endVibrate', () =>
	circle().position.x(-10,0.1).to(10,0.1),
);
```

### DIFF
--- a/packages/core/src/flow/loopUntil.ts
+++ b/packages/core/src/flow/loopUntil.ts
@@ -1,0 +1,40 @@
+import {useDuration, usePlayback, useThread} from '../utils';
+import {decorate, threadable} from '../decorators';
+import {ThreadGenerator} from '../threading';
+import {LoopCallback} from './loop';
+
+decorate(loopUntil, threadable());
+/**
+ * Run the given generator until an event occurs.
+ *
+ * @example
+ * ```ts
+ * yield* loopUntil(
+ *   'Stop Looping',
+ *   () => circle().position.x(-10, 0.1).to(10, 0.1)
+ * );
+ * ```
+ *
+ * @param event - The event.
+ * @param tasks - A list of tasks to run.
+ */
+export function* loopUntil(
+  event: string,
+  factory: LoopCallback,
+): ThreadGenerator {
+  const thread = useThread();
+  const step = usePlayback().framesToSeconds(1);
+  const targetTime = thread.time() + useDuration(event);
+
+  let iteration = 0;
+  while (targetTime - step > thread.fixed) {
+    const generator = factory(iteration);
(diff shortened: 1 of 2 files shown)


ksassnowski:
Looks like this PR is missing the export inside `flow/index.ts` of the old PR.

packages/core/src/flow/loopUntil.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @example
+ * ```ts
+ * yield* loopUntil(
+ *   'Stop Looping',
+ *   () => circle().position.x(-10, 0.1).to(10, 0.1)
+ * );
+ * ```
+ *
+ * @param event - The event.
+ * @param tasks - A list of tasks to run.
```
ksassnowski, line 19: This parameter description is incorrect. It parameter should be called factory instead of tasks. It's also not a list of tasks to run but a function creating the generator to run. This was probably incorrectly copy&pasted from the doc block of all.

I'd suggest using the description of factory from the loop function.

squigglesdev: Ok, yeah sorry I will fix that.

PR #626, opened 2023-04-27
feat(2d): add ref method on node class

ccaven:
Trivial change.

Adds a method, `ref() => Reference<this>` on the `Node` class.

Use case:
```tsx
return this.nodeStack.shift().ref(); // this.nodeStack is a Node[]
```

### DIFF
--- a/packages/2d/src/components/Node.ts
+++ b/packages/2d/src/components/Node.ts
@@ -25,6 +25,8 @@ import {
   DetailedError,
   ReferenceReceiver,
   useLogger,
+  createRef,
+  Reference,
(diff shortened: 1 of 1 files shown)


ccaven:
Oops, ignore this PR for now - fixing the build error 🤔

ccaven:
Maybe it's worth adding a `ReadOnlyReference`  type?

ksassnowski:
Hey @ccaven. Please open an issue first before opening a PR that changes the public API so we can discuss if this is something we'd like to add to the library.

PR #627, opened 2023-04-28
feat(2d): add arcLength helper methods to Curve

aarthificial:


packages/2d/src/components/Curve.ts
```diff
@@ -70,12 +167,66 @@ export abstract class Curve extends Shape {
...
+   *
+   * @param value - The percentage along the curve.
+   */
   public percentageToDistance(value: number): number {
-    const {arcLength} = this.profile();
+    return this.startOffset() + this.offsetArcLength() * value;
+  }
+
+  /**
+   * Covert a distance along the curve to a percentage.
```
ksassnowski, line 184: `Covert` -> `Convert`

ksassnowski (changes requested):

ksassnowski (approved):

PR #629, opened 2023-04-28
feat(2d): custom arrow heads

### LINKED ISSUE #380: Customizable shape for the Line Arrowheads

ksassnowski:
This is a continuation of #389.

Closes #380 

@strau0106 I have added you as a co-author to this PR since you did most of the work on this.

### DIFF
--- a/packages/2d/src/components/Curve.ts
+++ b/packages/2d/src/components/Curve.ts
@@ -30,7 +30,7 @@ export interface CurveProps extends ShapeProps {
   /**
    * {@inheritDoc Curve.startArrow}
    */
-  startArrow?: SignalValue<boolean>;
+  startArrow?: SignalValue<Node | boolean>;
   /**
    * {@inheritDoc Curve.end}
    */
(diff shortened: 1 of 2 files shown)


strau0106:
great thanks!

strau0106:
What's up with this?

ksassnowski:
We decided to change how the arrow head node gets positioned. Previously, it would put the node's center (its `position`) at the tip of the arrow. Now the arrow head gets positioned based on it's `offset`. This leads to a few degenerate cases when having to shrink the arrow head when the line's length decreases. I haven't had the time to figure out a solution to this, unfortunately.

strau0106:
That makes sense and I see where the problem lies. I‘ll maybe also have a look and try to understand the problem

PR #646, opened 2023-05-05
docs(fix): update example

gmile:
The example description doesn't seem to mention anything about circle, also the variable name suggests it's a rectangle


aarthificial:
Thanks for the fix!
Side note: changes in the content of the documentation (things in the `packages/docs/docs` directory) should always use the `docs:` scope. Things like `fix(docs)`, `feat(docs)`, etc. are meant for changes in the functionality of the website (things in the `packages/docs/src` directory).
This is for the purposes of changelog generation.

PR #651, opened 2023-05-07
feat(core): improve `SignalGenerator` chaining

### LINKED ISSUE #480: Method to hold an animation state for a specified time in compact form

aarthificial:
`SignalGenerator` now has four additional methods for complex chaining:
```ts
yield* circle()
  // initiate the generator
  .scale(0.5, 2)
  // tween to another value
  .to(2, 2)
  // execute a callback
  .do(() => circle().fill('red'))
  // wait for one second
  .wait(1)
  // run the given generator
  .run(circle().position.y(100, 2))
  // tween back to the initial value
  .back(2);
```

Closes #480

### DIFF
--- a/packages/core/src/signals/SignalContext.ts
+++ b/packages/core/src/signals/SignalContext.ts
@@ -6,7 +6,7 @@ import {
 } from '../tweening';
 import {errorToLog, useLogger} from '../utils';
 import {ThreadGenerator} from '../threading';
-import {run} from '../flow';
+import {run, waitFor} from '../flow';
 import {DependencyContext} from './DependencyContext';
 import {
   SignalExtensions,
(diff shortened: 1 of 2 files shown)


samhattangady:
This is amazing!

I'm sorry I ditched on my branch (#558)... but I felt that the alternative you had suggested was _way_ beyond my suggestion, and blindly taking your input would just lead to more confusion as opposed to a better product overall.

I hope to someday make an impactful contribution to your project =). But for now, just happy to see more functionality. =D

PR #673, opened 2023-05-11
feat(create): add exporter selection

aarthificial:


tomByrer:
Thanks so much @aarthificial; I was planning on doing something like this!

PR #684, opened 2023-05-13
fix(docs) nonexistent reference in media documentation

piano1029:
In the Video example without a reference in section about media in the documentation there is a `ref` in the JSX element, but the variable it binds to does not exist. This pull-request fixes that.


aarthificial:
Thanks for the fix!

PR #686, opened 2023-05-14
fix(docs): add second parameter `''` in `useScene().variables.get(...)`

rtkid-nt:


packages/docs/docs/advanced/project-variables.mdx
```diff
@@ -34,5 +34,5 @@ Accessing the variables inside of a scene is through
 [`useScene()`](/api/core/utils#useScene):
 
 ```ts
-const circleFill = useScene().variables.get('circleFill');
+const circleFill = useScene().variables.get('circleFill', '');
```
aarthificial, line 37: Should be some valid color like `'blue'` since an empty string will throw if you try to use it as a fill.

aarthificial (changes requested):

aarthificial (approved):

PR #687, opened 2023-05-17
docs: add looping section to flow

### LINKED ISSUE #483: Documentation on Looping / Animating Multiple Objects
**Description**
Currently we have [flow](https://motioncanvas.io/docs/flow#all) that somewhat handles this case, but I think it would be useful to categorize some of the ways that you can loop multiple things at once, like these:

Functional:
```tsx
yield* all(Object.values(harpsRefs).map((circleRef) => {
    return circleRef().position.y(100, 2);
}));
```

For loops:
```tsx
for (const ref of Object.values(harpsRefs)) {
    yield ref().position.y(100, 2);
}
yield* waitFor(2);
```

Array-based for loops:
```tsx
const generators = [];
for (const ref of Object.values(harpsRefs)) {
    generators.push(ref.position.y(100, 2));
}
yield* all(...generators);
```

and whatever this cursed snippet is:
```tsx
const generators = [];
for (const ref of Object.values(harpsRefs)) {
    generators.push(ref.position.y(100, 2));
}
yield* generators;
yield* join(...generators);
```

just to give people an overview of how to solve some of these problems with generators (and give me a place to link when people ask this on the discord because it's happened twice now)

hhenrichsen:
Closes #483.

### DIFF
--- a/packages/docs/docs/getting-started/flow.mdx
+++ b/packages/docs/docs/getting-started/flow.mdx
@@ -131,3 +131,87 @@ generator in the quickstart section, there's a few more:
 
 <ApiSnippet url={'/api/core/flow#loop'} />
 <hr />
+
+## Looping
+
+There are many ways to animate multiple objects. Here are some examples. Try
+using them in the below editor.
+
+```tsx editor ratio=2
+import {makeScene2D} from '@motion-canvas/2d';
+import {Rect} from '@motion-canvas/2d/lib/components';
+import {all, waitFor} from '@motion-canvas/core/lib/flow';
+import {makeRef, range} from '@motion-canvas/core/lib/utils';
+
+export default makeScene2D(function* (view) {
+  const rects: Rect[] = [];
+
+  // Create some rects
+  view.add(
+    range(5).map(i => (
+      <Rect
+        ref={makeRef(rects, i)}
+        width={100}
+        height={100}
+        x={-250 + 125 * i}
+        fill="#88C0D0"
+        radius={10}
+      />
+    )),
+  );
+
+  yield* waitFor(1);
+
+  // Animate them
+  yield* all(
+    ...rects.map(rect => rect.position.y(100, 1).to(-100, 2).to(0, 1)),
+  );
+});
+```
+
+### Using `Array.map` and `all`
+
+This is one of the most elegant ways to do simple tweens, but requires nesting
+`all` to do multiple tweens on an object since the `map` callback must return a
+`ThreadGenerator`.
+
+```tsx
+yield *
+  all(
+    ...rects.map(rect =>
+      // No yield or anything; we return this generator and deal with it outside
+      rect.position.y(100, 1).to(-100, 2).to(0, 1),
+    ),
+  );
+```
+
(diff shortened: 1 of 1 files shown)


packages/docs/docs/getting-started/flow.mdx
```diff
@@ -98,36 +98,128 @@ function. Generators like this are called tweens, because they animate
...
 
 <ApiSnippet url={'/api/core/flow#loop'} />
 <hr />
+
+## Looping
+
+There are many ways to animate multiple objects. Here are some examples. Try
+using them in the below editor.
+
+```tsx editor
```
aarthificial, line 142: You can use `ratio` to make the preview a bit higher so that the squares don't overflow it:
```suggestion
```tsx editor ratio=2
```

packages/docs/docs/getting-started/flow.mdx
```diff
@@ -98,36 +98,128 @@ function. Generators like this are called tweens, because they animate
...
+      <Rect
+        ref={makeRef(rects, i)}
+        width={100}
+        height={100}
+        x={-250 + 125 * i}
+        fill="#88C0D0"
+        radius={10}
+      />,
+    );
+  }
```
aarthificial, line 163: I think we should promote the use of `range` over manual loops:
```suggestion
  view.add(
    range(5).map(i => (
      <Rect
        ref={makeRef(rects, i)}
        width={100}
        height={100}
        x={-250 + 125 * i}
        fill="#88C0D0"
        radius={10}
      />
    )),
  );
```

packages/docs/docs/getting-started/flow.mdx
```diff
@@ -98,36 +98,128 @@ function. Generators like this are called tweens, because they animate
...
 <ApiSnippet url={'/api/core/flow#loop'} />
 <hr />
+
+## Looping
+
+There are many ways to animate multiple objects. Here are some examples. Try
+using them in the below editor.
+
+```tsx editor
+import {ThreadGenerator} from '@motion-canvas/core/lib/threading';
```
aarthificial, line 143: `ThreadGenerator` import is not necessary and the import for `makeScene2D` is missing
```suggestion
import {makeScene2D} from '@motion-canvas/2d';
```
Sadly, fiddles don't verify the imports.

packages/docs/docs/getting-started/flow.mdx
```diff
@@ -98,36 +98,128 @@ function. Generators like this are called tweens, because they animate
...
+yield *
+  all(
+    ...rects.map(rect =>
+      // No yield or anything; we return this generator and deal with it outside
+      rect.position.y(100, 1).to(-100, 2).to(0, 1),
+    ),
+  );
+```
+
+<hr />
```
aarthificial, line 190: I don't think the `<hr />`s here are necessary.
I use them when including `ApiSnippet`s because they are visually busy and blend together otherwise. 
But here it looks good without them.

aarthificial (changes requested):

aarthificial (approved):

Issue #689, opened 2023-05-17
Bug: project `variables` doesn't work in present and render mode


rtkid-nt:
I have project variables:
```ts
import { makeProject } from "@motion-canvas/core";

import example from "./scenes/example?scene";

export default makeProject({
  scenes: [example],
  variables: {
    mainFont: "Jetbrains Mono",
  },
});

```

Use in scene
```ts
import { makeScene2D } from "@motion-canvas/2d/lib/scenes";
import { CodeBlock } from "@motion-canvas/2d/lib/components/CodeBlock";
import { useScene } from "@motion-canvas/core/lib/utils";

export default makeScene2D(function* (view) {
  const mainFont = useScene().variables.get("mainFont", "");

  yield view.add(
    <CodeBlock fontFamily={mainFont} language="cpp" code={`int main() {}`} />
  );
});

```

In editor all right 

![editor](https://github.com/motion-canvas/motion-canvas/assets/121748076/c03c82f5-57d6-4ff7-ae25-0e3eed6e9f97)

But in present and render mode, scene doesn't see them

![present](https://github.com/motion-canvas/motion-canvas/assets/121748076/2d43e04a-60fe-44be-a8cc-cd2dbb560ac3)

log
```ts
console.log(useScene().variables)
```

Log in console:
1. ```variables: { mainFont: "Jetbrains Mono" }``` - editor
2. ```variables: { mainFont: "Jetbrains Mono" }``` - editor
3. ```variables: {  }``` - present
4. ```variables: {  }``` - present
5. ```variables: {  }``` - render
4. ```variables: {  }``` - render

JohnCido:
I have a project that heavily relies on project variables to apply themes. Is there a timeline for publishing this as a fix on NPM?

aarthificial:
@JohnCido The next version should be out in a day or two.

PR #690, opened 2023-05-17
fix: project `variables` #689

rtkid-nt:


packages/core/src/app/Renderer.ts
```diff
@@ -80,6 +80,7 @@ export class Renderer {
         resolutionScale: 1,
         timeEventsClass: ReadOnlyTimeEvents,
       });
+      scene.variables.updateSignals(this.project.variables ?? {});
```
aarthificial, line 83: I think we should do that in `reloadScenes` to stay consistent with the `Presenter` class.

aarthificial (changes requested):

aarthificial (approved):

aarthificial:
Thanks for the fix!
As a side note: if your PR fixes an issue, you should put it in the body of the commit message like this:
```
Closes: #689 
```
GitHub will then automatically close the issue upon merging

PR #696, opened 2023-05-19
docs: fix the code example to retrieve correct generator value

Proful:
Animation flow minor docs change for code example.


aarthificial:
Thanks for the fix!

PR #709, opened 2023-05-23
feat(core): add `gauss` function to `Random`

ajs1998:
`gauss(mean, stdev)` will return a normally distributed random value with a given `mean` and `stdev`.

The implementation was ported to TypeScript from Python's `random.gauss()`: https://docs.python.org/3/library/random.html#random.gauss

It includes a small optimization that caches a second independent random value
since the function always generates a pair.

### DIFF
--- a/packages/core/src/scenes/Random.ts
+++ b/packages/core/src/scenes/Random.ts
@@ -40,6 +40,28 @@ export class Random {
     return value;
   }
 
+  // This is an optimization.
+  // Since gauss() generates a pair of independent Gaussian random numbers,
+  // it returns one immediately and stores the other for the next call to gauss().
+  private nextGauss: number | null = null;
+
+  /**
+   * Get a random float from a gaussian distribution.
+   * @param mean - The mean of the distribution.
+   * @param stdev - The standard deviation of the distribution.
+   */
+  public gauss(mean = 0.0, stdev = 1.0): number {
+    let z = this.nextGauss;
+    this.nextGauss = null;
+    if (z === null) {
+      const x2pi = this.next() * 2 * Math.PI;
+      const g2rad = Math.sqrt(-2.0 * Math.log(1.0 - this.next()));
+      z = Math.cos(x2pi) * g2rad;
+      this.nextGauss = Math.sin(x2pi) * g2rad;
+    }
+    return mean + z * stdev;
+  }
+
   /**
    * Get an array filled with random floats in the given range.
    *
--- a/packages/core/src/scenes/Random.ts
+++ b/packages/core/src/scenes/Random.ts
@@ -6,6 +6,17 @@ import {range} from '../utils';
  * {@link https://gist.github.com/tommyettinger/46a874533244883189143505d203312c | Mulberry32}.
  */
 export class Random {
+  /**
+   * Previously generated Gaussian random number.
+   *
+   * @remarks
+   * This is an optimization.
+   * Since {@link gauss} generates a pair of independent Gaussian random
+   * numbers, it returns one immediately and stores the other for the next call
+   * to {@link gauss}.
+   */
+  private nextGauss: number | null = null;
+
   public constructor(private state: number) {}
 
   /**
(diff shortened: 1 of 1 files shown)


packages/core/src/scenes/Random.ts
```diff
@@ -40,6 +40,28 @@ export class Random {
     return value;
   }
 
+  // This is an optimization.
+  // Since gauss() generates a pair of independent Gaussian random numbers,
+  // it returns one immediately and stores the other for the next call to gauss().
+  private nextGauss: number | null = null;
```
ajs1998, line 46: I can move this to the top of the file or wherever. I thought it was most clear right above the `gauss()` method.

I would also like to test this but I'm not sure how. I ran `npx lerna run build` but I don't know how to use the built targets in my test project.

packages/core/src/scenes/Random.ts
```diff
@@ -40,6 +40,28 @@ export class Random {
...
+    let z = this.nextGauss;
+    this.nextGauss = null;
+    if (z === null) {
+      const x2pi = this.next() * 2 * Math.PI;
+      const g2rad = Math.sqrt(-2.0 * Math.log(1.0 - this.next()));
+      z = Math.cos(x2pi) * g2rad;
+      this.nextGauss = Math.sin(x2pi) * g2rad;
+    }
+    return mean + z * stdev;
+  }
```
ajs1998, line 63: For reference, this is Python's implementation:

```python
def gauss(self, mu=0.0, sigma=1.0):
  random = self.random
  z = self.gauss_next
  self.gauss_next = None
  if z is None:
      x2pi = random() * TWOPI
      g2rad = _sqrt(-2.0 * _log(1.0 - random()))
      z = _cos(x2pi) * g2rad
      self.gauss_next = _sin(x2pi) * g2rad

  return mu + z * sigma
```

packages/core/src/scenes/Random.ts
aarthificial: It should be moved to the very top to follow the structure of other classes:
- static properties
- static constructor
- static methods
- properties
- constructor
- methods

---

The easiest way to test new features is with the `template` package.
Running `npm run template:serve` will start an editor that uses the local builds of all packages.
It's possible to use the builds in an external project using `npm link` but it tends to be cumbersome.

packages/core/src/scenes/Random.ts
```diff
@@ -40,6 +40,28 @@ export class Random {
     return value;
   }
 
+  // This is an optimization.
+  // Since gauss() generates a pair of independent Gaussian random numbers,
+  // it returns one immediately and stores the other for the next call to gauss().
```
aarthificial, line 45: I think these remarks should be written as TSDoc:
```suggestion
  /**
   * Previously generated Gaussian random number.
   *
   * @remarks
   * This is an optimization.
   * Since {@link gauss} generates a pair of independent Gaussian random
   * numbers, it returns one immediately and stores the other for the next call
   * to {@link gauss}.
   */
```

packages/core/src/scenes/Random.ts
```diff
@@ -40,6 +40,28 @@ export class Random {
...
+   * Get a random float from a gaussian distribution.
+   * @param mean - The mean of the distribution.
+   * @param stdev - The standard deviation of the distribution.
+   */
+  public gauss(mean = 0.0, stdev = 1.0): number {
+    let z = this.nextGauss;
+    this.nextGauss = null;
+    if (z === null) {
+      const x2pi = this.next() * 2 * Math.PI;
+      const g2rad = Math.sqrt(-2.0 * Math.log(1.0 - this.next()));
```
aarthificial, line 58: JS doesn't differentiate between integers and floats so the trailing zeros are not necessary:
```suggestion
  public gauss(mean = 0, stdev = 1): number {
    let z = this.nextGauss;
    this.nextGauss = null;
    if (z === null) {
      const x2pi = this.next() * 2 * Math.PI;
      const g2rad = Math.sqrt(-2 * Math.log(1 - this.next()));
```

aarthificial (changes requested):

ajs1998:
The left is uniform distribution between 0-200

The right is gaussian distribution with a mean of 150 and standard deviation of 10

image

Here's another where the right having a mean of 150 and standard deviation of 50

image

I'm not experienced enough with Motion Canvas to make a histogram. I could probably do it though if you need further convincing that this works. Just eyeballing it I can see how the distribution isn't uniformly random, and my changes to the standard deviation match what I expect.

aarthificial (approved):

PR #719, opened 2023-05-28
Update quickstart.mdx

phdoerfler:
Added hint that the playback range selector may be the cause of prematurely ending animations


aarthificial:
Looks like you used the GitHub editor to modify these files. Unfortunately, you need to do that locally using an IDE because there are tools that need to be run to ensure the code is correctly formatted. This is described in [our contribution guide](https://github.com/motion-canvas/motion-canvas/blob/main/CONTRIBUTING.md#going-through-verification).

aarthificial:
Closing in favor of #749

PR #737, opened 2023-06-05
feat: add coordinates to preview

Logon27:
This feature was made in to address issue #289

Added Coordinates to preview. Press P to copy. There is a toggle for the coordinates in the settings. I added a video showing how the feature works. There will likely need to be some revisions to how the canvas is element is referenced. As well as probably some React best practices as this is fairly new to me.

@aarthificial whenever you have time to review, feel free.

(Video is outdated. Will be updated when comments are resolved.)
[2023-06-05 19-36-49.webm](https://github.com/motion-canvas/motion-canvas/assets/32345277/8a9afa22-5584-42e0-8247-722f502e74a7)


packages/ui/src/components/controls/Controls.module.scss
```diff
@@ -286,3 +286,22 @@
...
+  border-radius: var(--radius);
+  padding: 0 8px;
+  height: 24px;
+  flex-shrink: 1;
+  flex-grow: 1;
+  min-width: 0;
+  font-weight: bold;
+  display: block;
+  font-family: var(--font-family-mono);
+}
```
aarthificial, line 303: What's the difference between this and the `<Input>` component used for displaying framerate for example?

packages/ui/src/components/viewport/Coordinates.tsx
```diff
@@ -0,0 +1,90 @@
...
+      // This selector probably needs to be improved to select the overlay more reliably (future proof).
+      // Possibly use the internal state instead of a selector.
+      const canvas = document.getElementsByTagName('canvas')[1];
+      const rect = canvas.getBoundingClientRect();
+      const xPos = Math.round(
+        ((event.x - rect.x) / rect.width - 0.5) * canvas.width,
+      );
+      const yPos = Math.round(
+        ((event.y - rect.y) / rect.height - 0.5) * canvas.height,
+      );
```
aarthificial, line 32: The origin being located in the center of the scene is explicit to the `Scene2D` class.
Other scenes may implement a different coordinate system.

We should extend the `Inspectable` interface so that scenes can transform the mouse position however they want.
This would look something like this:
```ts
const scene = useCurrentScene();
if (isInspectable(scene)) {
  // calculate the absolute mouse position the same way as in EditorPreview.tsx:
  const absoluteCoords = ... 
  // use the new method on the Inspectable interface to transform it:
  const point = scene.transformMousePosition(absoluteCoords);
}
```

packages/ui/src/contexts/shortcuts.tsx
```diff
@@ -30,6 +30,8 @@ const InitialShortcuts: ShortcutsByModule = {
     {key: '=', action: 'Zoom in'},
     {key: '-', action: 'Zoom out'},
     {key: "'", action: 'Toggle grid'},
+    {key: 'v', action: 'Toggle coordinates'},
+    {key: 'c', action: 'Copy coordinates'},
```
aarthificial, line 34: I'm not a huge fan of adding multiple shortcuts to do one thing.
I'd imagine it working the same way as the color picker, so:
1. **Initial State** - the coordinates are displayed in the bottom left, next to a button
    a. Clicking on the button switches to the **Sampling State**
    b. Pressing `v` switches to the **Sampling State**
3. **Sampling State**
    a. Clicking LMB anywhere on the canvas copies the position.
    b. Clicking ESC cancels sampling.

packages/ui/src/components/viewport/EditorPreview.tsx
```diff
@@ -246,6 +252,8 @@ export function EditorPreview() {
             <GridIcon />
           </ButtonCheckbox>
           <ColorPicker />
+          {/* Key is set to the state so that the component renders when the state changes */}
+          <Coordinates visibility={coords} viewState={state} viewSize={size} />
```
aarthificial, line 256: `size.x` and `size.y` should be added to the `ViewportState`.
The state is already exposed via a context which means that you don't need to pass it as a property.
In `Coordinates` you can just do:
```ts
const state = useContext(ViewportContext);
```
To access the viewport state.

aarthificial, line 256: To conditionally toggle a component you simply do:
```ts
{coords && <Coordinates />}
```

packages/ui/src/components/viewport/EditorPreview.tsx
```diff
@@ -246,6 +252,8 @@ export function EditorPreview() {
             <GridIcon />
           </ButtonCheckbox>
           <ColorPicker />
+          {/* Key is set to the state so that the component renders when the state changes */}
```
aarthificial, line 255: Please elaborate

packages/ui/src/components/viewport/Coordinates.tsx
```diff
@@ -0,0 +1,90 @@
...
+      document.removeEventListener('mousemove', handleMouseMove);
+    };
+  }, [viewState, viewSize]);
+
+  // Below two methods are used for the copy to clipboard keybind
+  const copyCoords = useCallback(async (x: number, y: number) => {
+    console.log(`${x}, ${y}`);
+    const positionString = `${x}, ${y}`;
+    await window.navigator.clipboard.writeText(positionString);
+  }, []);
```
aarthificial, line 71: I don't think it's necessary to extract this method, it's only two lines of code used in just one place.

packages/ui/src/components/viewport/Coordinates.tsx
```diff
@@ -0,0 +1,90 @@
...
+    document.addEventListener('mousemove', handleMouseMove);
+
+    return () => {
+      document.removeEventListener('mousemove', handleMouseMove);
+    };
+  }, [viewState, viewSize]);
+
+  // Below two methods are used for the copy to clipboard keybind
+  const copyCoords = useCallback(async (x: number, y: number) => {
+    console.log(`${x}, ${y}`);
```
aarthificial, line 68: We shouldn't log things directly to console

aarthificial (changes requested):

packages/ui/src/components/viewport/EditorPreview.tsx
Logon27: Yeah I got some cleanup to do. That was a remnant comment after I changed the implementation. It is no longer required. I will remove.

packages/ui/src/components/viewport/Coordinates.tsx
Logon27: Very true. I will just get rid of the method and place it inline.

packages/ui/src/components/viewport/Coordinates.tsx
Logon27: Yeah that was purely for debugging on my end. It was not intended to stay. I must have missed it. Will remove.

packages/ui/src/components/viewport/EditorPreview.tsx
Logon27: Ohhhh. That's neat. I will modify the visibility and hover to just utilize that conditional rendering.

packages/ui/src/contexts/shortcuts.tsx
Logon27: I just thought I would add a toggle to turn the coordinates off in case someone did not want to see them. Do you want the coordinates always displayed?

packages/ui/src/components/viewport/EditorPreview.tsx
Logon27: So I updated the code to include the `size.x` and `size.y` in the ViewportState. Please let me know if I did not implement it the way you intended. I will look at the remaining 3 comments tomorrow as I need to go to bed.

packages/ui/src/components/controls/Controls.module.scss
Logon27: I could not figure out how to resize the input component so it was only the length of the text. That is the only difference.

packages/ui/src/components/viewport/Coordinates.tsx
Logon27: I understand what you mean here. But inside the Scene2D class how would I get the rect.x, rect.y, rect.width, and rect.height to perform the calculation?
```
public transformMousePosition(x: number, y: number): number | null {
  //   ???
}
```

Logon27: Would I just do this.getSize().x and this.getView().width?

Logon27: So the quoted code really only calculates the edge of the overlay window for the preview for bounds detection. The actual point calculation is done with...
```
let canvasPosX = event.x - state.size.x;
let canvasPosY = event.y - state.size.y;

canvasPosX -= state.x + state.width / 2;
canvasPosY -= state.y + state.height / 2;
canvasPosX /= state.zoom;
canvasPosY /= state.zoom;

canvasPosX = Math.round(canvasPosX);
canvasPosY = Math.round(canvasPosY);
```
But how would I get the Viewport state from within the scene's transformMousePosition without passing it as a parameter to the function?

Logon27: @aarthificial Can you provide me a bit more guidance on this one? I don't really know how to retrieve the ViewPort state from within the 'core' part of motion canvas. And when I try to use a transformMousePosition method (after added it as a interface and adding the method to the 2D scene class) my static analysis complains because...
```
const scene = useCurrentScene();
```
returns a Scene of type 'unknown'.

packages/ui/src/contexts/shortcuts.tsx
aarthificial: Yes, we could also go this way.
If you think hiding the coordinates can be useful then we can add it as a checkbox to the application settings.

packages/ui/src/components/viewport/Coordinates.tsx
aarthificial: Did you use the `isInspectable` function like this?
```ts
if (isInspectable(scene)) {
  // `scene` in this scope should be of type `Inspectable`
}
// outside the scope `scene` is back to being `Scene<unknown>`
```

packages/ui/src/components/controls/Controls.module.scss
aarthificial: I'd still use the `.input` class then, just with a `div` element.
The size being constant is caused by the `input` element itself and not the styling so it should be possible to use the same class.

packages/ui/src/components/viewport/Coordinates.tsx
Logon27: > Did you use the `isInspectable` function like this?
> 
> ```ts
> if (isInspectable(scene)) {
>   // `scene` in this scope should be of type `Inspectable`
> }
> // outside the scope `scene` is back to being `Scene<unknown>`
> ```

Yes, I did. It was considered to be of type 'Scene<unknown> & Inspectable'

Logon27: @aarthificial So basically my issues is...

It does not return the Scene type after I check if it is inspectable. Just scene type unknown. And I am not quite sure how to cast this scene to a specific type. Possibly a modification to the useCurrentScene function is required to return a scene with a specific type? I just am not sure how. Because I do not want to explicitly cast using...
const point = (scene as Scene2D).transformMousePosition(absoluteCoords);
because it is not guaranteed to always be Scene2D.

I 'think' that I figured out how to pull the viewport state in the same way as before with...
const state = useContext(ViewportContext);
with my issue being just an import path error. But I will have to wait and see on that one.

Logon27: I spoke with Hunter. He helped me resolve the issue. I did not know that I had to run a lerna build for cross package changes for the static analysis to update properly.

packages/ui/src/components/controls/Controls.module.scss
Logon27: Fair point. I just styled a div with the same `.input` class and remove my InfoBox component.

packages/ui/src/components/viewport/Coordinates.tsx
Logon27: I added the `transformMousePosition` function to the Inspectable interface and implemented it inside the Scene2D class. You will have to let me know if this implementation is sufficient.

packages/ui/src/contexts/shortcuts.tsx
Logon27: I added the coordinates as a toggle in the settings. I did not make it a select button like the color picker. Currently its just a keybind to copy the coordinates. This is kind of a stretch goal for me if I have the time.

packages/2d/src/scenes/Scene2D.ts
```diff
@@ -117,6 +117,22 @@ export class Scene2D extends GeneratorScene<View2D> implements Inspectable {
     }
   }
 
+  public transformMousePosition(
+    x: number,
+    y: number,
+    size?: {width: number; height: number},
+  ): Vector2 | null {
+    if (size !== undefined) {
```
hhenrichsen, line 125: nit: `if (size)`

packages/core/src/scenes/Inspectable.ts
```diff
@@ -67,6 +69,19 @@ export interface Inspectable {
...
+   * Transform the absolute mouse coordinates into the scene's coordinate system.
+   *
+   * @param x - The x coordinate.
+   * @param y - The y coordinate.
+   * @param size - (Optional) The size from the shared meta settings. Used to calculate coordinate offsets
+   */
+  transformMousePosition(
+    x: number,
+    y: number,
+    size?: {width: number; height: number},
```
aarthificial, line 83: There's no need to pass this value all the way from the editor. Scenes have access to the size of the project via the `this.size` property.

packages/2d/src/scenes/Scene2D.ts
```diff
@@ -117,6 +117,22 @@ export class Scene2D extends GeneratorScene<View2D> implements Inspectable {
...
+      x -= size.width / 2;
+      y -= size.height / 2;
+    }
+
+    x = Math.round(x);
+    y = Math.round(y);
+
+    return new Vector2(x, y);
```
aarthificial, line 133: Coordinates shouldn't be transformed manually - you should use the view matrix instead.
This is the same logic we use for hit detection:
```suggestion
    return new Vector2(x, y)
      .scale(this.resolutionScale)
      .transformAsPoint(this.getView().localToParent().inverse());
```

packages/2d/src/scenes/Scene2D.ts
```diff
@@ -117,6 +117,22 @@ export class Scene2D extends GeneratorScene<View2D> implements Inspectable {
...
+    y: number,
+    size?: {width: number; height: number},
+  ): Vector2 | null {
+    if (size !== undefined) {
+      x -= size.width / 2;
+      y -= size.height / 2;
+    }
+
+    x = Math.round(x);
+    y = Math.round(y);
```
aarthificial, line 131: Rounding should be done on the `ui` side when displaying the values

packages/ui/src/components/viewport/Coordinates.tsx
```diff
@@ -0,0 +1,97 @@
...
+import {useCallback, useContext} from 'preact/hooks';
+import {
+  useCurrentScene,
+  useDocumentEvent,
+  useSharedSettings,
+} from '../../hooks';
+import {ViewportContext} from './ViewportContext';
+import {isInspectable} from '@motion-canvas/core';
+import clsx from 'clsx';
+import styles from '@motion-canvas/ui/src/components/controls/Controls.module.scss';
```
aarthificial, line 11:
```suggestion
import styles from '../controls/Controls.module.scss';
```

packages/ui/src/components/viewport/Coordinates.tsx
```diff
@@ -0,0 +1,97 @@
...
+        }
+      },
+      [mousePos],
+    ),
+  );
+
+  return (
+    <>
+      {hover && (
+        <div title={'Coordinates'} className={clsx(styles.input)}>
```
aarthificial, line 91: `clsx` is for joining multiple classes together
```suggestion
        <div title={'Coordinates'} className={styles.input}>
```

packages/core/src/app/SettingsMetadata.ts
```diff
@@ -20,6 +20,9 @@ export function createSettingsMetadata() {
       font: new BoolMetaField('legacy font', false).describe(
         "Use the 'JetBrains Mono' font for the user interface.",
       ),
+      coordinates: new BoolMetaField('coordinates', true).describe(
+        'Display coordinates within the preview window.',
```
aarthificial, line 24:
```suggestion
        'Display mouse coordinates within the preview window.',
```

aarthificial (changes requested):

packages/core/src/app/SettingsMetadata.ts
Logon27: Modified the hover message to the suggested.

packages/ui/src/components/viewport/Coordinates.tsx
Logon27: Applied.

packages/ui/src/components/viewport/Coordinates.tsx
Logon27: Applied.

packages/2d/src/scenes/Scene2D.ts
Logon27: Rounding was moved to the UI side.

packages/2d/src/scenes/Scene2D.ts
Logon27: Applied.

packages/core/src/scenes/Inspectable.ts
Logon27: Size property was removed from the interface. I applied to fix from your other comment.

packages/2d/src/scenes/Scene2D.ts
Logon27: This code section was removed. Resolving comment.

packages/ui/src/contexts/shortcuts.tsx
Logon27: @aarthificial Couple clarity questions if you want it to be a sampling state instead.

For the sampling state do you want the same EyeDropper selection as the color picker? Or do you want something different?

What svg icon would you like for the button itself?

aarthificial (approved):

PR #739, opened 2023-06-11
fix(ui): add min-width and z-index on hover to range handle

### LINKED ISSUE #738: Support small ranges
**Describe the bug**
When the animation range is too small, the timeline handles disappear, and resizing becomes impossible.

**To Reproduce**
1. Set the start and end of the range to the same value.
2. Try to resize it using the timeline gizmo.

**Expected behavior**
Handles should still be visible so that the range can be resized.

**Package versions:**

- core: 3.9.0
- two: 3.9.0
- ui: 3.9.0
- vitePlugin: 3.9.0

Caesarovich:
Fixes #738 

Now the range handles are visible whenever range is hovered. 
So even when zooming out a lot it's still possible to change the range.

Let me know if you think a better solution could be done.

https://github.com/motion-canvas/motion-canvas/assets/38408878/24aaff8d-9511-48d3-8857-d1342317fa89

### DIFF
--- a/packages/ui/src/components/timeline/Timeline.module.scss
+++ b/packages/ui/src/components/timeline/Timeline.module.scss
@@ -302,6 +302,8 @@
 
   .range:hover > & {
     color: rgba(255, 255, 255, 0.24);
+    z-index: 1;
+    min-width: 24px;
   }
 
   .range > &:hover,


aarthificial (approved):

aarthificial:
Thanks for the fix!

PR #746, opened 2023-06-16
Docs: Fix makeRef import for Looping example code snippet

AkshatGiri:
importing makeRef from '@motion-canvas/core/lib/flow' does not work the javascript project. It has to be imported from @motion-canvas/core


aarthificial:
Thanks for the fix!

PR #761, opened 2023-07-03
feat(2d): add start and end signals to Grid node

ksassnowski:
This PR adds `start` and `end` signals to the `Grid` node. These signals can be used to animate the grid appearing on screen similar to lines or curves work.

## Example

```tsx
  const grid = createRef<Grid>();

  view.add(
    <Grid
      ref={grid}
      width={'100%'}
      height={'100%'}
      stroke={'#666'}
      start={0.5}
      end={0.5}
    />,
  );

  yield* all(
    grid().end(1, 1).wait(1).to(0.5, 1),
    grid().start(0, 1).wait(1).to(0.5, 1),
  );
});
```

https://github.com/motion-canvas/motion-canvas/assets/5139098/38f15add-dbb1-4c91-9a7c-8e0283251b8f

### DIFF
--- a/packages/2d/src/components/Grid.ts
+++ b/packages/2d/src/components/Grid.ts
@@ -40,14 +40,14 @@ export interface GridProps extends ShapeProps {
  *       width={'100%'}
  *       height={'100%'}
  *       stroke={'#666'}
- *       start={0.5}
- *       end={0.5}
+ *       start={0}
+ *       end={1}
  *     />,
  *   );
  *
  *   yield* all(
- *     grid().end(1, 1).wait(1).to(0.5, 1),
- *     grid().start(0, 1).wait(1).to(0.5, 1),
+ *     grid().end(0.5, 1).to(1, 1).wait(1),
+ *     grid().start(0.5, 1).to(0, 1).wait(1),
  *   );
  * });
  * ```
@@ -67,7 +67,7 @@ export class Grid extends Shape {
    * The portion of each grid line that comes before the given percentage will
    * be made invisible.
    *
-   * This property is usefully for animating the grid appearing on screen.
+   * This property is useful for animating the grid appearing on-screen.
    */
   @initial(0)
   @signal()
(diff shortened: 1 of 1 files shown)


packages/2d/src/components/Grid.ts
```diff
@@ -1,17 +1,91 @@
...
+ *       height={'100%'}
+ *       stroke={'#666'}
+ *       start={0.5}
+ *       end={0.5}
+ *     />,
+ *   );
+ *
+ *   yield* all(
+ *     grid().end(1, 1).wait(1).to(0.5, 1),
+ *     grid().start(0, 1).wait(1).to(0.5, 1),
```
aarthificial, line 50: When providing examples for Nodes, they should be initially visible and then hide/reappear
So that even if they don't press play, the node will be immediately visible.
```suggestion
 *       start={0}
 *       end={1}
 *     />,
 *   );
 *
 *   yield* all(
 *     grid().end(0.5, 1).to(1, 1).wait(1),
 *     grid().start(0.5, 1).to(0, 1).wait(1),
```

packages/2d/src/components/Grid.ts
```diff
@@ -1,17 +1,91 @@
...
   public declare readonly spacing: Vector2Signal<this>;
 
+  /**
+   * The percentage that should be clipped from the beginning of each grid line.
+   *
+   * @remarks
+   * The portion of each grid line that comes before the given percentage will
+   * be made invisible.
+   *
+   * This property is usefully for animating the grid appearing on screen.
```
aarthificial, line 70:
```suggestion
   * This property is useful for animating the grid appearing on the screen.
```

packages/2d/src/components/Grid.ts
```diff
@@ -1,17 +1,91 @@
...
+  public declare readonly start: SimpleSignal<number, this>;
+
+  /**
+   * The percentage that should be clipped from the end of each grid line.
+   *
+   * @remarks
+   * The portion of each grid line that comes after the given percentage will
+   * be made invisible.
+   *
+   * This property is usefully for animating the grid appearing on screen.
```
aarthificial, line 83:
```suggestion
   * This property is useful for animating the grid appearing on the screen.
```

aarthificial (changes requested):

aarthificial (approved):

PR #763, opened 2023-07-04
feat(2d): add SVG component

levirs565:
This will add `SVG` component. This component can render SVG and tween between two SVG.

This is part of [add Latex component pull request](https://github.com/motion-canvas/motion-canvas/pull/539).

## Tweening Algorithm

SVG tweening algorithm support insertion, deletion and tranformation. This algorithm is based on [ManimCommunity TransformMatchingShapes](https://github.com/ManimCommunity/manim/blob/main/manim/animation/transform_matching_parts.py)

This algorithm also support SVG that contain 2 or more element with same id. When number of element with some id in source and target do not match, all element will be transformed wihout insertion or deletion. 

When number of element with some id in source is greater than number of element with that id then excess element will be transformed into target last element with that id.

```
[1, 2, 1, 3, 1] -> [1, 2, 3]
```

> Second and third source element with id 1 will be transformed into first target element with id 1.

When number of element with some id in source is lower than number of element with that id then source last element with that id will be cloned and then transformed target element that index is matched.

```
[1, 2, 3] --> [1, 2, 1, 3, 1]
```

> First source element with id 1 will cloned two time and the two item will transformed to second and third target element with id 1 respectively

This algorithm does not support moving item index. Why? Because moving item index cannot transformed/animated smoothly.

This algorithm will insert element based on previous element not index. Why? For simplicity. When, there are inserted element and element previous item is moved then the element will inserted after previous item.

```
[1, 2, 3] --> [1, 3, 5, 2]

result: [1,2,3,5]
```
> 5 is after 3, 5 is inserted after 3.

### DIFF
--- a/packages/2d/src/components/SVG.ts
+++ b/packages/2d/src/components/SVG.ts
@@ -26,6 +26,7 @@ import {ThreadGenerator} from '@motion-canvas/core/lib/threading';
 import {Circle, CircleProps} from './Circle';
 import {Line, LineProps} from './Line';
 import {Img, ImgProps} from './Img';
+import {applyTransformDiff, getTransformDiff} from '../utils/diff';
 
 export interface SVGShape {
   id: string;
@@ -47,24 +48,6 @@ export interface SVGDocument {
 export interface SVGDocumentData {
   size: Vector2;
   nodes: SVGShapeData[];
-}
-
-interface SVGDiffShape {
-  index: number;
-  node: SVGShape;
-}
-
-interface SVGDiff {
-  fromSize: Vector2;
-  toSize: Vector2;
-  inserted: Array<SVGDiffShape>;
-  deleted: Array<SVGDiffShape>;
-  transformed: Array<{
-    insert: boolean;
-    deleted: boolean;
-    from: SVGDiffShape;
-    to: SVGDiffShape;
-  }>;
 }
 
 export interface SVGProps extends ShapeProps {
@@ -423,72 +406,6 @@ export class SVG extends Shape {
     }
   }
 
-  private getShapeMap(svg: SVGDocument) {
-    const map: Record<string, SVGDiffShape[]> = {};
-    for (const [index, node] of svg.nodes.entries()) {
-      if (!map[node.id]) map[node.id] = [];
-
-      map[node.id].push({index, node});
-    }
-    return map;
-  }
-
-  private diffSVG(from: SVGDocument, to: SVGDocument): SVGDiff {
-    const diff: SVGDiff = {
-      fromSize: from.size,
-      toSize: to.size,
-      inserted: [],
-      deleted: [],
-      transformed: [],
-    };
-
-    const fromMap = this.getShapeMap(from);
-    const fromKeys = Object.keys(fromMap);
-    const toMap = this.getShapeMap(to);
-    const toKeys = Object.keys(toMap);
-
-    while (fromKeys.length > 0) {
-      const key = fromKeys.pop()!;
-      const fromItem = fromMap[key];
-      const toIndex = toKeys.indexOf(key);
-      if (toIndex >= 0) {
-        toKeys.splice(toIndex, 1);
-        const toItem = toMap[key];
-
-        for (let i = 0; i < Math.max(fromItem.length, toItem.length); i++) {
-          const insert = i >= fromItem.length;
-          const deleted = i >= toItem.length;
-
-          const fromNode =
-            i < fromItem.length ? fromItem[i] : fromItem[fromItem.length - 1];
-          const toNode =
-            i < toItem.length ? toItem[i] : toItem[toItem.length - 1];
-
-          diff.transformed.push({
-            insert,
-            deleted,
-            from: fromNode,
-            to: toNode,
-          });
-        }
-      } else {
-        for (const node of fromItem) {
-          diff.deleted.push(node);
-        }
-      }
-    }
-
-    if (toKeys.length > 0) {
-      for (const key of toKeys) {
-        for (const node of toMap[key]) {
-          diff.inserted.push(node);
-        }
-      }
-    }
-
-    return diff;
-  }
-
   private cloneNodeExact(node: Node) {
     const props: ShapeProps = {
       position: node.position(),
@@ -539,66 +456,6 @@ export class SVG extends Shape {
     }
   }
 
-  private useTweenInitialNodes(diff: SVGDiff) {
-    const insertedList: SVGDiff['transformed'] = [];
-    for (const item of diff.transformed) {
-      if (!item.insert) continue;
-
-      const from = item.from;
-      item.from = {
-        index: from.index,
-        node: {
-          id: from.node.id,
-          shape: this.cloneNodeExact(from.node.shape),
-        },
-      };
-      insertedList.push(item);
-    }
-
-    const newChildren = this.wrapper.children().slice(0);
-    let insertShift = 0;
-
-    for (const {from} of insertedList.sort(
-      (a, b) => a.from.index - b.from.index,
-    )) {
-      newChildren.splice(from.index + insertShift, 0, from.node.shape);
-      from.node.shape.parent(this.wrapper);
-      insertShift++;
-    }
-
-    this.wrapper.children(newChildren);
-  }
-
-  private useTweenFinalNodes(diff: SVGDiff, newSVG: SVGDocument) {
-    const newChildren: Node[] = new Array(newSVG.nodes.length);
-    const deletedList: SVGDiff['transformed'] = [];
-
-    for (const item of diff.transformed) {
-      const {from, to, deleted} = item;
-      if (deleted) {
-        deletedList.push(item);
-        continue;
-      }
-      newChildren[to.index] = from.node.shape;
-      from.node.shape.parent(this.wrapper);
-    }
-    for (const {node, index} of diff.inserted) {
-      newChildren[index] = node.shape;
-      node.shape.parent(this.wrapper);
-    }
-
-    let insertShift = 0;
-    for (const {from, to} of deletedList.sort(
-      (a, b) => a.to.index - b.to.index,
-    )) {
-      newChildren.splice(to.index + insertShift, 0, from.node.shape);
-      from.node.shape.parent(this.wrapper);
-      insertShift++;
-    }
-
-    this.wrapper.children(newChildren);
-  }
-
   @threadable()
   protected *tweenSvg(
     value: SignalValue<string>,
@@ -607,7 +464,21 @@ export class SVG extends Shape {
   ) {
     const newValue = typeof value === 'string' ? value : value();
     const newSVG = this.parseSVG(newValue);
-    const diff = this.diffSVG(this.document(), newSVG);
+    const currentSVG = this.document();
+    const diff = getTransformDiff(currentSVG.nodes, newSVG.nodes);
+
+    const applyResult = applyTransformDiff(
+      currentSVG.nodes,
+      diff,
+      ({shape, ...rest}) => ({
+        ...rest,
+        shape: this.cloneNodeExact(shape),
+      }),
+    );
+    this.wrapper.children(currentSVG.nodes.map(shape => shape.shape));
+    for (const item of applyResult.inserted) {
+      item.current.shape.parent(this.wrapper);
+    }
 
     const beginning = 0.2;
     const ending = 0.8;
@@ -617,25 +488,16 @@ export class SVG extends Shape {
     const transformatorTime = (ending - beginning) * time;
     const transformatorDelay = beginning * time;
 
-    this.useTweenInitialNodes(diff);
-
-    for (const transformed of diff.transformed) {
+    for (const item of diff.transformed) {
       transformator.push(
         ...this.generateTransformator(
-          transformed.from.node.shape,
-          transformed.to.node.shape,
+          item.from.current.shape,
+          item.to.current.shape,
           transformatorTime,
           timingFunction,
         ),
       );
     }
-
-    let reordered = false;
-    const reorder = () => {
-      if (reordered) return;
-      this.useTweenFinalNodes(diff, newSVG);
-      reordered = true;
-    };
 
     const autoWidth = this.width.isInitial();
     const autoHeight = this.height.isInitial();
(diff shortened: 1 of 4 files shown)


levirs565:
@aarthificial Are you accept the tweening algorithm? If not, please give me Suggestion about tweening algoritm

Firnael:
Hi @levirs565 , with this final part merged, will LaTex tweening be finally possible ?
Thank you for your work here.

packages/2d/src/utils/diff.ts
```diff
@@ -0,0 +1,149 @@
...
+
+interface Idable {
+  id: string;
+}
+
+function getIdMap<T extends Idable>(list: T[]) {
+  const map = new Map<string, TransformDiffItem<T>[]>();
+  let before: T | undefined = undefined;
+  for (const [index, current] of list.entries()) {
+    let currentArray = map.get(current.id);
```
hhenrichsen, line 38: nit: I don't know if this is any better, but this is one line shorter and preserves immutability:
```tsx
const currentArray = map.get(current.id) ?? [];
if (!map.has(current.id)) {
    map.set(current.id, currentArray);
}
```

packages/2d/src/components/SVG.ts
```diff
@@ -0,0 +1,560 @@
...
+      parentTransform,
+    );
+    const style = SVG.getElementStyle(child, inheritedStyle);
+    const id = child.id ?? '';
+    if (child.tagName === 'g') {
+      yield* SVG.extractGroupNodes(child, svgRoot, transformMatrix, style);
+    } else if (child.tagName === 'use') {
+      const hrefElement = svgRoot.querySelector(
+        (child as SVGUseElement).href.baseVal,
+      )!;
```
hhenrichsen, line 303: Generally, I dislike the `!` operator to unnull something, especially because you effectively check for nulls on the next line anyways.

levirs565:
> Hi @levirs565 , with this final part merged, will LaTex tweening be finally possible ? Thank you for your work here.

Yes, Latex tweening will be possible. How we insert new node when Tweening SVG is not affecting Latex tweening. Insertion order when tweening Latex is not important.

levirs565:
In my opinion, this pull request are ready

packages/2d/src/components/SVG.ts
```diff
@@ -0,0 +1,684 @@
+import {SignalValue, SimpleSignal} from '@motion-canvas/core/lib/signals';
```
ksassnowski, line 1: An export for this component should also be added to `packages/2d/src/components/index.ts`

packages/2d/src/components/SVG.ts
```diff
@@ -0,0 +1,684 @@
...
+  }
+
+  /**
+   * Create tweening list to tween node from initial node to final node.
+   * @param from - Initial node
+   * @param to - Final node
+   * @param time - time for tweening
+   * @param timing - timing for tweening
+   */
+  protected *generateTransformator(
```
ksassnowski, line 173: `Transformator` -> `Transformer`. There are a few other occurrences like this in the PR.

packages/2d/src/components/SVG.ts
```diff
@@ -0,0 +1,684 @@
...
+    }
+  }
+
+  @threadable()
+  protected *tweenSvg(
+    value: SignalValue<string>,
+    time: number,
+    timingFunction: TimingFunction,
+  ) {
+    const newValue = typeof value === 'string' ? value : value();
```
ksassnowski, line 216: Should use
```ts
import { isReactive } from '@motion-canvas/core';

const newValue = isReactive(value) ? value() : value;
```

ksassnowski (changes requested):
When trying to tween between SVGs that use `stroke-linecap="round"` and/or `stroke-linejoin="round"`, the line cap and line join get discarded during the tween, then suddenly get reapplied after the tween has finished.

Code to reproduce 

```ts
import {createRef} from '@motion-canvas/core';
import {SVG} from '@motion-canvas/2d/lib/components/SVG';

const icon1 = `
<svg width="128" height="128" fill="none" viewBox="0 0 24 24">
  <path stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5" d="M16.75 17.25H17.25C18.3546 17.25 19.25 16.3546 19.25 15.25V6.75C19.25 5.64543 18.3546 4.75 17.25 4.75H6.75C5.64543 4.75 4.75 5.64543 4.75 6.75V15.25C4.75 16.3546 5.64543 17.25 6.75 17.25H7.25"></path>
  <path stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5" d="M12 15.75L9 19.25H15L12 15.75Z"></path>
</svg>
`;

const icon2 = `
<svg width="128" height="128" fill="none" viewBox="0 0 24 24">
  <path stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5" d="M4.75 8.75C4.75 7.64543 5.64543 6.75 6.75 6.75H15.25C16.3546 6.75 17.25 7.64543 17.25 8.75V15.25C17.25 16.3546 16.3546 17.25 15.25 17.25H6.75C5.64543 17.25 4.75 16.3546 4.75 15.25V8.75Z"></path>
  <path stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5" d="M17.75 10.75H18C18.6904 10.75 19.25 11.3096 19.25 12V12C19.25 12.6904 18.6904 13.25 18 13.25H17.75"></path>
</svg>
`;

  const svg = createRef<SVG>();

  view.add(<SVG ref={svg} svg={icon1} stroke={'white'} />);

  yield* svg().svg(icon2, 1);
});
```

https://github.com/motion-canvas/motion-canvas/assets/5139098/07d10e9b-3f8a-4d5e-8cf3-610d02f615eb


I'm not sure if this is something we can realistically fix, though.

ksassnowski:
I also think that the `SVG` component should use the node's `size` signal (if provided) to set the `width` and `height` of the actual SVG DOM node. From what I can tell, you have to explicitly change the `width` and `height` attributes of the actual SVG string if you want to make it smaller or larger at the moment.

packages/2d/src/components/SVG.ts
```diff
@@ -0,0 +1,684 @@
...
+    return transformMatrix;
+  }
+
+  /**
+   * Convert SVG style into MotionCanvas Shape
+   * @param element - SVG element
+   * @param inheritedStyle - Parent style that must inherited to this element
+   * @returns - MotionCanvas Shape properties
+   */
+  private static getElementStyle(
```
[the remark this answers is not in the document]
levirs565: SVG has `opacity` attribute. This property must converted to MotionCanvas `opacity` property instead of `stroke-opacity`.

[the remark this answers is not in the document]
levirs565: I cannot find MotionCanvas property that is simmiliar to SVG `stroke-opacity` property.

levirs565:
> 



> When trying to tween between SVGs that use `stroke-linecap="round"` and/or `stroke-linejoin="round"`, the line cap and line join get discarded during the tween, then suddenly get reapplied after the tween has finished.
> 
> Code to reproduce
> 
> ```ts
> import {makeScene2D} from '@motion-canvas/2d';
> import {createRef} from '@motion-canvas/core';
> import {SVG} from '@motion-canvas/2d/lib/components/SVG';
> 
> const icon1 = `
> <svg width="128" height="128" fill="none" viewBox="0 0 24 24">
>   <path stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5" d="M16.75 17.25H17.25C18.3546 17.25 19.25 16.3546 19.25 15.25V6.75C19.25 5.64543 18.3546 4.75 17.25 4.75H6.75C5.64543 4.75 4.75 5.64543 4.75 6.75V15.25C4.75 16.3546 5.64543 17.25 6.75 17.25H7.25"></path>
>   <path stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5" d="M12 15.75L9 19.25H15L12 15.75Z"></path>
> </svg>
> `;
> 
> const icon2 = `
> <svg width="128" height="128" fill="none" viewBox="0 0 24 24">
>   <path stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5" d="M4.75 8.75C4.75 7.64543 5.64543 6.75 6.75 6.75H15.25C16.3546 6.75 17.25 7.64543 17.25 8.75V15.25C17.25 16.3546 16.3546 17.25 15.25 17.25H6.75C5.64543 17.25 4.75 16.3546 4.75 15.25V8.75Z"></path>
>   <path stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5" d="M17.75 10.75H18C18.6904 10.75 19.25 11.3096 19.25 12V12C19.25 12.6904 18.6904 13.25 18 13.25H17.75"></path>
> </svg>
> `;
> 
> export default makeScene2D(function* (view) {
>   const svg = createRef<SVG>();
> 
>   view.add(<SVG ref={svg} svg={icon1} stroke={'white'} />);
> 
>   yield* svg().svg(icon2, 1);
> });
> ```
> 
>  svgtween.mp4 
> I'm not sure if this is something we can realistically fix, though.

The problem is not from SVG component but from Path component. To fix this issue you must make Path interpolation precision higher (lower number). By default path component interpolation precision is 5. When we set this to 1, this issue will fixed.

Currently, to change the precision we must edit [`createCurveProfileLerp.ts`](https://github.com/motion-canvas/motion-canvas/blob/main/packages/2d/src/curves/createCurveProfileLerp.ts) file. Below are place of the precision:

```ts
export function createCurveProfileLerp(a: CurveProfile, b: CurveProfile) {
  const interpolations = getInterpolationPolygon(a, b, 5, 4);
  // more
}
```

The precision is 5. Maybe, we must allow user to change the interpolation precision.

levirs565:
> I also think that the `SVG` component should use the node's `size` signal (if provided) to set the `width` and `height` of the actual SVG DOM node. From what I can tell, you have to explicitly change the `width` and `height` attributes of the actual SVG string if you want to make it smaller or larger at the moment.

Current available way to change width or height we can change SVG wrapper scale.

```ts
svg().wrapper.scale([2, 1]);
````

**Comment by levirs565** [comment]
> I also think that the `SVG` component should use the node's `size` signal (if provided) to set the `width` and `height` of the actual SVG DOM node. From what I can tell, you have to explicitly change the `width` and `height` attributes of the actual SVG string if you want to make it smaller or larger at the moment.

Now, we can change SVG size with `width` or `height` property. When only `width` or only `height` property are specified, unspecified property will follow specified property and SVG document actual aspect ratio.

**Comment by levirs565** [comment]
Is there anything that needs to be fixed in this PR?

File: packages/2d/src/components/SVG.ts [hunk 7]
```diff
@@ -0,0 +1,791 @@
...
+   * @param to - Final node
+   * @param time - time for tweening
+   * @param timing - timing for tweening
+   */
+  protected *generateTransformer(
+    from: Node,
+    to: Node,
+    time: number,
+    timing: TimingFunction,
+  ): Generator<ThreadGenerator> {
```
aarthificial, line 208:
```suggestion
   * Create a tweening list to tween between two SVG nodes.
   * @param from - The initial node,
   * @param to - The final node.
   * @param duration - The duration of the tween.
   * @param timing - The timing function.
   */
  protected *generateTransformer(
    from: Node,
    to: Node,
    duration : number,
    timing: TimingFunction,
  ): Generator<ThreadGenerator> {
```

packages/2d/src/components/SVG.ts
```diff
@@ -0,0 +1,791 @@
...
+  private wrapperScale(): Vector2 {
+    return this.calculateWrapperScale(
+      this.document().size,
+      this.getCurrentSize(),
+    );
+  }
+
+  /**
+   * Get current SVGDocument
+   * @returns
```
aarthificial, line 356:
```suggestion
   * Get the current `SVGDocument`.
```

packages/2d/src/components/SVG.ts
```diff
@@ -0,0 +1,791 @@
...
+      }
+      return this.parseSVG(src);
+    } finally {
+      this.lastTweenTargetSrc = null;
+      this.lastTweenTargetDocument = null;
+    }
+  }
+
+  /**
+   * Get current document nodes
```
aarthificial, line 373:
```suggestion
   * Get current document nodes.
```

packages/2d/src/components/SVG.ts
```diff
@@ -0,0 +1,791 @@
...
+   */
+  @computed()
+  private documentNodes() {
+    return this.document().nodes.map(node => node.shape);
+  }
+
+  /**
+   * Convert SVG color in Shape properties into MotionCanvas color.
+   * @param param - Shape properties
+   * @returns Converted Shape properties
```
aarthificial, line 383:
```suggestion
   * Convert SVG colors in Shape properties to Motion Canvas colors.
   * @param param - Shape properties.
   * @returns Converted Shape properties.
```

packages/2d/src/components/SVG.ts
```diff
@@ -0,0 +1,791 @@
...
+      ...rest,
+    };
+  }
+
+  /**
+   * Parse SVG as SVGDocumentData
+   * @param svg - svg string to be parsed
+   * @returns SVGDocumentData that can be used to build SVGDocument
+   */
+
```
aarthificial, line 399:
```suggestion
  /**
   * Parse an SVG string as `SVGDocumentData`.
   * @param svg - And SVG string to be parsed.
   * @returns `SVGDocumentData` that can be used to build SVGDocument.
   */
```

packages/2d/src/components/SVG.ts
```diff
@@ -0,0 +1,791 @@
...
+    if (value.endsWith('%')) return parseFloat(value) / 100;
+    return parseFloat(value);
+  }
+
+  /**
+   * Convert SVG style into MotionCanvas Shape
+   * @param element - SVG element
+   * @param inheritedStyle - Parent style that must inherited to this element
+   * @returns - MotionCanvas Shape properties
+   */
```
aarthificial, line 570:
```suggestion
  /**
   * Convert the SVG element's style to a Motion Canvas Shape properties.
   * @param element - An SVG element whose style should be converted.
   * @param inheritedStyle - The parent style that should be inherited.
   */
```

aarthificial (changes requested):
Apologies for the delay.
I left some comments, mostly about the documentation.
I'd genuinely recommend using some free grammar checker because there's a lot of mistakes that could be easily picked up by such software.

There are also some problems with the implementation:
- tweening between two `polygon`s doesn't seem to work.
- percentage size is not supported (`width={'50%'}`, etc).

But I think we can fix them in future PRs to get this one done.

levirs565:
> tweening between two polygons doesn't seem to work.

In order to support this feature, the `Line` component must support `points` tweening. This must implemented in other PR.

> percentage size is not supported

When we can get calculated size, we can support percentage size. This can implemented in other PR. Latex tweening does not need this feature.

aarthificial:
> In order to support this feature, the Line component must support points tweening. This must implemented in other PR.

Yes, but `Line` already has limited support for tweening (when arrays have the same size, for example), but the `points` property is not accounted for by the `generateTransformer()` method so even that doesn't work.

> When we can get calculated size, we can support percentage size.

We'll never be able to get "calculated size" because the point of the `desiredSize` function is to provide the basis for layout calculations. If you try to get the calculated size at that point you'll create a circular dependency which I'm sure you've noticed already. Classes like `Img` and `Video` are good examples of using `ratio` to support percentage sizes.

aarthificial (approved):

aarthificial:
Anyway, thanks for the PR!

PR #802, opened 2023-09-03
feat(2d): add line counter for CodeBlock

SleeklyCoding:
I think it would be useful to add a feature that allows for retrieving the current count of CodeBlock's lines.

line-count

e.g. I'm using this for my custom 'CodeEditor' component that has numbering on the left side.

line-count-code-editor

```tsx
private createSideNumbers = (n: number): string => Array.from({ length: n }, (_, i) => i + 1).join('\n');
...
<Txt
     ...
      text={() => this.createSideNumbers(this.codeBlock().lineCount())}/>
...
```

BTW: When getting this value it has some delay. It could be probably improved in some way, but I'm not sure how.

### DIFF
--- a/packages/2d/src/components/CodeBlock.ts
+++ b/packages/2d/src/components/CodeBlock.ts
@@ -99,6 +99,38 @@ export class CodeBlock extends Shape {
   private selectionProgress = createSignal<number | null>(null);
   private oldSelection: CodeRange[] | null = null;
   private diffed: MorphToken[] | null = null;
+  private currentLineCount = 0;
+  private newLineCount = 0;
+
+  protected getLineCountOfTokenArray(tokens: Token[]): number {
+    let count = 0;
+
+    for (const token of tokens) {
+      for (let i = 0; i < token.code.length; i++) {
+        if (token.code[i] === '\n') {
+          count++;
+        }
+      }
+    }
+
+    if (tokens.length > 0) {
+      count++;
+    }
+
+    return count;
+  }
+
+  @computed()
+  public lineCount(): number {
+    const progress = this.codeProgress();
+    if (progress !== null) {
+      return Math.round(
+        map(this.currentLineCount, this.newLineCount, progress),
+      );
+    }
+
+    return this.getLineCountOfTokenArray(this.parsed());
+  }
 
   @computed()
   protected parsed() {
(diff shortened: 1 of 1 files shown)


packages/2d/src/components/CodeBlock.ts
```diff
@@ -53,6 +53,10 @@ export class CodeBlock extends Shape {
     false,
   );
 
+  @initial(0)
+  @signal()
+  public declare readonly lineCount: SimpleSignal<number, this>;
+
```
aarthificial, line 59: It doesn't really make sense for `lineCount` to be a signal since it's not possible to assign it a value in a meaningful way.
Instead, it should be a method marked as `computed`:
```suggestion
  @computed()
  public lineCount(): number {
    // calculate the count
  }
```
`Computed` can be thought of as a read-only signal and can be used in most places a signal can.

aarthificial (changes requested):

aarthificial:
> BTW: When getting this value it has some delay. It could be probably improved in some way, but I'm not sure how.

This happens because during tweening, a temporary structure stored in `this.diffed` is used instead of `this.parsed();`
You could account for that by doing the following:
1. When `tweenCode()` is called, calculate and store the number of lines in the current source code and the new one.
2. In `lineCount`, check if `this.progress() != null`, and if so, use the progress to interpolate between these two line counts.

SleeklyCoding:
@aarthificial  I applied your suggestions and looks good in my opinion.

improved-line-count

packages/2d/src/components/CodeBlock.ts
```diff
@@ -99,6 +99,40 @@ export class CodeBlock extends Shape {
...
+  @computed()
+  public getLineCount(): number {
+    const progress = this.codeProgress();
+    if (progress !== null) {
+      // Interpolate between the two line counts based on progress (lerping)
+      return Math.round(
+        this.currentLineCount +
+          progress * (this.newLineCount - this.currentLineCount),
+      );
```
aarthificial, line 131: We have a function for linear interpolation:
```suggestion
      return Math.round(
        map(this.currentLineCount, this.newLineCount, progress)
      );
```

packages/2d/src/components/CodeBlock.ts
```diff
@@ -99,6 +99,40 @@ export class CodeBlock extends Shape {
...
+
+    if (tokens.length > 0) {
+      count++;
+    }
+
+    return count;
+  }
+
+  @computed()
+  public getLineCount(): number {
```
aarthificial, line 124: By convention, computed methods are called like properties:
```suggestion
  public lineCount(): number {
```

aarthificial (changes requested):
Looks good! Needs just two little fixes

aarthificial (approved):

PR #808, opened 2023-09-28
feat(ui): vertical line on time event

### LINKED ISSUE #804: Hovering and dragging time events should display vertical line when such event happens
**Description**
When you drag time event there is a vertical line below your cursor. It's useful when synchronizing events with voice, but in my opinion this line should display "at the start" of time event rather then "where I clicked the tag".
This is how it's now:
image

This fix will make it clear where time events start and allow for easier and more precise adjustments.

**Proposed solution**
Where the line should be (red):
image

alsongarbuja:
display vertical line at start of time event on hover and drag of time event

successfull merge closes #804


aarthificial (changes requested):
The current implementation works but only when the mouse is hovering over the time event label.
It also immediately snaps to it when hovered, even if the event is not being dragged.
Maybe we could try creating a Preact signal set in `Label.tsx` when the label is being dragged? It could then be used by the timeline to offset the line.

alsongarbuja:
Ok i will work on it

alsongarbuja:
@aarthificial I had used the `getBoundingClientRect().left` method to get the left position of the `LabelClip` relative to the viewport and offset the line.

> const elem = document.elementFromPoint(event.clientX, event.clientY);
> ....
> elem.getBoundingClientRect().left + offset - rect.x

While we can use the Preact signal to offset the line only when the `LabelClip` is dragged but I am still using the `getBoundingClientRect()` method for offset calculation . So I was wondering if that the part of the implementation is okay or is there any other better way to find the relative position, I cannot find any other methods.

aarthificial:
@alsongarbuja I believe you should be able to just use `eventTime` for that. You can find it in the `Label` component.
Then, in the `Timeline` component, you can use `player.status.secondsToFrames` and `framesToPixels` to turn the event time in seconds into a correct offset in pixels.

packages/template/src/project.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
```
aarthificial, line 1: This file shouldn't be committed

packages/template/src/scenes/example.js
```diff
@@ -0,0 +1,149 @@
+'use strict';
```
aarthificial, line 1: This file shouldn't be committed

packages/ui/src/signals/labelClip.ts
```diff
@@ -0,0 +1,11 @@
+import {signal} from '@preact/signals';
+
+interface LabelClipDrag {
+  left: number;
+  dragging: boolean;
+}
+
+export const labelClipDragging = signal<LabelClipDrag>({
```
aarthificial, line 8: The name of the signal should be suffixed with `Signal` to match other signals we use.

packages/ui/src/components/timeline/Label.tsx
```diff
@@ -59,6 +61,9 @@ export function Label({event, scene}: LabelProps) {
             handleDrag(e);
           }
         }}
+        onMouseLeave={() =>
+          (labelClipDragging.value = {left: 0, dragging: false})
+        }
```
aarthificial, line 66: The second callback passed to `useDrag` is called when the dragging ends. That's when this signal should be reset.

packages/ui/src/signals/labelClip.ts
```diff
@@ -0,0 +1,11 @@
+import {signal} from '@preact/signals';
+
+interface LabelClipDrag {
+  left: number;
+  dragging: boolean;
+}
```
aarthificial, line 6: I think this could be expressed with `number | null` instead of this interface.
Right now `left` will be invalid/unusable when `dragging` is false, but there's nothing in the type that implies it.
A nullable number clearly states "if I'm `null` there's no valid number to use"

aarthificial (changes requested):

packages/template/src/project.js
alsongarbuja: Should i just delete this file in my new commit or should i revert back to the previous HEAD and not commit this files.

packages/ui/src/components/timeline/Label.tsx
alsongarbuja: When i use reset the value in the second callback there seems to be a bug when I leave the mouse click and not drag the mouse the `signal always becomes true` 

Also the `pixelsToFrames` and `secondsToFrame` methods are calculating the distance according to the timeline visible in the viewport making the vertical line appear in different positions

alsongarbuja: **When not dragging and leaving the mouse click**

https://github.com/motion-canvas/motion-canvas/assets/42911859/cfa81ce2-2d6d-4b59-8658-e688117f7c8f


**PixelsToFrames and SecondsToFrame methods**

https://github.com/motion-canvas/motion-canvas/assets/42911859/ca78d682-3232-4984-8876-f05589f685a8

aarthificial: `framesToPixels` calculates the absolute position on the timeline, not related to the view.
You just need to account for the padding on the left caused by negative frames:
```ts
const position = state.framesToPixels(
  player.status.secondsToFrames(labelClipDragging.value.left),
) + sizes.paddingLeft;
```
Note how the time of the event is expressed in seconds, so you need to first call `secondsToFrames` to get the frames and *then* `framesToPixels` to turn those frames into seconds.

aarthificial: Seems that there's a bug with `useDrag` that causes the first callback to be invoked one more time, right after the drag has finished.
The best approach would be to get rid of `useDrag` altogether (since it's a really hacky solution) and use pointer capture instead. This is something we do in newer components, you can find an example here: 
https://github.com/motion-canvas/motion-canvas/blob/main/packages/ui/src/components/layout/ResizeableLayout.tsx#L51-L67

packages/template/src/project.js
aarthificial: Whatever works for you. We use "squash and merge" in this repo to keep linear history so all your changes will be turned into a single commit regardless.

packages/ui/src/signals/labelClip.ts
```diff
@@ -0,0 +1,11 @@
+import {signal} from '@preact/signals';
+
+interface LabelClipDrag {
+  left: number | null;
+  dragging: boolean;
+}
```
aarthificial, line 6: I meant to replace the entire interface with `number | null`
Now you have both the nullable `left` and `dragging` carrying duplicated information for no reason.

To be clear, this is what I mean:
```ts
type LabelClipDrag = number | null;

export const labelClipDraggingSignal = signal<LabelClipDrag>(null);
```

packages/ui/src/components/timeline/Label.tsx
```diff
@@ -59,6 +60,18 @@ export function Label({event, scene}: LabelProps) {
...
+        }}
+        onPointerMove={event => {
+          if (event.currentTarget.hasPointerCapture(event.pointerId)) {
+            labelClipDraggingSignal.value = {left: eventTime, dragging: true};
+          }
+        }}
+        onPointerUp={event => {
+          event.currentTarget.releasePointerCapture(event.pointerId);
+          labelClipDraggingSignal.value = {left: null, dragging: false};
+        }}
```
aarthificial, line 74: You didn't remove the old `useDrag`.
It should pretty straightforward, just move the first callback from `useDrag` to `onPointerMove` and the second one to `onPointerUp`.
Same goes for `onMouseDown`, there's no need to have both it and `onPointerDown`.

packages/template/src/project.meta
```diff
@@ -4,7 +4,7 @@
     "background": "rgb(20,20,20)",
     "range": [
       0,
-      null
+      11.036440586280815
```
aarthificial, line 7: Changes to the template package also shouldn't be committed unless necessary.

aarthificial (changes requested):

packages/ui/src/signals/labelClip.ts
alsongarbuja: Changed the type to `<number | null>` for label drag signal

packages/ui/src/components/timeline/Label.tsx
alsongarbuja: Moved everything to `pointer events` removing the old `useDrag hook` and `mouse event (onMouseDown)`

packages/template/src/project.meta
alsongarbuja: Changed everything from `template package` back to their original values including

`template/src/project.meta`
```diff
"range": [
  0,
+ null
- 11.0364...
]
```

`template/src/scenes/example.meta`
```diff
"timeEvents": [
    {
      "name": "circle",
+    "targetTime": 3
-     "targetTime": 3.02....
    }
  ]
```

packages/ui/src/components/timeline/Label.tsx
```diff
@@ -48,15 +29,31 @@ export function Label({event, scene}: LabelProps) {
...
-            handleDrag(e);
+          }
+        }}
+        onPointerMove={e => {
+          if (e.currentTarget.hasPointerCapture(e.pointerId)) {
+            labelClipDraggingLeftSignal.value = eventTime;
+            setEventTime(
+              eventTime +
+                player.status.framesToSeconds(pixelsToFrames(e.movementX)),
+            );
```
aarthificial, line 48: One final thing, notice how you assign the `eventTime` here but then in the next line its value is updated to account for the drag. This causes the gray line to be delayed by one frame.
You could improve that by assigning the newly calculated value to the signal, as such
```suggestion
            const newTime = eventTime +
                player.status.framesToSeconds(pixelsToFrames(e.movementX));
            labelClipDraggingLeftSignal.value = newTime;
            setEventTime(newTime);
```

aarthificial (changes requested):

aarthificial (approved):

aarthificial:
Awesome, thanks for the PR!

PR #814, opened 2023-10-10
feat(ui): add goto start and goto end buttons

### LINKED ISSUE #812: Feature | Media control buttons for "go to start" and "go to end"
**Description**
Media control buttons for "go to start" and "go to end".

**Proposed solution**
We could use these buttons⏮and⏭respectively for start and end. These buttons could be shown in the media control area, directly visible or semi-hidden to not mess up the current UI.

**Considered alternatives**
- We can use the timeline to do this as of now, but it requires us to scroll the timeline from where it is right now
- > You can use SHIFT + ArrowLeft and SHIFT + ArrowRight to go to the start and end respectively
  - The above is great but a UI control is necessary I believe, for those times we just keep mouse in one hand and a mug of coffee in the other 😅

me-nkr:
update playback control buttons
    - `|<` for go to start
    - `<<` for previous frame
    - `>>` for next frame
    - `>|` for go to end
    
Closes #812


aarthificial (changes requested):
All icons in the editor should come from Material Icons unless something custom is necessary.
The SVG source code of your icons seems different than [the one available here](https://fonts.google.com/icons?icon.set=Material+Icons&selected=Material+Icons:fast_forward:&icon.style=Filled&icon.query=fast).

Additionally, names of icons should be the same as in Material Icons.
So in this case `SkipNext` and `SkipPrevious` should stay the same and the new ones should be called `FastRewind` and `FastForward`.

me-nkr:
It would be nice to mention the fact our icons follow Material icons. I spent a night learning SVG and creating the custom icons not knowing the source 🙂

aarthificial:
Apologies, I linked to the icons in [my comment](https://github.com/motion-canvas/motion-canvas/issues/812#issuecomment-1755919617) and thought that would be enough.
Writing a coding guide for the UI is in my backlog, I'll remember to put that information there to avoid future confusions.

aarthificial (approved):

aarthificial:
Thanks for the PR!

Issue #764, opened 2023-07-04
Project path is not correctly escaped


RodricoSBlackmanII:
**Describe the bug**
I am getting an internal server error from Vite saying it cannot parse the source for import analysis because the content contains invalid JS syntax. I've tried adding "**/*.meta" to assetsInclude and it did not fix the issue.

**To Reproduce**
1. Clone and Install repository
2. Follow instructions from [Quick Start](https://motioncanvas.io/docs/quickstart) page.
3. Run "npm run serve"
4. Attempt to go to http://localhost:9000/
5. Get a blank screen and an error in Node saying "Internal server error: Failed to parse source for import analysis because the content contains invalid JS syntax. You may need to install appropriate plugins to handle the .meta file format, or if it's an asset, add "**/*.meta" to `assetsInclude` in your configuration."

**Console errors**
![image](https://github.com/motion-canvas/motion-canvas/assets/25956755/fc3baf24-bbba-49d0-9551-bf3306ba826b)

**Package versions:**
Node.js 18.16.1
npm 9.7.2
motion-canvas 3.9.0
Vite 4.3.9

aarthificial:
Looks like our Vite plugin doesn't escape the project path.
What causes the crash is the apostrophe (`'`) in your `Rico's Things` folder.
Until we fix the problem, you can temporarily move your project to a different location without a `'` in it.

RodricoSBlackmanII:
Thank you for the help. Much appreciated.

aarthificial:
No worries, I'm leaving this issue open because we still need to fix it.

shaikan:
@aarthificial You have tagged #786 with "d-adopt-me"
so can I raise new PR for this issue ?

aarthificial:
@shaikan Yes

PR #821, opened 2023-10-15
fix: for handling unusual characters in file names

shaikan:
This fix is for solving issue [#764](https://github.com/motion-canvas/motion-canvas/issues/764) and any others that might be caused if the user's folder names include unusual characters which might be considered invalid by JS.


aarthificial:
Thanks for the fix!

Issue #822, opened 2023-10-16
Reset to the start of animation in disable loop option


Vaibhavi-Sita:
**Description**
For the disable loop option in the UI, it is ideal for the animation to go back to starting of the time frame on reaching the end of animation. This behavior is often observed in popular video editors such as Adobe Premier Pro, DaVinci Resolve, etc. 

Although this is a minor feature enhancement, this would be an huge intuitive solution to regular video/animation creators who are used to the standard behavior. 

**Proposed solution**
This is a UI enhancement. Just a quick reset to the start of animation [0th frame] on hitting the end of animation will do the trick. This behavior is similar to clicking the "Start (Left + Shift arrow)" button in the UI.

This is an intuitive solution that mimics popular and latest video/animation editors.

Vaibhavi-Sita:
Will work on this issue

PR #823, opened 2023-10-17
feat(core): seek to beginning of timeline in disable loop mode

Vaibhavi-Sita:
For disable loop mode, this enhancement will seek to the beginning of timeline. Also, this behavior can now be observed for the case where user may hide the timeline. The feature is in line with latest video/animation editors, and hence is made to fit its intuitive usage. 

Closes https://github.com/motion-canvas/motion-canvas/issues/822

### DIFF
--- a/packages/core/src/app/Player.ts
+++ b/packages/core/src/app/Player.ts
@@ -338,14 +338,15 @@ export class Player {
     }
 
     // Pause if reached the end or the range is 0
-    // Seek to the beginning
+    // Seek to the beginning for non-empty scenes
     if (
       (!state.loop && this.finished && !state.paused && state.seek < 0) ||
       this.endFrame === this.startFrame
     ) {
       this.togglePlayback(false);
       state.paused = true;
-      state.seek = this.startFrame;
+      state.seek =
+        this.endFrame === this.startFrame ? state.seek : this.startFrame;
     }
 
     // Seek to the beginning if looping is enabled
--- a/packages/core/src/app/Player.ts
+++ b/packages/core/src/app/Player.ts
@@ -338,12 +338,14 @@ export class Player {
     }
 
     // Pause if reached the end or the range is 0
+    // Seek to the beginning
     if (
       (!state.loop && this.finished && !state.paused && state.seek < 0) ||
       this.endFrame === this.startFrame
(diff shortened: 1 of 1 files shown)


packages/core/src/app/Player.ts
```diff
@@ -338,12 +338,14 @@ export class Player {
...
 
     // Pause if reached the end or the range is 0
+    // Seek to the beginning
     if (
       (!state.loop && this.finished && !state.paused && state.seek < 0) ||
       this.endFrame === this.startFrame
     ) {
       this.togglePlayback(false);
       state.paused = true;
+      state.seek = this.startFrame;
```
aarthificial, line 348: This will cause endless seeking when the end frame is the same as the start frame.
As is the case in an empty animation, for example.

aarthificial (changes requested):

packages/core/src/app/Player.ts
Vaibhavi-Sita: It would seem that it will cause endless seeking, but the player will remain paused once it seeks to the beginning in disable loop mode. Hence, unlike the loop situation- where the player still remains playing upon reaching the end, this will pause on reaching the start frame and will not seek further.

On unit testing for the case of an empty animation, it behaves the same way as before.  It does not seek further as it remains paused.

If this is a wrong interpretation, please let me know how to correct it

aarthificial: I'm not sure what unit test you're referring to, but if you edit `packages/template/src/scenes/example.tsx` to contain an empty scene with no animations:
```tsx

  // do nothing
});
```
...and then run `npm run template:dev` to test your changes, you will see a "seek time" message being logged over and over again. This means that the player loop continously performs seeking for no reason:
https://github.com/motion-canvas/motion-canvas/blob/main/packages/core/src/app/Player.ts#L378-L383

Vaibhavi-Sita: Updated based on your feedback. I've added a check in case of an empty scene before updating seek. 
The `seek time` is now being logged as expected.

aarthificial (approved):

aarthificial:
Thanks for the PR!

Issue #711, opened 2023-05-23
Playback Speed for Video Elements


PixelRifts:
**Description**
There is currently no way of changing the playback speed of a video element, which would be quite helpful to have.

**Proposed solution**
The playback speed can probably be a signal.

rmhartog:
I'd be happy to try and put together a PR.

PR #831, opened 2023-10-24
feat(2d): add playbackRate signal to Video component

rmhartog:
Added a `playbackRate` signal to the Video component that mirrors [`HTMLMediaElement.playbackRate`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/playbackRate).

Solves #711.

### DIFF
--- a/packages/2d/src/components/Video.ts
+++ b/packages/2d/src/components/Video.ts
@@ -2,7 +2,7 @@ import {BBox, SerializedVector2} from '@motion-canvas/core/lib/types';
 import {drawImage} from '../utils';
 import {computed, initial, signal} from '../decorators';
 import {useThread} from '@motion-canvas/core/lib/utils';
-import {PlaybackState} from '@motion-canvas/core';
+import {PlaybackState, useLogger} from '@motion-canvas/core';
 import {clamp} from '@motion-canvas/core/lib/tweening';
 import {Rect, RectProps} from './Rect';
 import {DesiredLength} from '../partials';
@@ -10,7 +10,9 @@ import {
   DependencyContext,
   SignalValue,
   SimpleSignal,
+  isReactive,
 } from '@motion-canvas/core/lib/signals';
+import reactivePlaybackRate from './__logs__/reactive-playback-rate.md';
 
 export interface VideoProps extends RectProps {
   /**
@@ -29,6 +31,10 @@ export interface VideoProps extends RectProps {
    * {@inheritDoc Video.loop}
    */
   loop?: SignalValue<boolean>;
+  /**
+   * {@inheritDoc Video.playbackRate}
+   */
+  playbackRate?: number;
   /**
    * The starting time for this video in seconds.
    */
@@ -87,6 +93,15 @@ export class Video extends Rect {
   @initial(false)
   @signal()
   public declare readonly loop: SimpleSignal<boolean, this>;
+
+  /**
+   * The rate at which the video plays, as multiples of the normal speed.
+   *
+   * @defaultValue 1
+   */
+  @initial(1)
+  @signal()
+  public declare readonly playbackRate: SimpleSignal<number, this>;
 
   @initial(0)
   @signal()
@@ -166,6 +181,8 @@ export class Video extends Rect {
     const video = this.video();
     const time = this.clampTime(this.time());
 
+    video.playbackRate = this.playbackRate();
+
     if (!video.paused) {
       video.pause();
     }
@@ -183,11 +200,15 @@ export class Video extends Rect {
   protected fastSeekedVideo(): HTMLVideoElement {
     const video = this.video();
     const time = this.clampTime(this.time());
+
+    video.playbackRate = this.playbackRate();
+
     if (this.lastTime === time) {
       return video;
     }
 
-    const playing = this.playing() && time < video.duration;
+    const playing =
+      this.playing() && time < video.duration && video.playbackRate > 0;
     if (playing) {
       if (video.paused) {
         DependencyContext.collectPromise(video.play());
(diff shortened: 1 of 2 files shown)


packages/2d/src/components/Video.ts
```diff
@@ -88,6 +92,15 @@ export class Video extends Rect {
   @signal()
   public declare readonly loop: SimpleSignal<boolean, this>;
 
+  /**
+   * The rate at which the video plays, as multiples of the normal speed.
+   *
+   * @defaultValue 1.0
+   */
+  @initial(1.0)
```
aarthificial, line 100:
```suggestion
   * @defaultValue 1
   */
  @initial(1)
```

aarthificial (changes requested):
Don't have a way to test it rn but shouldn't `fastSeekedVideo` set the `playbackRate` attribute on the actual video element too?
Seems like the current implementation would cause the video to constantly get out of sync with the playback.

rmhartog:
> Don't have a way to test it rn but shouldn't `fastSeekedVideo` set the `playbackRate` attribute on the actual video element too? Seems like the current implementation would cause the video to constantly get out of sync with the playback.

Whenever the `playbackRate` is updated, it is also set on the video element. I've tested based on your comment, but did not find a scenario in which they'd get out of sync. Happy to dive into it further, but could you help me understand this scenario a little bit better?

aarthificial:
Sorry, didn't notice you were using the HTML attribute to store the value.
In that case though, wouldn't the value reset whenever the source changes?
For example:
```tsx
import videoA from './videoA.mp4';
import videoB from './videoB.mp4';

// ...

const video = createRef<Video>();
view.add(<Video ref={video} src={videoA} playbackRate={2} />);

// this will reset the playback rate back to 1:
video().src(videoB);
```

aarthificial:
@rmhartog 
If it helps, you can have a custom setter and getter while still storing the value in the signal like so:
```ts
  protected getPlaybackRate() {
    // invoke the native getter:
    return this.playbackRate.context.getter();
  }

  protected setPlaybackRate(playbackRate: number) {
    // invoke the native setter:
    this.playbackRate.context.setter(playbackRate);
  
    // do some other things:
    const video = this.video();
    video.playbackRate = playbackRate;

    if (playbackRate === 0) {
      this.pause();
    } else {
      const time = useThread().time;
      const start = time();
      const offset = this.time();
      this.time(() => this.clampTime(offset + (time() - start) * playbackRate));
    }
  }
```

rmhartog:
Thanks @aarthificial, that was the missing piece! I've updated the code to imperatively update the time signal if the playbackRate changes and the video is playing, but the `HTMLVideoElement.playbackRate` is set reactively in the `seekedVideo` and `fastSeekedVideo` respectively.

packages/2d/src/components/Video.ts
```diff
@@ -263,11 +281,34 @@ export class Video extends Rect {
     }
   }
 
+  protected getPlaybackRate() {
+    return this.playbackRate.context.getter();
+  }
```
aarthificial, line 286: I mentioned this to give you a complete picture but it's already the default behavior
Conceptually, it's similar to calling the `super` method:
```ts
protected getPlaybackRate() {
  return super.getPlaybackRate();
}
```
So the entire thing can be omitted.

aarthificial (changes requested):

packages/2d/src/components/Video.ts
rmhartog: Yeah, that makes sense. It didn't work without a getter before, but that was obviously because I didn't call the default setter. Will get rid of this method.

rmhartog:
Removed the superfluous getter method, think I've covered all comments @aarthificial!

packages/2d/src/components/Video.ts
```diff
@@ -263,11 +281,30 @@ export class Video extends Rect {
     }
   }
 
+  protected setPlaybackRate(playbackRate: number) {
```
aarthificial, line 284: For the sake of completeness, the value passed here is not a number but a `SignalValue<number>`
As mentioned previously, this implementation won't work with functions so we should check for that here and log a warning:
```suggestion
  protected setPlaybackRate(playbackRate: SignalValue<number>) {
    let value: number;
    if (isReactive(playbackRate)) {
      value = playbackRate();
      useLogger().warn({
        message: 'Invalid value set as the playback rate',
        remarks: ...,
        inspect: this.key,
      });
    } else {
      value = playbackRate;
    }
```

aarthificial (changes requested):
One more thing, `setPlaybackRate` pauses the video when the value is `0` but this behavior is not reflected in `fastSeekVideo` so there's a possibility that the video won't get paused after the source is changed. It should be easy to fix by expanding this condition at line 208:
```ts
const playing = this.playing() && time < video.duration;
```

rmhartog:
Updated based on your feedback, last main question for me is the `remarks` of the log message. Not sure if it is necessary and/or how detailed it should be.

aarthificial:
I'd go for something like this:
````md
The `playbackRate` of a `Video` cannot be reactive.

Make sure to use a concrete value and not a function:

```ts
// wrong ✗
video.playbackRate(() => 7);
// correct ✓
video.playbackRate(7);
```

If you're using a signal, extract its value before passing it to the property:

```ts
// wrong ✗
video.playbackRate(mySignal);
// correct ✓
video.playbackRate(mySignal());
```
````

File: packages/2d/src/components/Video.ts [hunk 4]
```diff
@@ -263,11 +284,39 @@ export class Video extends Rect {
...
+    let value: number;
+    if (isReactive(playbackRate)) {
+      value = playbackRate();
+      useLogger().warn({
+        message: 'Invalid value set as the playback rate',
+        remarks: reactivePlaybackRate,
+        inspect: this.key,
```
aarthificial, line 294: Forgot about this but we should also add a stack trace here:
```suggestion
        inspect: this.key,
        stack: new Error().stack,
```

aarthificial (approved):

screaminc:
@rmhartog @aarthificial Is there any limits on playbackrate values, can we go more than 16x?

Issue #369, opened 2023-02-18
Changing the order of scenes during transition


aarthificial:
**Description**
Currently, the next scene is always rendered on top of the current one during transitions.
Certain effects require this order to be reversed or change mid-animation. 

**Proposed solution**
Extend `useTransition` with a third argument specifying the order:
```ts
useTransition(
  prevCtx => {},
  nextCtx => {},
  true, // previous on top
);
```
It should accept a `SignalValue` to make changing order mid-animation possible:
```ts
useTransition(
  prevCtx => {},
  nextCtx => {},
  () => scale.x() > 0.5,
);
```

monstajoe2002:
I would like to work on this issue please

rmhartog:
Interested in putting together a PR for this.

PR #832, opened 2023-10-24
feat(core): allow ordering of scenes during transition

rmhartog:
This PR adds a third argument to `useTransition`, which is a boolean or signal indicating whether the previous scene is rendered on top of the current scene during the transition.

Solves #369.

### DIFF
--- a/packages/core/src/scenes/GeneratorScene.ts
+++ b/packages/core/src/scenes/GeneratorScene.ts
@@ -23,7 +23,7 @@ import {Threadable} from './Threadable';
 import {BBox, Vector2} from '../types';
 import {SceneState} from './SceneState';
 import {Random} from './Random';
-import {DependencyContext} from '../signals';
+import {DependencyContext, SignalValue} from '../signals';
 import {SceneMetadata} from './SceneMetadata';
 import {Slides} from './Slides';
 
@@ -48,6 +48,7 @@ export abstract class GeneratorScene<T>
   public readonly variables: Variables;
   public random: Random;
   public creationStack?: string;
+  public previousOnTop: SignalValue<boolean>;
 
   public get firstFrame() {
     return this.cache.current.firstFrame;
@@ -134,6 +135,7 @@ export abstract class GeneratorScene<T>
     this.slides = new Slides(this);
 
     this.random = new Random(this.meta.seed.get());
+    this.previousOnTop = false;
   }
 
   public abstract getView(): T;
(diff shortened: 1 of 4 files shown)


packages/core/src/scenes/GeneratorScene.ts
```diff
@@ -48,6 +48,7 @@ export abstract class GeneratorScene<T>
   public readonly variables: Variables;
   public random: Random;
   public creationStack?: string;
+  public readonly belowPrevious: SimpleSignal<boolean, this>;
```
aarthificial, line 51: Scenes should not use signals directly as it may lead to memory leaks.
You should store the raw `SignalValue` instead, and use `unwrap` to access it whenever necessary.

packages/core/src/transitions/useTransition.ts
```diff
@@ -1,14 +1,17 @@
+import {SignalValue} from '../signals/types';
 import {useScene} from '../utils';
 
 /**
  * Transition to the current scene by altering the Context2D before scenes are rendered.
  *
  * @param current - The callback to use before the current scene is rendered.
  * @param previous - The callback to use before the previous scene is rendered.
+ * @param previousOnTop - A boolean or signal indicating whether the previous scene should be rendered on top of the current scene.
```
aarthificial, line 9: Lines should not exceed 80 characters

packages/core/src/scenes/Scene.ts
```diff
@@ -301,4 +302,9 @@ export interface Scene<T = unknown> {
    * Should always return `true`.
    */
   isCached(): boolean;
+
+  /**
+   * Should this scene be rendered below the previous scene during a transition?
+   */
+  belowPrevious: SimpleSignal<boolean, this>;
```
aarthificial, line 309: It's a bit confusing that the transition API refers to this concept as `previousOnTop` but the internal implementation uses the name `belowPrevious`. Couldn't they be the same?

aarthificial (changes requested):

packages/core/src/scenes/GeneratorScene.ts
aarthificial: It would also be a goof idea to reset this property in the `reset` method

rmhartog: Was not aware of that, good catch. Updated to store the raw `SignalValue`.

packages/core/src/transitions/useTransition.ts
rmhartog: Fixed, didn't realise as the linter did not complain. Would it be an idea to add a `max-len` rule?

packages/core/src/scenes/Scene.ts
rmhartog: Made the naming consistent as `previousOnTop` everywhere.

rmhartog:
@aarthificial I've updated the PR based on your comments. Could you take another look? 🙂

aarthificial (approved):

aarthificial:
Looks great, thanks!

PR #836, opened 2023-10-26
fix(2d): calculate Txt cache bbox from contents

rmhartog:
Override the `getCacheBBox` method in `Txt` to calculate the cache bounding box based on the actual contents.

In reference to https://github.com/motion-canvas/motion-canvas/pull/525#issuecomment-1468869328.

Closes: #465

### DIFF
--- a/packages/2d/src/components/Txt.ts
+++ b/packages/2d/src/components/Txt.ts
@@ -105,6 +105,24 @@ export class Txt extends Shape {
     context.restore();
   }
 
+  protected override getCacheBBox(): BBox {
+    const size = this.computedSize();
+    const range = document.createRange();
+    range.selectNodeContents(this.element);
+    const bbox = range.getBoundingClientRect();
+
+    const lineWidth = this.lineWidth();
+    // We take the default value of the miterLimit as 10.
+    const miterLimitCoefficient = this.lineJoin() === 'miter' ? 0.5 * 10 : 0.5;
+
+    return new BBox(
+      -size.width / 2,
+      -size.height / 2,
+      bbox.width,
+      bbox.height,
+    ).expand(lineWidth * miterLimitCoefficient);
+  }
+
   @computed()
   protected formattedText() {
     Txt.formatter.innerText = this.text();


packages/2d/src/components/Txt.ts
```diff
@@ -105,6 +105,20 @@ export class Txt extends Shape {
...
 
+  protected override getCacheBBox(): BBox {
+    const cache = super.getCacheBBox();
+
+    if (!this.element.hasChildNodes) return cache;
+
+    const size = this.computedSize();
+    const range = document.createRange();
+    range.setStartBefore(this.element.firstChild!);
+    range.setEndAfter(this.element.lastChild!);
```
aarthificial, line 116: Wouldn't this work?
```ts
range.selectNodeContents(this.element);
```

packages/2d/src/components/Txt.ts
```diff
@@ -105,6 +105,20 @@ export class Txt extends Shape {
...
+    const bbox = range.getBoundingClientRect();
+
+    return new BBox(-size.width / 2, -size.height / 2, bbox.width, bbox.height);
```
aarthificial, line 119: We should expand the signature of `BBox.fromSizeCentered` and use it here isntead:
```suggestion
    return BBox.fromSizeCentered(bbox);
```

aarthificial (changes requested):

aarthificial:
Also, it doesn't always work:
image
<details>
<summary>Repro</summary>

```tsx
import {makeScene2D, Txt} from '@motion-canvas/2d';
import {waitFor, createRef} from '@motion-canvas/core';

  const text = createRef<Txt>();

  view.add(
    <Txt
      fontSize={200}
      ref={text}
      height={160}
      width={600}
      fill={'white'}
      opacity={0.5}
    >
      Example
    </Txt>,
  );

  yield* waitFor(5);
});
```

</details>

packages/2d/src/components/Txt.ts
aarthificial: Nvm, just noticed it uses the other size in the first few arguments.

aarthificial:
We should also account for the line width:

image
<details>
<summary>Repro</summary>

```tsx
import {makeScene2D, Txt} from '@motion-canvas/2d';
import {waitFor, createRef} from '@motion-canvas/core';

  const text = createRef<Txt>();

  view.add(
    <Txt
      fontSize={200}
      ref={text}
      height={160}
      width={600}
      stroke={'white'}
      fill={'black'}
      strokeFirst
      lineWidth={96}
      opacity={0.5}
    >
      Example
    </Txt>,
  );

  yield* waitFor(5);
});
```

</details>

Here's how curves handle this:
```ts
const coefficient = this.lineCap() === 'square' ? 0.5 * 1.4143 : 0.5;
return bbox.expand(this.lineWidth() * coefficient);
```

packages/2d/src/components/Txt.ts
rmhartog: I'll try it again, but I recall when I tested that the width of the div changed the result.

rmhartog:
> We should also account for the line width:

Totally forgot about strokes, will dive into it.

rmhartog:
After some tweaking, I found that most likely the `miterLimit` is indeed the worst case scenario. However, we can only determine the current `miterLimit` when drawing on the canvas. The other option would be to add a `miterLimit` signal to `Shape`. Curious to know what you think about that, @aarthificial.

aarthificial:
The default value of `miterLimit` [is defined to be `10`](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/miterLimit#value) so you could just use that.

If you want to add a `miterLimit` signal feel free to do so but in a separate PR please.

packages/2d/src/components/Txt.ts
```diff
@@ -105,6 +105,28 @@ export class Txt extends Shape {
     context.restore();
   }
 
+  protected override getCacheBBox(): BBox {
+    const cache = super.getCacheBBox();
+
+    if (!this.element.hasChildNodes) return cache;
```
aarthificial, line 111: One final thing, these lines shouldn't be necessary now because we no longer use children for selection.
Additionally, `hasChildNodes` is a function, not a property, so right now this if statement is always false anyway.

aarthificial (changes requested):

packages/2d/src/components/Txt.ts
```diff
@@ -105,6 +105,28 @@ export class Txt extends Shape {
...
+
+    const size = this.computedSize();
+    const range = document.createRange();
+    range.selectNodeContents(this.element);
+    const bbox = range.getBoundingClientRect();
+
+    const lineWidth = this.lineWidth();
+    // We take the default value of the miterLimit as 10.
+    const miterLimitCoefficient = 0.5 * 10;
```
aarthificial, line 120: Also, this coefficient should only be used if `this.lineJoin()` is set to `miter`.
For both `bevel` and `round`, the coefficient can be just `0.5`

rmhartog:
Apologies, committed some code that was not fully completed. Thanks for your patience and suggestions.

aarthificial (approved):

aarthificial:
Thank you too!

PR #839, opened 2023-11-11
docs: add step by step local development setup

TungVietLe:
The step-by-step guide was quite hard for me to follow personally. Especially, I was struggling when picking development packages because they only mentioned `template:build` rather than `npm run template:build`, etc...


aarthificial:
Closing this in favor of #867

PR #852, opened 2023-11-21
feat(2d): add querying helpers

aarthificial:
Introduces multiple helpers for querying the node hierarchy and dealing with type casting.

### `Node.findAll()`
Finds all descendants of the node that match the given predicate:
```ts
const texts = node.findAll<Txt>(child => child instanceof Txt);
```

### `Node.findFirst()`
Finds the first descendant that matches.
Uses depth-first traversal, like `querySelector` in DOM.

### `Node.findLast()`
Finds the last descendant that matches.
Order is the same as in `findFirst` but in reverse.

### `Node.findAncestor()`
Finds the first ancestor that matches.

---

### `is()`
A helper function for creating `instanceof` predicates.
It's less verbose than manually writing `instanceof` and, thanks to type guards, TypeScript can infer the type of the resulting array automatically:
```ts
const texts = node.find(is(Txt));
```

---

### `Node.[child/children/parent]As()`
A set of helpers for easier casting. They remove the need for a `(...)` wrapper that stems from using `as`.
Solely for convenience, they don't actually use `instanceof` to type-check anything:
```ts
// old way:
(node.children()[2] as Txt).fill('white');
// new way
node.childAs<Txt>(2).fill('white');
```
```ts
// old way:
yield* all(...node.children().map(child => (child as Txt).fill('white', 2)));
// new way
yield* all(...node.childrenAs<Txt>().map(child => child.fill('white', 2)));
```
```ts
// old way:
(node.parent() as Txt).fill('white');
// new way
(node.parentAs<Txt>().fill('white');
```

### DIFF
--- a/packages/2d/src/components/Node.ts
+++ b/packages/2d/src/components/Node.ts
@@ -909,6 +909,163 @@ export class Node implements Promisable<Node> {
   }
 
   /**
+   * Find all descendants of this node that match the given predicate.
+   *
+   * @param predicate - A function that returns true if the node matches.
+   */
+  public find<T extends Node>(predicate: (node: any) => node is T): T[];
+  /**
+   * Find all descendants of this node that match the given predicate.
+   *
+   * @param predicate - A function that returns true if the node matches.
+   */
+  public find<T extends Node = Node>(predicate: (node: any) => boolean): T[];
+  public find<T extends Node>(predicate: (node: any) => node is T): T[] {
+    const result: T[] = [];
+    const queue = this.reversedChildren();
+    while (queue.length > 0) {
+      const node = queue.pop()!;
+      if (predicate(node)) {
+        result.push(node);
+      }
+      const children = node.children();
+      for (let i = children.length - 1; i >= 0; i--) {
+        queue.push(children[i]);
+      }
+    }
+
+    return result;
+  }
+
+  /**
+   * Find the first descendant of this node that matches the given predicate.
+   *
+   * @param predicate - A function that returns true if the node matches.
+   */
+  public findFirst<T extends Node>(
+    predicate: (node: Node) => node is T,
+  ): T | null;
+  /**
+   * Find the first descendant of this node that matches the given predicate.
+   *
+   * @param predicate - A function that returns true if the node matches.
+   */
+  public findFirst<T extends Node = Node>(
+    predicate: (node: Node) => boolean,
+  ): T | null;
+  public findFirst<T extends Node>(
+    predicate: (node: Node) => node is T,
+  ): T | null {
+    const queue = this.reversedChildren();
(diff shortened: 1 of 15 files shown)


packages/2d/src/components/Node.ts
```diff
@@ -908,6 +908,163 @@ export class Node implements Promisable<Node> {
...
+  /**
+   * Find the first ancestor of this node that matches the given predicate.
+   *
+   * @param predicate - A function that returns true if the node matches.
+   */
+  public findAncestor<T extends Node>(
+    predicate: (node: Node) => node is T,
+  ): T | null;
+  /**
+   * Find the first ancestor of this node that matches the given predicate.
```
Ross-Esmond, line 1024: It could say "closest" ancestor, rather than first. I didn't immediately know if "first" means highest or lowest.

packages/2d/src/components/Node.ts
```diff
@@ -1453,6 +1610,15 @@ export class Node implements Promisable<Node> {
...
   }
+
+  private reversedChildren() {
+    const children = this.children();
+    const result: Node[] = [];
+    for (let i = children.length - 1; i >= 0; i--) {
+      result.push(children[i]);
+    }
+    return result;
+  }
```
Ross-Esmond, line 1621: This one could be handled natively. JS has an `array.reverse()` already which is heavily supported. You could use that instead or just let the users call it themselves.

```js
node.children().reverse();
```

packages/2d/src/components/Node.ts
```diff
@@ -908,6 +908,163 @@ export class Node implements Promisable<Node> {
...
+   * @param predicate - A function that returns true if the node matches.
+   */
+  public find<T extends Node>(predicate: (node: any) => node is T): T[];
+  /**
+   * Find all descendants of this node that match the given predicate.
+   *
+   * @param predicate - A function that returns true if the node matches.
+   */
+  public find<T extends Node = Node>(predicate: (node: any) => boolean): T[];
+  public find<T extends Node>(predicate: (node: any) => node is T): T[] {
```
Ross-Esmond, line 923: `find` on the JS array returns one element—the first one. Shouldn't this be `filter`.

packages/2d/src/components/Node.ts
aarthificial: `reverse` modifies the array in place. It would change the cached array in the signal.
I could do `[...node.children()].reverse()` but that's double the iteration for no reason.

Eventually, we'll change it to `toReversed` but this requires ES2023 and typedoc is holding us back at the moment.

packages/2d/src/components/Node.ts
aarthificial: Changed it to `findAll` in relation to the discord discussion.

PR #855, opened 2023-11-23
feat(core): add static properties to Vector2 corresponding to Origins

### LINKED ISSUE #844: Add Vector constants for all origins
**Description**
The [`Vector2`](https://github.com/motion-canvas/motion-canvas/blob/main/packages/core/src/types/Vector.ts) class should be extended by new static properties with vectors representing all [`Origin`s](https://github.com/motion-canvas/motion-canvas/blob/main/packages/core/src/types/Origin.ts).

For example:
```ts
public static readonly top = new Vector2(0, -1);
```
The value of each constant should be equal to the vector returned by `Vector2.fromOrigin` for the given origin:
```ts
Vector2.fromOrigin(Origin.Top); // Vector2(0, -1)
```

CactusPuppy:
This PR adds static properties﻿ corresponding to the all currently available Origins.

Resolves #844.

### DIFF
--- a/packages/core/src/types/Vector.ts
+++ b/packages/core/src/types/Vector.ts
@@ -47,6 +47,14 @@ export class Vector2 implements Type {
   public static readonly up = new Vector2(0, 1);
   public static readonly down = new Vector2(0, -1);
 
+  public static readonly middle = new Vector2();
+  public static readonly top = new Vector2(0, -1);
+  public static readonly bottom = new Vector2(0, 1);
+  public static readonly topLeft = new Vector2(-1, -1);
+  public static readonly topRight = new Vector2(1, -1);
+  public static readonly bottomLeft = new Vector2(-1, 1);
+  public static readonly bottomRight = new Vector2(1, 1);
+
   public x = 0;
   public y = 0;
 
--- a/packages/core/src/types/Vector.ts
+++ b/packages/core/src/types/Vector.ts
@@ -47,12 +47,29 @@ export class Vector2 implements Type {
   public static readonly up = new Vector2(0, 1);
   public static readonly down = new Vector2(0, -1);
 
-  public static readonly middle = new Vector2();
+  /**
+   * A constant equal to `Vector2(0, -1)`
+   */
   public static readonly top = new Vector2(0, -1);
+  /**
+   * A constant equal to `Vector2(0, 1)`
+   */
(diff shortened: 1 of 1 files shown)


packages/core/src/types/Vector.ts
```diff
@@ -47,6 +47,14 @@ export class Vector2 implements Type {
   public static readonly up = new Vector2(0, 1);
   public static readonly down = new Vector2(0, -1);
```
CactusPuppy, line 48: These properties appear to be incorrect, but since they belong to a previous commit, and changing them might break existing projects, I elected to leave them alone and flag them for review.

From my understanding of Motion Canvas, the following changes should be made:
```suggestion
  public static readonly up = new Vector2(0, -1);
  public static readonly down = new Vector2(0, 1);
```

aarthificial: They are "correct" in that they're aligned with a mathematical expectation for the `up` and `down` vectors.
Regardless, we cannot change them right now as that would be considered a breaking change. This would require raising the version to `4.x.x` which we are not ready for just yet.

<details>
<summary>Here's a <a href="https://discord.com/channels/1071029581009657896/1175315979174162453">Discord conversation</a> with @hhenrichsen for more detail</summary>

image

</details>

packages/core/src/types/Vector.ts
```diff
@@ -47,6 +47,14 @@ export class Vector2 implements Type {
...
 
+  public static readonly middle = new Vector2();
```
aarthificial, line 50: I think this once can be skipped as it's the same as `Vector2.zero`

packages/core/src/types/Vector.ts
```diff
@@ -47,6 +47,14 @@ export class Vector2 implements Type {
...
+  public static readonly top = new Vector2(0, -1);
```
aarthificial, line 51: Could you also add some TSDoc containing the value?
Something like:
```suggestion
  /**
   * A constant for `Vector2(0, -1)`.
   */
  public static readonly top = new Vector2(0, -1);
```
This would make it easier to quickly peak the value in an IDE.

aarthificial (changes requested):

aarthificial (approved):

aarthificial:
Thanks for the PR!

Issue #320, opened 2023-02-11
Add some form of volume control for the audio playback.


Aurora2500:
**Description**
I've noticed that there is no audio volume control for the playback at the moment.
Which can be annoying if the audio file itself is too loud for your liking. 

**Proposed solution**
The default editor itself would need to have some audio volume slider to it that can allow the user to control the playback volume.

**Considered alternatives**
The browser itself doesn't really support controlling audio per tab, so you'd either need to edit the audio track itself in a third party to make it more quieter, or resort to lowering the system-wide volume, which can be annoying if you have it set perfectly for everything else that you use & just want the motion canvas audio playback to be quieter.

PR #856, opened 2023-11-28
feat(core & ui): add audio volume control through arrow keys

Vija02:
Adds the ability to control the volume of audio through arrow keys (up and down). This will update the volume by 10% increments.

Solves #320 partially.

There's no visible UI element yet since I think some discussion on how that should look like before implementing would be good. 
However, this will at least give people the option to control the volume if they know the button to press.

### DIFF
--- a/packages/core/src/app/Player.ts
+++ b/packages/core/src/app/Player.ts
@@ -18,6 +18,7 @@ export interface PlayerState extends Record<string, unknown> {
   paused: boolean;
   loop: boolean;
   muted: boolean;
+  volume: number;
   speed: number;
 }
 
@@ -127,6 +128,7 @@ export class Player {
     this.playerState = new ValueDispatcher<PlayerState>({
       loop: true,
       muted: true,
+      volume: 1,
       speed: 1,
       ...initialState,
       paused: true,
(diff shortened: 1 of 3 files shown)


aarthificial:
I'd imagine the volume slider to appear above the sound icon when hovering it:
image

Let me know if you'd be interested in implementing it in the context of this PR.

packages/core/src/app/Player.ts
```diff
@@ -278,6 +280,25 @@ export class Player {
     }
   }
 
+  public setAudioVolume(value: number): void {
+    if (value !== this.playerState.current.volume) {
```
aarthificial, line 284: The value should be clamped in this method as well.

packages/core/src/app/Player.ts
```diff
@@ -278,6 +280,25 @@ export class Player {
...
+  }
+
+  public addAudioVolume(value: number): void {
+    const newValue = clamp(0, 1, this.playerState.current.volume + value);
+    if (newValue !== this.playerState.current.volume) {
+      this.playerState.current = {
+        ...this.playerState.current,
+        volume: newValue,
+      };
+    }
```
aarthificial, line 299: Once `setAudioVolume` clamps the value, this could be written as simply:
```suggestion
    this.setAudioVolume(this.playerState.current.volume + value);
```

aarthificial (changes requested):

Vija02:
Updated your changes. 

UI: Hmm, the only thing stopping me would be implementing the popover component which doesn't seem to exist in the project yet. Maybe we can use something like [Popper](https://popper.js.org/)? 

Let's get this PR merged and I'll make a separate PR if I get around to it.

aarthificial:
Thanks for the PR!

I think Popper would be a bit of an overkill for this. 
This element is more like a dropdown and since overflow is not disabled in this control bar, you should be able to easily position the slider with `position: absolute`:

image

jmaen:
Are you currently working on this @Vija02 ?

Vija02:
@jmaen No I'm not

PR #867, opened 2023-12-11
docs: add instructions for local installation

aarthificial:


Gabryxx7:
I see my hard work has been eventually used for good in the README, coincidentally right after I posted some issues linking at my own fork with these exact changes!

aarthificial:
@Gabryxx7 Can you link me to those exact changes of yours because I genuinely don't know what you're referring to?
IIRC correctly, I wrote this installation guide myself.

PR #869, opened 2023-12-18
Update Features.tsx to use current API

colinta:
For some reason this `export makeScene()` bothered me – is that valid JS!? (no I don't think it is),
and in fixing it I realized that also the function name is incorrect... which is probably just an API change from long ago, I'm guessing?


aarthificial:
Thanks for the fix!

PR #872, opened 2023-12-19
feat(ui): add volume slider

### LINKED ISSUE #320: Add some form of volume control for the audio playback.

jmaen:
Add a volume slider to adjust audio volume in addition to the existing key controls. The slider appears when hovering the sound icon, as described in #856

I had some trouble getting the default `<input type="range"/>` to look consistent accross different browsers, especially with the vertical orientation, so I build a slider component from scratch.

Closes #320

### DIFF
--- a/packages/ui/src/components/controls/Controls.module.scss
+++ b/packages/ui/src/components/controls/Controls.module.scss
@@ -286,3 +286,36 @@
     }
   }
 }
+
+.slider {
+  display: flex;
+  justify-content: center;
+  height: 120px;
+  width: 24px;
+  cursor: pointer;
+}
+
+.sliderTrack {
+  position: relative;
+  width: 4px;
+  border-radius: var(--radius);
+  background-color: var(--background-color-dark);
+}
+
+.sliderProgress {
+  position: absolute;
+  bottom: 0;
+  width: 4px;
+  border-radius: var(--radius);
+  background-color: var(--theme);
+}
+
+.sliderThumb {
+  position: absolute;
+  left: -4px;
+  bottom: calc(100% - 2px);
+  width: 12px;
+  height: 4px;
+  border-radius: var(--radius);
+  background-color: var(--theme);
+}
--- a/packages/ui/src/components/controls/Controls.module.scss
+++ b/packages/ui/src/components/controls/Controls.module.scss
@@ -297,25 +297,24 @@
 
 .sliderTrack {
   position: relative;
-  width: 4px;
+  display: flex;
+  align-items: end;
+  width: 2px;
   border-radius: var(--radius);
   background-color: var(--background-color-dark);
 }
 
 .sliderProgress {
-  position: absolute;
-  bottom: 0;
-  width: 4px;
+  width: 2px;
   border-radius: var(--radius);
   background-color: var(--theme);
 }
 
 .sliderThumb {
-  position: absolute;
-  left: -4px;
-  bottom: calc(100% - 2px);
-  width: 12px;
-  height: 4px;
-  border-radius: var(--radius);
+  width: 8px;
+  height: 8px;
+  margin-left: -3px;
+  margin-top: -4px;
+  border-radius: 4px;
   background-color: var(--theme);
 }
--- a/packages/ui/src/components/controls/Slider.tsx
+++ b/packages/ui/src/components/controls/Slider.tsx
@@ -1,0 +1,63 @@
+import {useEffect, useState} from 'preact/hooks';
+import {MouseButton, clamp} from '../../utils';
+import styles from './Controls.module.scss';
+
+export interface SliderProps {
+  value?: number;
+  onChange?: (value: number) => void;
+}
+
+export function Slider({value, onChange}: SliderProps) {
+  const [internalValue, setInternalValue] = useState(value);
+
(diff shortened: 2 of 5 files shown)


packages/ui/src/components/playback/Playback.module.scss
```diff
@@ -18,3 +18,33 @@
...
+  align-items: center;
+  width: 24px;
+  height: 144px;
+  border-radius: var(--radius);
+  background-color: var(--surface-color);
+}
+
+.volumeTrigger:hover .volumeMargin {
+  display: grid;
+}
```
aarthificial, line 50: Many hardcoded sizes here could be removed to let the container expand to whatever size the slider wants.
I'd also stick to the `24px` width. I'm assuming that these `48px` are meant to make the hover hitbox bigger but I'm not sure if that's necessary.
```suggestion
.volumeMargin {
  position: absolute;
  bottom: 24px;
  left: 0;
  right: 0;
  padding-bottom: 18px;
  display: none;
}

.volume {
  padding: 12px 0;
  border-radius: var(--radius);
  background-color: var(--surface-color);
}

.volumeTrigger:hover .volumeMargin {
  display: block;
}
```

packages/ui/src/components/controls/Controls.module.scss
```diff
@@ -286,3 +286,36 @@
...
+
+.sliderThumb {
+  position: absolute;
+  left: -4px;
+  bottom: calc(100% - 2px);
+  width: 12px;
+  height: 4px;
+  border-radius: var(--radius);
+  background-color: var(--theme);
+}
```
aarthificial, line 321: I adjusted the styles to match the mockup.
Mainly the width of the track should be `2px` and the handle should be an `8px` circle.
I'd also get rid of the `absolute` positioning with `calc` on the thumb. A child element will be naturally placed in the top left corner of its parent so simply nudging its position a bit to the top left will place it in the correct place:
```suggestion
.slider {
  display: flex;
  justify-content: center;
  height: 120px;
  width: 24px;
  cursor: pointer;
}

.sliderTrack {
  display: flex;
  align-items: end;
  position: relative;
  width: 2px;
  border-radius: var(--radius);
  background-color: var(--background-color-dark);
}

.sliderProgress {
  width: 2px;
  border-radius: var(--radius);
  background-color: var(--theme);
}

.sliderThumb {
  width: 8px;
  height: 8px;
  border-radius: 4px;
  margin-left: -3px;
  margin-top: -4px;
  background-color: var(--theme);
}
```

packages/ui/src/components/controls/Slider.tsx
```diff
@@ -0,0 +1,63 @@
...
+        if (event.button === MouseButton.Left) {
+          event.preventDefault();
+          event.stopPropagation();
+          event.currentTarget.setPointerCapture(event.pointerId);
+
+          const rect = event.currentTarget.getBoundingClientRect();
+          // +2 & +4 to account for thumb offset
+          const y = event.clientY - rect.y + 2;
+          const newInternalValue = 1 - y / (rect.height + 4);
+          setInternalValue(clamp(0, 1, newInternalValue));
```
aarthificial, line 30: The thumb offset is not necessary because the thumb should be able to overflow the slider in extreme positions:
image

packages/ui/src/components/playback/PlaybackControls.tsx
```diff
@@ -84,14 +84,32 @@ export function PlaybackControls() {
...
+            <div className={styles.volume}>
+              <Slider
+                value={state.volume}
+                onChange={volume => {
+                  if (isNaN(volume)) {
+                    volume = 0;
+                  }
+                  player.setAudioVolume(volume);
+                }}
+              ></Slider>
```
aarthificial, line 108:
```suggestion
              />
```

aarthificial (changes requested):
Looks good!
There are just a few tweaks to the styles necessary

packages/ui/src/components/controls/Slider.tsx
jmaen: The idea is that the thumb itself is clickable as well. So for a track of length `120px`, the clickable range would actually be `-4px` - `124px`, relative to the track (for the case of a `8px` wide circle).

That's why I add the offset and then divide by `height + circle width`.

But looking at it now, clamping `y` to stay between `0` and `height` might be a better solution.

jmaen:
I adressed your styling improvements and updated the value calculation in the `Slider` to use `clamp` instead of the offset. I think that also improves readability and it's not dependent on the thumb size anymore.

aarthificial (approved):

aarthificial:
Awesome, thanks for the PR!

PR #876, opened 2023-12-20
feat: add experimental features

aarthificial:
Adds a project setting for enabling experimental features:
```ts
// project.ts

export default makeProject({
  experimentalFeatures: true,
  // ...
});
```
The setting can be accessed in the UI:
```ts
const {project} = useApplication();
console.log(project.experimentalFeatures);
```
As well as in the runtime code:
```ts
const scene = useScene();
console.log(scene.experimentalFeatures);
```

Once enabled, the UI displays the following indication:
image


packages/docs/src/components/Api/Comment/index.tsx
```diff
@@ -33,16 +33,7 @@ function Experimental({comment}: {comment: JSONOutput.Comment}) {
...
-    <>
-      {experimental && (
-        <Admonition type="experimental">
-          This is an experimental feature. The behavior and API may change
-          drastically between major releases.
-        </Admonition>
-      )}
-    </>
-  );
+  return <>{experimental && <ExperimentalWarning />}</>;
```
hhenrichsen, line 36: (Nit) From what I understand of JSX, it's good practice to write this as:
```
{experimental ? <ExperimentalWarning /> : null}
```
This prevents `experimental` leaking into the DOM in certain conditions.

packages/template/src/project.ts
```diff
@@ -3,5 +3,6 @@ import {makeProject} from '@motion-canvas/core';
 import example from './scenes/example?scene';
 
 export default makeProject({
+  experimentalFeatures: true,
```
hhenrichsen, line 6: Is this intended to be set to true in the template project?

hhenrichsen (approved):
Couple minor questions, no blockers. Excited to be able to use this for fun new APIs!

packages/docs/src/components/Api/Comment/index.tsx
aarthificial: Changed it and removed the wrapping fragment altogether.

packages/template/src/project.ts
aarthificial: I think so. This is an internal package meant for maintainers to test new features without having to set up npm links etc. so I think having them enabled is a better default

PR #899, opened 2024-01-01
fix(2d): supress src-Warnings in Icon- and Latex-Component when correctly using component

WaldemarLehner:
This is a bugfix that prevents wrong / confusing Warn-Messages for the Icon Component.

The Icon component does not set the `src` Signal explicitly. It overrides the `src`'s getter by defining a  `getSrc` function. 

Because it doesn't set the Signal, a Warning is emitted by the parent `Img`-Class. 

This PR explicitly sets the `src` to `null` to supress the warning.

The `Icon` component also overrides the `setSrc` to tell the User that it is not supported to explicitly set the src.

In the PR, this Warning Message gets supressed until the component is fully initialized. Otherwise setting the `src` to 'null` in the constructor would also trigger that Warning.

Edit: [Additional context from Discord](https://discord.com/channels/1071029581009657896/1073222950028845138/1191273234235134144)

Edit2: Scope has been widened to also fix the `Latex` component. [See Discord](https://discord.com/channels/1071029581009657896/1071067015881707670/1191937738639949985)

### DIFF
--- a/packages/2d/src/lib/components/Icon.ts
+++ b/packages/2d/src/lib/components/Icon.ts
@@ -54,17 +54,11 @@ export class Icon extends Img {
   @colorSignal()
   public declare color: ColorSignal<this>;
 
-  /**
-   * The init check is only used to supress an error warning about setting the `src` Property.
-   */
-  private init = false;
-
   public constructor(props: IconProps) {
     super({
       ...props,
-      src: null!,
+      src: null,
     });
-    this.init = true;
   }
 
   /**
@@ -90,8 +84,8 @@ export class Icon extends Img {
    * overrides `Image.src` setter to warn the user that the value
    * is not used
    */
-  protected setSrc() {
-    if (!this.init) {
+  protected setSrc(src: string | null) {
+    if (src === null) {
       return;
     }
     useLogger().warn(
--- a/packages/2d/src/lib/components/Icon.ts
+++ b/packages/2d/src/lib/components/Icon.ts
@@ -54,8 +54,17 @@ export class Icon extends Img {
   @colorSignal()
   public declare color: ColorSignal<this>;
 
+  /**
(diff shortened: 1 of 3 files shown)


packages/2d/src/lib/components/Icon.ts
```diff
@@ -54,8 +54,17 @@ export class Icon extends Img {
...
+  /**
+   * The init check is only used to supress an error warning about setting the `src` Property.
+   */
+  private init = false;
+
   public constructor(props: IconProps) {
-    super(props);
+    super({
+      ...props,
+      src: null!,
```
hhenrichsen, line 65: Anytime I see `!` bypassing the type system, I wonder if there's a different way. Maybe we check src where it's used and emit the error there, instead of making initialization / error checking more complex (and bypassing type checking at the same time).

Maybe the ideal is we set up an overridable function in `Node` called checkProps that is called after props are set and can be overridden to add or remove the error message about `src`.

WaldemarLehner: In this specific case it might make sense to allow `src` to accept `null`, as the warning message indicates that one can do so:

image

hhenrichsen: I think that would be a fine change as well, but should be one that updates the allowed types as well instead of bypassing them 🙂

packages/2d/src/lib/components/Icon.ts
```diff
@@ -82,6 +91,9 @@ export class Icon extends Img {
    * is not used
    */
   protected setSrc() {
+    if (!this.init) {
```
aarthificial, line 94: I think simply checking if the newly set `src` is `null` should be enough

aarthificial (changes requested):

hhenrichsen (approved):

aarthificial (approved):

PR #907, opened 2024-01-05
feat(ui): add direct range selection & playhead shortcuts

### LINKED ISSUE #875: Directly select timeline range
**Description**
Since the timeline scrubbing PR #862 it's possible to move the selected range and edit it using the two handles while pressing `SHIFT`. 
I think it would be nice to be able do directly select the range (start handle is positioned on `mousedown` and end handle on `mouseup`). That would be especially useful when zoomed in on a specific part in the timeline, as then the rangle handles will often times be off-screen.

**Proposed solution**
Using another key, e.g. `ALT` or any other key combination (maybe using `RMB` might work as well?), that will allow to edit the range by directly "drawing it in". I suppose most of the (dragging) logic should already be there in the `RangeSelector.tsx`.

I'd imagine something like this:

range selection

jmaen:
Add the possibility to edit the range directly by "drawing it in" when `SHIFT` and `CTRL` are pressed. To also enable this feature outside of the current range, an additional `rangeContainer` is added as well.

Additionally, pressing `B`/`N` now moves the start/end of the range to the current playhead position. These shortcuts are also added to the footer.

Closes #875

### DIFF
--- a/packages/ui/src/components/timeline/RangeSelector.tsx
+++ b/packages/ui/src/components/timeline/RangeSelector.tsx
@@ -9,6 +9,7 @@ import {
   useKeyHold,
   usePreviewSettings,
   useSharedSettings,
+  useSize,
 } from '../../hooks';
 import {labelClipDraggingLeftSignal} from '../../signals';
 import {MouseButton} from '../../utils';
@@ -29,7 +30,9 @@ export function RangeSelector({rangeRef}: RangeSelectorProps) {
   const endFrame = Math.min(player.status.secondsToFrames(range[1]), duration);
   const [start, setStart] = useState(startFrame);
   const [end, setEnd] = useState(endFrame);
+  const rect = useSize(rangeRef);
   const shiftHeld = useKeyHold('Shift');
+  const controlHeld = useKeyHold('Control');
 
   const onDrop = useCallback(() => {
     labelClipDraggingLeftSignal.value = null;
@@ -51,24 +54,22 @@ export function RangeSelector({rangeRef}: RangeSelectorProps) {
   return (
     <div
       ref={rangeRef}
-      style={{
-        flexDirection: start > end ? 'row-reverse' : 'row',
-        left: `${framesToPercents(Math.ceil(Math.max(0, normalizedStart)))}%`,
-        right: `${
-          100 - framesToPercents(Math.ceil(Math.min(duration, normalizedEnd)))
-        }%`,
-      }}
-      className={clsx(styles.range, shiftHeld && styles.active)}
+      className={clsx(
+        styles.rangeContainer,
+        shiftHeld && controlHeld && styles.active,
+      )}
       onPointerDown={event => {
         if (event.button === MouseButton.Left) {
           event.preventDefault();
           event.stopPropagation();
           event.currentTarget.setPointerCapture(event.pointerId);
+
+          setStart(pixelsToFrames(event.clientX - rect.x));
+          setEnd(pixelsToFrames(event.clientX - rect.x));
         }
       }}
       onPointerMove={event => {
         if (event.currentTarget.hasPointerCapture(event.pointerId)) {
-          setStart(start + pixelsToFrames(event.movementX));
           setEnd(end + pixelsToFrames(event.movementX));
         }
       }}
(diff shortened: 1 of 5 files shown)


jmaen:
Here's a quick demo of what that looks like:

range-selection

packages/ui/src/components/timeline/RangeSelector.tsx
```diff
@@ -51,24 +54,22 @@ export function RangeSelector({rangeRef}: RangeSelectorProps) {
...
+        shiftHeld && controlHeld && styles.active,
+      )}
       onPointerDown={event => {
         if (event.button === MouseButton.Left) {
           event.preventDefault();
           event.stopPropagation();
           event.currentTarget.setPointerCapture(event.pointerId);
+
+          setStart(pixelsToFrames(event.clientX - rect.x));
+          setEnd(pixelsToFrames(event.clientX - rect.x));
```
aarthificial, line 68: Mapping from the pointer's position to the corresponding animation frame is common logic between the playhead and this. I'd suggest we add a function for this to `TimelineState`. This could be done in `Timeline.tsx` when the state is created:
```ts
const startPosition = sizes.paddingLeft + rect.x - offset;
const pointerToFrames: (value: number) =>
  conversion.pixelsToFrames(value - startPosition),
```
And then used both here and in the playhead scrubbing function.
It would also remove the need for another `useSize` in this component.

packages/ui/src/components/timeline/RangeSelector.tsx
```diff
@@ -78,13 +79,46 @@ export function RangeSelector({rangeRef}: RangeSelectorProps) {
...
+        style={{
+          flexDirection: start > end ? 'row-reverse' : 'row',
+          left: `${framesToPercents(Math.ceil(Math.max(0, normalizedStart)))}%`,
+          right: `${
+            100 - framesToPercents(Math.ceil(Math.min(duration, normalizedEnd)))
+          }%`,
+        }}
+        className={clsx(
+          styles.range,
+          shiftHeld && !controlHeld && styles.active,
```
aarthificial, line 93: When holding `CTRL` + `SHIFT` the range should still be highlighted with the handles remaining invisible. Right now there's no indication that anything can be done with the range.

packages/ui/src/components/timeline/Timeline.module.scss
```diff
@@ -283,6 +275,15 @@
   }
 }
 
+.rangeContainer {
```
aarthificial, line 278: To remain consistent with other components, this should be called `rangeTrack`
```suggestion
.rangeTrack {
```
image

packages/ui/src/components/timeline/Timeline.module.scss
```diff
@@ -283,6 +275,15 @@
...
+  height: 32px;
+  pointer-events: none;
```
aarthificial, line 280: The track itself should have a `pointer` cursor to signify that it can be edited:
```suggestion
  pointer-events: none;
  cursor: pointer;
```

aarthificial (changes requested):

packages/ui/src/components/timeline/RangeSelector.tsx
jmaen: Perhaps we can find a way to highlight the whole track because all of it can be edited, not just the current range. I think this might be a little more intuitive, don't you?

packages/ui/src/components/timeline/Timeline.module.scss
jmaen: I agree, but I think the cursor should be changed only when the `rangeTrack` is active (i.e. when `SHIFT` and `CTRL` are pressed)

packages/ui/src/components/timeline/RangeSelector.tsx
jmaen: Because as of right now, you can also draw the new range "through" the current range, which I think is convenient. But highlighting the range and then manipulating the underlying component feels even more confusing to me then giving no indication (apart from the changing cursor, but that is obviously only visible when hovering the range track..). I'm not quite sure what would be the best solution here

packages/ui/src/components/timeline/Timeline.module.scss
aarthificial: Exactly, the cursor won't be changed when `pointer-events` is set to `none`

jmaen: Oh yeah of course youre right, didn't think of that. My bad :)

packages/ui/src/components/timeline/RangeSelector.tsx
aarthificial: I phrased it poorly. What I meant is that by default both the range and the audio track are greyed out because they are not editable. Holding `SHIFT` is meant to enter an edit mode where both of them can be easily edited without mistakenly moving the playhead. This is why when `SHIFT` is held both of these things become lighter - to symbolize that they are no longer disabled. Holding `SHIFT` + `CTRL` should still visually act like we're in edit mode.

jmaen: Oh okay, that makes sense!

aarthificial (approved):

aarthificial:
Awesome, thanks for the PR!

PR #917, opened 2024-01-11
feat(ui): add number input dragging

### LINKED ISSUE #799: Dragging MetaFieldView Input
**Description**
The current input field only supports typing the value. For numerical fields like number fields and vector fields, it'll be more intuitive to support dragging behavior.

Peek 2023-08-23 17-05

**Proposed solution**
Left-click and drag to change the value. Single left-click to edit the value manually.

Peek 2023-08-23 17-13

**Considered alternatives**
Blender solution to the dragging numerical field: https://docs.blender.org/manual/en/latest/interface/controls/buttons/fields.html

jmaen:
Add a new `NumberInput` component that supports dragging to edit numbers. When hovering over the input, the cursor is a left-right arrow to indicate the dragging option. For the dragging behavior, the `PointerLock` API is used.

In order to edit the value manually via keyboard input, users can click the component without dragging. Dragging is then disabled and the cursor switches to a regular text cursor, allowing text selection. Up and down arrow keys are also supported through the regular `<input type="number"/>` element.

The new `NumberInput` also supports `min`, `max` and `step` properties for input via the arrow keys and via dragging.

Closes #799

https://github.com/motion-canvas/motion-canvas/assets/47495425/6677bb5c-9b73-4819-aa33-4f499ab4b337

### DIFF
--- a/packages/ui/src/components/controls/NumberInput.tsx
+++ b/packages/ui/src/components/controls/NumberInput.tsx
@@ -47,13 +47,16 @@ export function NumberInput({
     [min, max, step],
   );
 
-  useDocumentEvent('pointerlockchange', () => {
-    if (document.pointerLockElement === inputRef.current) {
-      document.addEventListener('pointermove', handleDrag);
-    } else {
-      document.removeEventListener('pointermove', handleDrag);
-    }
-  });
+  useDocumentEvent(
+    'pointerlockchange',
+    useCallback(() => {
+      if (document.pointerLockElement === inputRef.current) {
+        document.addEventListener('pointermove', handleDrag);
+      } else {
+        document.removeEventListener('pointermove', handleDrag);
+      }
+    }, [handleDrag]),
+  );
 
   return (
     <input
--- a/packages/ui/src/components/controls/NumberInput.tsx
+++ b/packages/ui/src/components/controls/NumberInput.tsx
@@ -1,0 +1,126 @@
+import clsx from 'clsx';
+import type {JSX} from 'preact';
+import {useCallback, useEffect, useRef, useState} from 'preact/hooks';
+import {useDocumentEvent} from '../../hooks';
+import {MouseButton, clamp} from '../../utils';
+import styles from './Controls.module.scss';
+
+type NumberInputProps = Omit<
+  JSX.HTMLAttributes<HTMLInputElement>,
+  'min' | 'max' | 'step' | 'value' | 'onChange' | 'onChangeCapture'
+> & {
+  min?: number;
+  max?: number;
+  step?: number;
+  value: number;
+  onChange?: (value: number) => void;
+  onChangeCapture?: (value: number) => void;
+};
+
+export function NumberInput({
+  min = -Infinity,
+  max = Infinity,
+  step = 1,
+  value = 0,
+  onChange,
+  onChangeCapture,
+  ...props
+}: NumberInputProps) {
+  const handleChange = onChangeCapture ? onChange : undefined;
+  const handleChangeCapture = onChangeCapture ?? onChange;
+
+  const inputRef = useRef<HTMLInputElement>();
+  const [currentValue, setCurrentValue] = useState(value);
+  const [active, setActive] = useState(false);
+  const [startX, setStartX] = useState(0);
+
+  useEffect(() => {
+    setCurrentValue(value);
+  }, [value]);
+
+  const handleDrag = useCallback(
+    (event: MouseEvent) => {
+      setCurrentValue(value => {
+        // ignore very large jumps in movement, as they often result from
+        // mouse acceleration issues
+        if (Math.abs(event.movementX) > 100) {
+          return value;
+        }
+
+        const newValue = clamp(min, max, value + event.movementX * step);
+        handleChange?.(newValue);
+        return newValue;
+      });
+    },
+    [min, max, step],
+  );
+
+  useDocumentEvent('pointerlockchange', () => {
+    if (document.pointerLockElement === inputRef.current) {
+      document.addEventListener('pointermove', handleDrag);
+    } else {
+      document.removeEventListener('pointermove', handleDrag);
+    }
+  });
+
+  return (
+    <input
+      type={'number'}
+      ref={inputRef}
+      min={min}
+      max={max}
+      step={step}
+      value={currentValue}
+      onChangeCapture={event => {
+        const value = parseInt((event.target as HTMLInputElement).value);
+        handleChangeCapture?.(value);
+      }}
+      onChange={event => {
+        const value = parseInt((event.target as HTMLInputElement).value);
+        console.log(value);
+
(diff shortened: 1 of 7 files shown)


packages/ui/src/components/controls/NumberInput.tsx
```diff
@@ -0,0 +1,126 @@
...
+  onChange,
+  onChangeCapture,
+  ...props
+}: NumberInputProps) {
+  const handleChange = onChangeCapture ? onChange : undefined;
+  const handleChangeCapture = onChangeCapture ?? onChange;
+
+  const inputRef = useRef<HTMLInputElement>();
+  const [currentValue, setCurrentValue] = useState(value);
+  const [active, setActive] = useState(false);
```
aarthificial, line 34: This state is not necessary since the value is only ever retrieved in event handles.
So whether the field has focus can be checked with just `document.activeElement === inputRef.value`.
Similarly, the styles can be changed using the `:focus` pseudo-class.

packages/ui/src/components/controls/NumberInput.tsx
```diff
@@ -0,0 +1,126 @@
...
+  min = -Infinity,
+  max = Infinity,
+  step = 1,
+  value = 0,
+  onChange,
+  onChangeCapture,
+  ...props
+}: NumberInputProps) {
+  const handleChange = onChangeCapture ? onChange : undefined;
+  const handleChangeCapture = onChangeCapture ?? onChange;
```
aarthificial, line 30: I don't think this is necessary here.
The `Input` component corrects the discrepancy between Preact's `onChange` and the browser's `onchange` by using `onChangeCapture` and since it's a generic component I left the possibility to not use the correction.
This is a pretty concrete component that has its own custom `onChange` callback so I think we can just dispatch it during `onChangeCapture` as that's how it's intended to work.

packages/ui/src/components/controls/NumberInput.tsx
```diff
@@ -0,0 +1,126 @@
...
+    [min, max, step],
+  );
+
+  useDocumentEvent('pointerlockchange', () => {
+    if (document.pointerLockElement === inputRef.current) {
+      document.addEventListener('pointermove', handleDrag);
+    } else {
+      document.removeEventListener('pointermove', handleDrag);
+    }
+  });
```
aarthificial, line 64: This captures `handleDrag` and won't update it even when `min`, `max`, or `step` change.

packages/ui/src/components/controls/NumberInput.tsx
```diff
@@ -0,0 +1,126 @@
...
+      max={max}
+      step={step}
+      value={currentValue}
+      onChangeCapture={event => {
+        const value = parseInt((event.target as HTMLInputElement).value);
+        handleChangeCapture?.(value);
+      }}
+      onChange={event => {
+        const value = parseInt((event.target as HTMLInputElement).value);
+        console.log(value);
```
aarthificial, line 80: 👀

packages/ui/src/components/controls/NumberInputSelect.tsx
```diff
@@ -0,0 +1,34 @@
...
+  options,
+  value,
+  onChange,
+  ...props
+}: NumberInputSelectProps) {
+  return (
+    <div className={styles.inputSelect}>
+      <NumberInput
+        value={value}
+        onChange={value => onChange(value)}
```
aarthificial, line 28: Couldn't it just be this?
```suggestion
        onChange={onChange}
```

packages/ui/src/components/controls/NumberInput.tsx
```diff
@@ -0,0 +1,126 @@
...
+  return (
+    <input
+      type={'number'}
+      ref={inputRef}
+      min={min}
+      max={max}
+      step={step}
+      value={currentValue}
+      onChangeCapture={event => {
+        const value = parseInt((event.target as HTMLInputElement).value);
```
aarthificial, line 75: In the component you use `inputRef.current` and `event.target` in somewhat randomly, I think we should stick to one.
```suggestion
        const value = parseInt(inputRef.current.value);
```

aarthificial (changes requested):
I think we should add the following handler to the input as well:
```ts
onKeyDown={event => {
  if (event.key === 'Enter') {
    inputRef.current.blur();
  }
  if (event.key === 'Escape') {
    inputRef.current.value = currentValue.toString();
    inputRef.current.blur();
  }
}}
```

packages/ui/src/components/controls/NumberInput.tsx
jmaen: So you mean we get rid of `onChangeCapture` entirely and use `onChange` instead? Would the `onChange` callback still be called for every step or, like `onChangeCapture` before, only when the pointer lock is released?

packages/ui/src/components/controls/NumberInput.tsx
aarthificial: No, never mind. I thought the second argument was wrapped in `useCallback` since that's how `useDocumentEvent` is usually used but you pass a raw function so the whole `pointerlockchange` will be resubscribed on each re-render. Is this something you do on purpose? Because I've noticed you removed `useCallback` from the timeline document event as well.

packages/ui/src/components/controls/NumberInput.tsx
aarthificial: The public interface should have an `onChange` property that would be called during `onChangeCapture` on the native input element.

packages/ui/src/components/controls/NumberInput.tsx
jmaen: In that case I did it on purpose for exactly that reason but if you have a more elegant solution please let me know! To be honest I don't exactly remember what I was thinking when working on the timeline, I'd hope there was a reason for that as well.. :)

aarthificial: I'm not sure about elegant but usually you'd use `useCallback` to memoize the function and pass that to the hook:
```suggestion
  useDocumentEvent(
    'pointerlockchange',
    useCallback(() => {
      if (document.pointerLockElement === inputRef.current) {
        document.addEventListener('pointermove', handleDrag);
      } else {
        document.removeEventListener('pointermove', handleDrag);
      }
    }, [handleDrag]),
  );
```
This will only re-register the handler when `handleDrag` or its dependencies change.
I'm planning on improving `useDocumentEvent` so that it doesn't rely on external memoization but right now this would be the way to go.

jmaen: Alright, sounds good!

jmaen:
I think I adressed all your requested changes. Apparently the pipeline failed to build, but the error seems to have occured in `core` which is not something I worked on.

aarthificial (approved):

aarthificial:
Awesome, thanks!

PR #925, opened 2024-01-27
fix: improper cloning of custom fields

shmishtopher:
Fixes a bug where `MetaObjectField` passes custom fields into the constructor after calling `clone`, causing the editor to crash due to missing `.get()` methods on the custom fields.

### DIFF
--- a/packages/core/src/meta/ObjectMetaField.ts
+++ b/packages/core/src/meta/ObjectMetaField.ts
@@ -72,11 +72,20 @@ class ObjectMetaFieldInternal<
   }
 
   public override serialize(): ValueOf<T> {
-    return this.transform('serialize');
+    return {
+      ...this.transform('serialize'),
+      ...this.customFields,
+    };
   }
 
   public override clone(): this {
-    return new (<any>this.constructor)(this.name, this.transform('clone'));
+    const cloned = new (<any>this.constructor)(
+      this.name,
+      this.transform('clone'),
+    );
+    cloned.customFields = structuredClone(this.customFields);
+
+    return cloned;
   }
 
   protected handleChange = () => {
@@ -91,10 +100,7 @@ class ObjectMetaFieldInternal<
       Array.from(this.fields, ([name, field]) => [name, field[fn]()]),
     ) as TransformationOf<T, TKey>;
 
-    return {
-      ...transformed,
-      ...this.customFields,
-    };
+    return transformed;
   }
 }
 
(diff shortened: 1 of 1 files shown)


aarthificial (changes requested):
I believe `this.transform('get')` should also include `customFields`, just like how `serialize` does.

shmishtopher:
I'm not sure how much sense it makes to include custom fields in the transform itself, given that any calls to `get` on custom fields would fail anyway.  I added an override to the `ObjectMetaField::get` method in the latest commit as a workaround.  Alternatively, we can return to your original implementation of `ObjectMetaField::transform` and add special handling to the `ObjectMetaField` constructor to fix the erroneous calls to `get`.

aarthificial:
Apologies for the confusion. I meant that the place where `this.transform('get')` is invoked (in the `handleChange` callback) should include `customFields`. This is where the value is calculated and so corresponding calls to `get` would return it too:
```ts
protected handleChange = () => {
  if (this.ignoreChange) return;
  this.value.current = {
    ...this.transform('get'),
    ...this.customFields,
  };
};
```

packages/core/src/meta/ObjectMetaField.ts
```diff
@@ -72,16 +72,28 @@ class ObjectMetaFieldInternal<
...
+    };
   }
 
   public override clone(): this {
-    return new (<any>this.constructor)(this.name, this.transform('clone'));
+    const cloned = new (<any>this.constructor)(
+      this.name,
+      this.transform('clone'),
+    );
+    cloned.customFields = structuredClone(this.customFields);
```
aarthificial, line 86: After testing it, it doesn't seem to work as expected.
The initial value of the meta field is set in the constructor, which doesn't have access to `customFields` so the value ends up not including them.
Here's a simple reproduction:
```ts
const field = new ObjectMetaField('example', {
  bool: new BoolMetaField('bool', false),
});
field.set({
  bool: true,
  custom: 'custom',
});

const clone = field.clone();

// This fails:
console.assert(clone.get().custom === 'custom');
```
I think `customFields` should be passed as an optional argument to the constructor instead of being assigned after the object has been created.

aarthificial (changes requested):

packages/core/src/meta/ObjectMetaField.ts
shmishtopher: I modified the constructor originally, but it required more extensive modifications to the call sites to split the custom fields from the built-ins when reading from the scene's meta file.  Instead of assigning to the object's internal `customFields` directly, we can call the `set` method.  This calls `handleChange` properly and only changes a single line. Give the latest commit a try, it works in all my cases.

aarthificial (approved):

aarthificial:
Awesome, thanks for the fix!

PR #948, opened 2024-02-09
feat(2d): code bounding box helpers

aarthificial:


packages/2d/src/lib/components/Code.ts
```diff
@@ -368,7 +376,123 @@ export class Code extends Shape {
    * @param pattern - Either a string or a regular expression to match.
    */
   public findLastRange(pattern: string | RegExp): CodeRange {
-    return findAllCodeRanges(this.parsed(), pattern).at(-1) ?? [[0, 0], [0, 0]];
+    return (
+      findAllCodeRanges(this.parsed(), pattern).at(-1) ?? [
```
hhenrichsen, line 380: Is there optimization we should worry about here? Normally `findLast` or `findFirst` are used to avoid a full string iteration if possible, although I'm not sure it's worth it here.

packages/2d/src/lib/components/Code.ts
```diff
@@ -368,7 +376,123 @@ export class Code extends Shape {
...
+
+  /**
+   * Return the bounding box of the given point (character) in the code.
+   *
+   * @remarks
+   * The returned bound box is in local space of the `Code` node.
+   *
+   * @param point - The point to get the bounding box for.
+   */
+  public getPointBbox(point: CodePoint): BBox {
```
hhenrichsen, line 395: Very excited for this.

packages/2d/src/lib/components/Code.ts
aarthificial: I don't know how to do that while supporting regex tbh.
The function is mainly for convenience but I'm always open to optimization propositions.

PR #960, opened 2024-02-15
Sample motion blur

zachHixson:
A rough implementation of sample based motion blur. Still contains a lot of hacks needed to get data from point A to point B which will need to be cleaned up, as well as a massive bug causing the motion blur samples to "overshoot," the next frame.


ItsSunnyMonster:
Is there anything stopping this pull request from being merged? I would really love this feature.

PR #968, opened 2024-02-28
934 slanted fonts fix

### LINKED ISSUE #934: Slanted fonts can result in characters at the start or end being cut off when opacity is lower than 1 or a shadow is placed on a Txt node
**Describe the bug**
When using fonts which have slanted characters (such as Montserrat Italic, which is what is pictured below), changing the opacity to be less than 1 or adding a shadow can cause characters to become incorrectly clipped.

**Example**
image

**What the font looks like without opacity**
image

**To Reproduce**
Prerequisite: set up the project to import a font with italics included. For example, placing the following

```css
@import url('https://fonts.googleapis.com/css2?family=Montserrat:ital,wght@1,300&family=Noto+Sans:ital,wght@1,300&family=Roboto:ital,wght@1,300&display=swap');
```
into an imported CSS file will import Montserrat, which is what is used in the following problematic snippet

```ts
  // When the opacity is 1, the text does not incorrectly clip the ends of characters
  const testText = createRef<Txt>();
  view.add(<Txt ref={testText} fontSize={256} fontFamily={"Montserrat"}>Lorem ipsumy</Txt>);
  yield* waitFor(1);
  // Now, as the opacity is reduced, the text incorrectly clips the ends of characters
  yield* testText().opacity(0.5, 0.5);
});
```

**Expected behavior**
The characters at the end are expected to not be clipped

**Console errors**
N/A

**Package versions:**
*Click on the version in the bottom right corner to copy the versions*

- core: 3.14.0
- two: 3.14.0
- ui: 3.14.0
- vitePlugin: 3.14.0

**Additional context**
Possibly related to #370

CactusPuppy:
This PR attempts to fix #934 by expanding the `BBox` returned by `getCacheBBox` for `TxtLeaf` by half of the element's font size horizontally.

### DIFF
--- a/packages/2d/src/lib/components/TxtLeaf.ts
+++ b/packages/2d/src/lib/components/TxtLeaf.ts
@@ -139,7 +139,7 @@ export class TxtLeaf extends Shape {
       -size.height / 2,
       bbox.width,
       bbox.height,
-    ).expand(lineWidth * miterLimitCoefficient);
+    ).expand([0, this.fontSize() * 0.5 * this.absoluteScale().x]).expand(lineWidth * miterLimitCoefficient);
   }
 
   protected override applyFlex() {
--- a/packages/2d/src/lib/components/TxtLeaf.ts
+++ b/packages/2d/src/lib/components/TxtLeaf.ts
@@ -139,7 +139,9 @@ export class TxtLeaf extends Shape {
       -size.height / 2,
       bbox.width,
       bbox.height,
-    ).expand([0, this.fontSize() * 0.5 * this.absoluteScale().x]).expand(lineWidth * miterLimitCoefficient);
+    )
+      .expand([0, this.fontSize() * 0.5])
+      .expand(lineWidth * miterLimitCoefficient);
   }
 
   protected override applyFlex() {


packages/2d/src/lib/components/TxtLeaf.ts
```diff
@@ -139,7 +139,7 @@ export class TxtLeaf extends Shape {
       -size.height / 2,
       bbox.width,
       bbox.height,
-    ).expand(lineWidth * miterLimitCoefficient);
+    ).expand([0, this.fontSize() * 0.5 * this.absoluteScale().x]).expand(lineWidth * miterLimitCoefficient);
```
aarthificial, line 142: To quote the docs of `getCacheBBox`:
> The returned bounding box should be in local space.

Is there a reason to multiply the font size by `absoluteScale` here?

aarthificial, line 142: Looks like something broke in our prettier configuration. This line is longer than 80 characters and should be choped down:
```ts
    return new BBox(-size.width / 2, -size.height / 2, bbox.width, bbox.height)
      .expand([0, this.fontSize() * 0.5 * this.absoluteScale().x])
      .expand(lineWidth * miterLimitCoefficient);
```
Normally prettier would be run on commit, I'll see what's up with that.

aarthificial (changes requested):

packages/2d/src/lib/components/TxtLeaf.ts
CactusPuppy: There is not, I just didn't notice that it was in local space. Will fix.

CactusPuppy: > Looks like something broke in our prettier configuration. This line is longer than 80 characters and should be choped down:
> 
> ```ts
>     return new BBox(-size.width / 2, -size.height / 2, bbox.width, bbox.height)
>       .expand([0, this.fontSize() * 0.5 * this.absoluteScale().x])
>       .expand(lineWidth * miterLimitCoefficient);
> ```
> 
> Normally prettier would be run on commit, I'll see what's up with that.

As it turns out, since TypeScript files are linted by ESLint and not Prettier. There's nothing wrong with the prettier configuration, it's just that the ESLint configuration doesn't have a rule restricting the maximum line length for TypeScript files.

I tried adding the following line to the `.eslintrc.json`, and upon rerunning `npx lint-staged` with a long line staged, an error was correctly thrown.

```diff
   "rules": {
     "require-yield": "off",
     "grouped-accessor-pairs": ["error", "getBeforeSet"],
     "eqeqeq": ["error", "always", {"null": "ignore"}],
     "curly": ["error", "multi-line"],
+    "max-len": ["error", {"code": 80}],
     "@typescript-eslint/explicit-member-accessibility": "error",
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/ban-ts-comment": "off",
     "@typescript-eslint/no-non-null-assertion": "off",
```

image

aarthificial (approved):

PR #969, opened 2024-02-29
Restrict line length of TypeScript files to 80 characters

CactusPuppy:
As discovered while reviewing #968 (specifically [this comment](https://github.com/motion-canvas/motion-canvas/pull/968#discussion_r1506780315)), although JavaScript's Prettier linting was configured to prevent staging of lines which exceed 80 characters in length, TypeScript's corresponding ESLint configuration was not set up the same way. This PR fixes this discrepancy.

Of note: although ESLint currently has a native [`max-len` rule](https://eslint.org/docs/latest/rules/max-len), the docs state that the rule is deprecated and should be migrated to the `@stylistic` rule used here. Rationale for why the stylistic rules were split off from the core ESLint project is [available here](https://eslint.style/guide/why).

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -16,12 +16,13 @@
     "sourceType": "module"
   },
   "root": true,
-  "plugins": ["@typescript-eslint", "eslint-plugin-tsdoc"],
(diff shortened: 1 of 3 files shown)


aarthificial:
Closing this in favor of #971

PR #978, opened 2024-03-11
feat(ui): support for presentations using logitech R400

Niikelion:
Logitech R400 presenter uses keys incompatible with motion canvas(`PageUp, PageDown, Escape/F5, .`). To make motion canvas compatible with it(and probably other presenters) I suggest adding adequate input handling to presentation mode.

### DIFF
--- a/packages/ui/src/components/playback/PlaybackControls.tsx
+++ b/packages/ui/src/components/playback/PlaybackControls.tsx
@@ -30,10 +30,12 @@ export function PlaybackControls() {
           return;
         }
         switch (event.key) {
+          case '.': // "blank" button on logitech R400 since play button alternates between ESC and F5
           case ' ':
             event.preventDefault();
             player.togglePlayback();
             break;
+          case 'PageUp':
           case 'ArrowLeft':
             event.preventDefault();
             if (event.shiftKey) {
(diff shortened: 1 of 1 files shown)


aarthificial (changes requested):
The controls for presentation mode are handled in a different file:
https://github.com/motion-canvas/motion-canvas/blob/6c9bbfe8e6a45a60975beada15256be93b5d65b3/packages/ui/src/components/presentation/PresentationControls.tsx#L34-L59

aarthificial:
As a side note, plugins can be used to easily define custom shortcuts:
```ts
import {makePlugin, makeProject} from '@motion-canvas/core';

import example from './scenes/example?scene';

export default makeProject({
  scenes: [example],
  plugins: [
    makePlugin({
      name: 'shortcuts',
      presenter(presenter) {
        document.addEventListener('keydown', event => {
          switch (event.key) {
            case 'PageUp':
              presenter.requestPreviousSlide();
              break;
          }
        });
      },
    }),
  ],
});
```
But I think adding `PageDown`, `PageUp`, etc. is still a good idea since that's the scheme power-point-like software use.

Niikelion:
My bad. I will fix the pr and remove `.` handling since in PowerPoint it does different thing which could confuse users.

PR #1002, opened 2024-03-20
feat(2d): simplify highlighters

aarthificial:


packages/docs/src/components/Fiddle/SharedPlayer.ts
```diff
@@ -120,11 +119,14 @@ export async function borrowPlayer(
       ValueDispatcher,
       DefaultPlugin,
     } = await import(/* webpackIgnore: true */ '@motion-canvas/core');
-    const {makeScene2D, LezerHighlighter} = await import(
+    const {makeScene2D, Code, LezerHighlighter} = await import(
       /* webpackIgnore: true */ '@motion-canvas/2d'
     );
-    LezerHighlighter.registerParser(javascript);
-    LezerHighlighter.registerParser(rust, 'rs');
+    Code.defaultHighlighter = new LezerHighlighter(
```
hhenrichsen, line 125: I still think we should include this somewhere in the docs, too.

packages/docs/docs/components/code/index.mdx
```diff
@@ -724,61 +724,105 @@ function hello() {
...
+import {parser} from '@lezer/rust';
 
-  view.add(
-    <>
-      <Code
-        // this node uses the default parser (javascript)
-        offsetX={-1}
-        x={-400}
-        code={`
+const Rust = new LezerHighlighter(parser);
```
hhenrichsen, line 763: Maybe name this `RustHighlighter` instead to avoid confusion with the `Rust` node?

hhenrichsen (approved):
I think this simplifies things which is good. I was previously configuring the highlighters in the project file which _worked_ but wasn't very clean.

packages/docs/src/components/Fiddle/SharedPlayer.ts
aarthificial: Done

PR #1016, opened 2024-03-29
fix(core): reset playback state on presentation abort

gtutunaru:
Hi,

I discovered a bug in the presentation mode. When entering the presentation mode and playing it until the last slide (so getting into the `finished` state), when closing the presentation and opening it again, it was not possible to start playing the slides, even though the play button is active (clicking on it had no effect).

After going through the code, I discovered that on `abort` the finished state is never reset.

For the fix, I was not sure if a separate method was needed to only update the `finished` flag, or an actual `playback.reset` was needed (which I did in the end).

### DIFF
--- a/packages/core/src/app/Presenter.ts
+++ b/packages/core/src/app/Presenter.ts
@@ -123,11 +123,10 @@ export class Presenter {
   /**
    * Abort the ongoing presentation process.
    */
-  public async abort() {
+  public abort() {
     if (this.state.current === PresenterState.Initial) return;
     this.abortController?.abort();
     this.state.current = PresenterState.Aborting;
-    await this.playback.reset();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/core/src/app/Presenter.ts
```diff
@@ -123,10 +123,11 @@ export class Presenter {
   /**
    * Abort the ongoing presentation process.
    */
-  public abort() {
+  public async abort() {
     if (this.state.current === PresenterState.Initial) return;
     this.abortController?.abort();
     this.state.current = PresenterState.Aborting;
+    await this.playback.reset();
```
aarthificial, line 130: I believe this is not necessary. Calling `present` already resets the playback here:
https://github.com/motion-canvas/motion-canvas/blob/5bd7c8e4f047408b4efa64f7e62db4257f409c57/packages/core/src/app/Presenter.ts#L176

I tested it, and resetting the `finished` flag that you added to the `PlaybackManager` seems to solve the issue just fine but let me know if I missed something.

gtutunaru: Oh yes, I missed that `reset` was already called. Removed it

aarthificial:
Thanks for the fix!

PR #1019, opened 2024-04-01
feat(2d): add camera node

ksassnowski:
This PR adds an orthographic camera node.

There a still a few kinks that I'd like to work out before merging this:

- The `hit` method selects the correct object, but the actual overlay in the editor is offset. I'm not sure why this happens.
- I would like most of the camera's methods to return a `SignalGenerator` in order to be chainable. I'm not sure what the best way to do this would be, however.

### DIFF
--- a/packages/2d/src/lib/components/Camera.ts
+++ b/packages/2d/src/lib/components/Camera.ts
@@ -37,8 +37,8 @@ export interface CameraProps extends NodeProps {
  *
  * @preview
  * ```tsx editor
- * import {all} from '@motion-canvas/core';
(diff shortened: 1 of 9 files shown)


aarthificial:
We could probably make the scene accept any `ComponentChildren` including a fragment (`<>`) by specifying a parser for the signal like this:
```ts
@parser((children: ComponentChildren) => {
  if (children instanceof Node) {
    return children;
  } else {
    return new Node({children});
  }
})
@signal()
public declare readonly scene: Signal<ComponentChildren, Node, this>;
```

Issue #1018, opened 2024-04-01
Preview-only Background Color


mxsdev:
**Description**
Would be nice to have a separate preview/viewport background color setting which doesn't affect the render. This is particularly useful when you want to export with a transparent background (to add perhaps a more complex background in post), but still want to get an idea of how it will look roughly while editing.

**Proposed solution**
Add an extra Video Setting under preview for "background". When empty, this will default to the General background, and when non-empty, this will become the background color for the viewport. This will not affect the render result at all - that will still use the background under "General".

**Considered alternatives**
Alternative is to change the background every time you need to render and then change it back for preview purposes. For scenarios where there are many scenes to render, this becomes tiresome quickly.

**Additional context**
Happy to make a PR for this, it seems fairly straight-forward :)

aarthificial:
I think it's a great idea. But we should stay consistent with other settings and make the background setting separate for the preview and for the rendering without any inheritance/default behavior (for now). 

Having a way to override the settings when previewing is an interesting idea but it should be implemented in a separate PR and affect other similar settings like FPS or Scale.

PR #1027, opened 2024-04-07
feat: separate background by preview/rendering

mxsdev:
Closes: #1018 

Some issues:

- Currently the default setting for background will apply to both preview and rendering video settings - we could also introduce two separate settings (maybe best in a separate PR?)
- **Backwards Compatibility**: Background settings from existing projects will be dropped and set to null for both preview and render
- **Forwards Compatibility**: Background settings from newer projects read on older versions will not be recognized, and this will become null here as well
- I did not write any tests for this change, but happy to (the code I changed didn't appear to be covered originally)

To address backwards compatibility issues, one approach is to introduce some sort of a "deprecated" flag for old fields to hide them in the preview, and then set the preview & render backgrounds accordingly on initial project load. Happy to do this in a separate PR before merging this.

To address forwards compatibility issues, we could sync the deprecated background field above to one of the preview or render settings (not sure which is better). Although I'm not sure how concerned we are with forwards compatibility.

Let me know your thoughts @aarthificial

### DIFF
--- a/packages/ui/src/components/viewport/StageView.tsx
+++ b/packages/ui/src/components/viewport/StageView.tsx
@@ -2,7 +2,7 @@ import type {Stage} from '@motion-canvas/core';
 import clsx from 'clsx';
 import {JSX} from 'preact';
 import {MutableRef, useLayoutEffect, useRef} from 'preact/hooks';
-import {useSharedSettings} from '../../hooks';
+import {usePreviewSettings} from '../../hooks';
 import styles from './Viewport.module.scss';
 
 export interface StageViewProps extends JSX.HTMLAttributes<HTMLDivElement> {
@@ -17,7 +17,7 @@ export function StageView({
   ...rest
 }: StageViewProps) {
   const localRef = useRef<HTMLDivElement>();
-  const {background} = useSharedSettings();
+  const {background} = usePreviewSettings();
   const ref = forwardRef ?? localRef;
 
   useLayoutEffect(() => {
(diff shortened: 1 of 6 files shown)


aarthificial:
> one approach is to introduce some sort of a "deprecated"/"hidden" flag for old fields to hide them in the preview

There's already a `disable` method that can be called when defining meta fields that will hide them in the inspector.
But I don't think this would be necessary, undeclared fields in meta files are preserved so the migration can be performed by overriding the `set` method of `ProjectMetadata`:
```ts
export class ProjectMetadata /* ... */ {
  // ...

  public override set(
    value: Partial<ValueOf<ReturnType<typeof createProjectMetadata>>>,
  ) {
    if (value.shared && 'background' in value.shared) {
      const background = value.shared.background as Color | null;
      delete value.shared.background;
      if (value.preview) {
        value.preview.background = background;
      }
      if (value.rendering) {
        value.rendering.background = background;
      }
    }
    super.set(value);
  }
}
```

We also don't support forward compatibility since that's practically impossible. There are lots of new features being added in each version so trying to open a project with an older editor will almost always fail.

packages/ui/src/components/viewport/StageView.tsx
```diff
@@ -17,7 +17,7 @@ export function StageView({
   ...rest
 }: StageViewProps) {
   const localRef = useRef<HTMLDivElement>();
-  const {background} = useSharedSettings();
+  const {background} = usePreviewSettings();
```
aarthificial, line 20: `StageView` is used to display the animation in all three places:
- `PresentationMode` - uses rendering settings
- `RenderingViewport` - uses rendering settings
- `PreviewStage` - uses preview settings

So the background should probably be passed here as a prop depending on the context in which the stage is used.

aarthificial (changes requested):

Issue #967, opened 2024-02-26
Mathematical operations for Vector2 signals


aarthificial:
**Description**
Modifying a vector by applying some operation (addition, subtraction, etc) is currently really verbose.
Imagine you want to animate the position of `node` by moving it by `[200, 400]` relative to its current position:
```ts
yield* node.position(node.position().add([200, 400]), 0.6);
```
In general, whenever we want to derive the new value of a signal using its current value things become verbose.

**Proposed solution**
Introduce a set of methods for Vector2 signals that reflect the operations of the `Vector2` class.
```ts
// before
yield* node.position(node.position().add([200, 400]), 0.6);
// after
yield* node.position.add([200, 400], 0.6);
```
These could be used to either tween or immediately modify the signal, just like you normally would:
```ts
// set
node.position.add([200, 400]);
// tween
yield* node.position.add([200, 400], 0.6);
```

On top of that, a new `edit` method should be introduced to derive the new value in whatever way necessary:
```ts
// this:
node.position.add([200, 400]);
// is the same as this:
node.position.edit(current => current.add([200, 400]));
```
While on itself still pretty verbose, this interface would allow you to create reusable functions for operations that are not natively supported.
Imagine you want to snap the vector to a `20x20` grid:
```ts
// define a reusable method:
function snap(grid: PossibleVector2) {
  return (current: Vector2) => current.div(grid).floored.mul(grid);
}

// use it:
node.position.edit(snap(20));
```

mancopp:
I am working on this one

PR #1030, opened 2024-04-13
Vector2 compound signal operations

mancopp:
This pull request is related to issue #967 

Implementations of mathematical operations in this pull request concern Vector2s based on CompoundSignal exclusively - meaning node fields with decorator `@vector2Signal()`.
Example code to test the functionality:

```ts

  view.add(
    <Rect
      ref={rect}
      size={320}
      radius={80}
      x={18}
      y={121}
      smoothCorners
      fill={'#f3303f'}
    />,
  );

  // before
  yield* rect().position(node.position().add([200, 400]), 0.6);
  // after
  yield* rect().position.add([200, 400], 0.6);

  // set
  rect().position.add([200, 400]);
  // tween
  yield* rect().position.add([200, 400], 0.6);
  
  // this:
  rect().position.add([200, 400]);
  // is the same as this:
  rect().position.edit(current => current.add([200, 400]));

  // define a reusable method:
  function snap(grid: PossibleVector2) {
    return (current: Vector2) => current.div(grid).floored.mul(grid);
  }
  
  // use it:
  rect().position.edit(snap(20));
```


packages/core/src/signals/Vector2CompoundSignalContext.ts
```diff
@@ -0,0 +1,247 @@
...
+    timingFunction?: TimingFunction,
+    interpolationFunction?: InterpolationFunction<Vector2>,
+  ): Vector2 | TOwner | SignalGenerator<PossibleVector2, Vector2>;
+  mod(
+    possibleVector: PossibleVector2,
+    duration?: number,
+    timingFunction?: TimingFunction,
+    interpolationFunction?: InterpolationFunction<Vector2>,
+  ): Vector2 | TOwner | SignalGenerator<PossibleVector2, Vector2>;
+};
```
aarthificial, line 65: You cannot declare these methods using single signatures as this doesn't contain all the necessary information. Depending on the passed parameters, a different type will be returned. This requires method overloading.
For example, this code won't work right now:
```ts
rect().position.add(20).rotation(45);
```
We know that calling `add` with only the value will return the owner of the signal, allowing us to chain other operations. But the type system doesn't know this.

Instead, you can take inspiration from the `CodeSignal`

First, declare interfaces for your operations using overloading:
```ts
export interface Vector2Edit<TOwner> {
  (callback: (current: Vector2) => SignalValue<PossibleVector2>): TOwner;
  (
    callback: (current: Vector2) => SignalValue<PossibleVector2>,
    duration: number,
    timingFunction?: TimingFunction,
    interpolationFunction?: InterpolationFunction<Vector2>,
  ): SignalGenerator<PossibleVector2, Vector2>;
}

export interface Vector2Operation<TOwner> {
  (possibleVector: PossibleVector2): TOwner;
  (
    value: PossibleVector2,
    duration: number,
    timingFunction?: TimingFunction,
    interpolationFunction?: InterpolationFunction<Vector2>,
  ): SignalGenerator<PossibleVector2, Vector2>;
}
```
Then define the interface with all the helper methods:
```ts
export interface Vector2SignalHelpers<TOwner> {
  edit: Vector2Edit<TOwner>;
  mul: Vector2Operation<TOwner>;
  div: Vector2Operation<TOwner>;
  add: Vector2Operation<TOwner>;
  sub: Vector2Operation<TOwner>;
  dot: Vector2Operation<TOwner>;
  cross: Vector2Operation<TOwner>;
  mod: Vector2Operation<TOwner>;
}
```
Use it to both, declare the `Vector2Signal` type:
```ts
export type Vector2Signal<
  TOwner = void,
  TContext = Vector2CompoundSignalContext<TOwner>,
> = CompoundSignal<PossibleVector2, Vector2, 'x' | 'y', TOwner, TContext> &
  Vector2SignalHelpers<TOwner>;
```
And as an interface implemented by the context:
```ts
export class Vector2CompoundSignalContext<TOwner = void>
  extends CompoundSignalContext<PossibleVector2, Vector2, 'x' | 'y', TOwner>
  implements Vector2SignalHelpers<TOwner>
{
```
(This will make sure the context implements all the necessary methods)

Methods on the context should now also feature overloading:
```ts
class ... {
  public edit(
    callback: (current: Vector2) => SignalValue<PossibleVector2>,
  ): TOwner;
  public edit(
    callback: (current: Vector2) => SignalValue<PossibleVector2>,
    duration: number,
    timingFunction?: TimingFunction,
    interpolationFunction?: InterpolationFunction<Vector2>,
  ): SignalGenerator<PossibleVector2, Vector2>;
  public edit(
    callback: (current: Vector2) => SignalValue<PossibleVector2>,
    duration?: number,
    timingFunction?: TimingFunction,
    interpolationFunction?: InterpolationFunction<Vector2>,
  ): SignalGenerator<PossibleVector2, Vector2> | TOwner {
    const newValue = callback(this.get());
    return this.invoke(
      newValue,
      duration,
      timingFunction,
      interpolationFunction,
    ) as SignalGenerator<PossibleVector2, Vector2> | TOwner;
  }
}
```

packages/core/src/signals/Vector2CompoundSignalContext.ts
```diff
@@ -0,0 +1,247 @@
...
+    return this.invokable;
+  }
+
+  public edit(
+    callback: (current: Vector2) => SignalValue<PossibleVector2>,
+    duration?: number,
+    timingFunction?: TimingFunction,
+    interpolationFunction?: InterpolationFunction<Vector2>,
+  ) {
+    const newValue = callback(this.invokable());
```
aarthificial, line 127: Could just use `get`:
```suggestion
    const newValue = callback(this.get());
```

packages/core/src/signals/Vector2CompoundSignalContext.ts
```diff
@@ -0,0 +1,247 @@
...
+  ): Vector2 | TOwner | SignalGenerator<PossibleVector2, Vector2>;
+  mod(
+    possibleVector: PossibleVector2,
+    duration?: number,
+    timingFunction?: TimingFunction,
+    interpolationFunction?: InterpolationFunction<Vector2>,
+  ): Vector2 | TOwner | SignalGenerator<PossibleVector2, Vector2>;
+};
+
+export class Vector2CompoundSignalContext<
```
aarthificial, line 67: The class should be called just `Vector2SignalContext`
```suggestion
export class Vector2SignalContext<
```

aarthificial (changes requested):

aarthificial (approved):

aarthificial:
Thanks for the PR!

PR #1072, opened 2024-06-30
docs: add hint about `?scene` import path to quickstart guide

DATADEER:
When I tried to render my project without the `?scene` import postfix I got the following errors.
I'd like to propose to add a hint about the import postfix to the quickstart guide. 

```ts import {makeProject} from '@motion-canvas/core';
import podcast from "./scenes/podcast/podcast"; // without ?scene

export default makeProject({
  scenes: [ podcast],
});
```

## Errors
CleanShot 2024-06-30 at 15 17 11

CleanShot 2024-06-30 at 15 17 39

### DIFF
--- a/packages/docs/docs/getting-started/quickstart.mdx
+++ b/packages/docs/docs/getting-started/quickstart.mdx
@@ -122,8 +122,8 @@ export default makeProject({
 ```
 
 When creating a project, we need to provide it with an array of scenes to
-display. In this case, we use only one scene imported from
-`src/scenes/example.tsx`.
+display. In this case, we use only one scene imported from `src/scenes/example.tsx?scene`. 
+(the `?scene` is required for TypeScript to recognize the file as a scene module.)
 
 A scene is a set of elements displayed on the screen and an animation that
 governs them. The most basic scene looks as follows:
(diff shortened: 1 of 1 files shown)


packages/docs/docs/getting-started/quickstart.mdx
```diff
@@ -122,8 +122,8 @@ export default makeProject({
 ```
 
 When creating a project, we need to provide it with an array of scenes to
-display. In this case, we use only one scene imported from
-`src/scenes/example.tsx`.
+display. In this case, we use only one scene imported from `src/scenes/example.tsx?scene`. 
+(the `?scene` is required for TypeScript to recognize the file as a scene module.)
```
aarthificial, line 126: I think we can expand it a bit more and put it in its own paragraph without the parenthesis since forgetting about the `?scene` is a common mistake:
```suggestion
Note the `?scene` at the end - it's required to transform the imported module into a proper scene. Among other things, it makes it possible to dynamically refresh the preview whenever you modify the scene. The editor will not work without it.
```

aarthificial (changes requested):

aarthificial (approved):

Issue #1079, opened 2024-07-31
Getting chroma_js error on Quickstart


aman-kalsi:
**Describe the bug**
On following the instructions provided in the [Quickstart guide](https://motioncanvas.io/docs/quickstart), I am seeing the following error:

![Screenshot 2024-07-31 000418](https://github.com/user-attachments/assets/1d0e03f6-234e-4854-8f79-3ef13b519656)

I am on node v21.4.0

**To Reproduce**
npm init @motion-canvas@latest
cd project-path
npm install
npm start

**Expected behavior**
Application should start as expected on fresh install.

**Package versions:**
- core: [3.16.0]
- ui: [3.16.0]

**Additional context**
It appears to be a problem with [chroma_js v2.6](https://www.npmjs.com/package/chroma-js?activeTab=versions), which was released a few hours ago. Running projects on chroma_js 2.4.2 works fine.

Elucidation:
For anyone still dealing with this, you can temporarily downgrade the offending package with 

`npm install chroma-js@2.4.2`  in your animation directory

goforu:
> Casual user here: For anyone dealing with this, you can also temporarily downgrade the offending package with
> 
> `npm install chroma-js@2.4.2` in your animation directory

it works, thanks bro!

aarthificial:
Will be fixed by #1086 but let's leave this issue open until https://github.com/gka/chroma.js/issues/346 is fixed

PR #1080, opened 2024-07-31
fix(chroma_js version #1079)

Neptunium931:
set version of chroma-js to 2.4.2


goforu:
waiting for being merged

aarthificial:
Thanks for the PR!
I'm closing it in favor of #1086 because I had to add the missing `package.lock.json` changes and also lock the `@types/chroma-js` version (so it doesn't break once 2.6 is released)

Issue #1076, opened 2024-07-26
Project Selection - Project File Name Must be Different


Urchmaney:
**Description**
While setting up multiple project in my repository, I was under the false assumption that if the path to the project is different it would work.
But found out after sometime that the file name is what's being used in motion canvas and not the path.
In my <b>vite.config.ts</b> file I had:
```
motionCanvas({ project: [
    "./src/blazer/project.ts",
    "./src/linked/project.ts"
  ]}
)
```
so I needed to change it to
```
motionCanvas({ project: [
    "./src/blazer/blazer_project.ts",
    "./src/linked/linked_project.ts"
  ]}
)
```
for it to work.

**How can it be improved?**
- Specify explicitly in the doc (under Getting Started/Configuration) that project file names must be different
- (OR) Allow the usage of path. If there is collision with the project file name, folder name can be used. (I can create a PR for this if it is considered)

aarthificial:
Yes, using the entire file path as the url would be the desired solution as mentioned here: https://github.com/motion-canvas/motion-canvas/issues/297#issuecomment-1746172903
Feel free to open a PR if you're still interested

aarthificial:
Closed by #1090

PR #1090, opened 2024-08-14
fix(vite-plugin, ui): project file path on selection

Urchmaney:
fix issue of conflict with projects with same file name. Projects with same project file name in different folders, only runs the last project computed.

Fixes issue #1076

### DIFF
--- a/packages/vite-plugin/src/partials/editor.ts
+++ b/packages/vite-plugin/src/partials/editor.ts
@@ -20,15 +20,24 @@ export function editorPlugin({editor, projects}: EditorPluginConfig): Plugin {
 
   const lookup = new Map<string, ProjectData>();
   for (const project of projects) {
-    lookup.set(project.name, project);
+    lookup.set(project.filePath, project);
   }
+
+  let query: string;
 
   return {
     name: 'motion-canvas:editor',
 
-    async load(id) {
-      const [, query] = id.split('?');
+    async resolveId(source) {
+      if (source.startsWith(resolvedEditorId)) {
+        const [id, projectQuery] = source.split('?');
+        query = projectQuery;
+        return id;
+      }
+      return null;
+    },
 
+    async load(id) {
       if (id.startsWith(resolvedEditorId)) {
         if (projects.length === 1) {
           /* language=typescript */
--- a/packages/vite-plugin/src/partials/editor.ts
+++ b/packages/vite-plugin/src/partials/editor.ts
@@ -20,30 +20,21 @@ export function editorPlugin({editor, projects}: EditorPluginConfig): Plugin {
 
   const lookup = new Map<string, ProjectData>();
   for (const project of projects) {
-    lookup.set(project.filePath, project);
+    lookup.set(project.url, project);
   }
-
-  let query: string;
 
   return {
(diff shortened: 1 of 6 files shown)


packages/vite-plugin/src/partials/editor.ts
```diff
@@ -20,15 +20,24 @@ export function editorPlugin({editor, projects}: EditorPluginConfig): Plugin {
...
-    async load(id) {
-      const [, query] = id.split('?');
+    async resolveId(source) {
+      if (source.startsWith(resolvedEditorId)) {
+        const [id, projectQuery] = source.split('?');
+        query = projectQuery;
+        return id;
+      }
+      return null;
+    },
```
aarthificial, line 38: This is a bit too hacky. There's no guarantee the corresponding `load` will be called right after `resolveId`. The same server may be used to serve two projects at the same time, for example.
Also, this global `query` variable will break the project selections screen:
1. Visit localhost:9000
2. Choose one of the projects
3. Go back to localhost:9000
4. Instead of the selection screen, the previously chosen project is immediately loaded

The reason you were getting an error was because the name was not escaped when generating the `virtual:editor?project=[name]` url. All you need to do is add `encodeURIComponent` in the server a few lines below:
```ts
if (name && lookup.has(name)) {
  res.setHeader('Content-Type', 'text/html');
  res.end(
    createHtml(
      // right here:
      `/@id/__x00__virtual:editor?project=${encodeURIComponent(
        name,
      )}`,
    ),
  );
  return;
}
```
and it will work just fine without `resolveId` and the global `query` variable.

packages/vite-plugin/src/plugins.ts
```diff
@@ -12,6 +12,10 @@ export interface ProjectData {
    * The file name containing the project.
    */
   fileName: string;
+  /**
+   * The path to the file containing the project
+   */
+  filePath: string;
```
aarthificial, line 18: This is too confusing I think. `url` is storing the actual path and `filePath` is the url used to access the project (with no file extension etc) Maybe we could swap the two?

packages/vite-plugin/src/utils.ts
```diff
@@ -20,7 +20,13 @@ export function getProjects(project: string | string[]): ProjectData[] {
     const {name, dir} = path.posix.parse(url);
     const metaFile = `${name}.meta`;
     const metaData = getMeta(path.join(dir, metaFile));
-    const data = {name: metaData?.name ?? name, fileName: name, url};
+    const filePath = [...dir.split('/').slice(2), name].join('/');
```
aarthificial, line 23: We don't know if people use `src` to store their projects, someone could use a `videoA/project.ts`, `videoB/project.ts` structure with no `src`.
I think we should use the entire path here and not strip the `src`:
```suggestion
    const filePath = path.posix.join(dir, name);
```

aarthificial (changes requested):

packages/vite-plugin/src/partials/editor.ts
Urchmaney: Ah. Ok got it.

aarthificial (approved):
Awesome, thanks so much!

PR #1111, opened 2024-09-22
docs: proper link syntax to Spline from Path

AbsoluteWisp:
(Documentation) The "Path" page tried to link to "Spline", but it used invalid syntax `[display][link]` instead of `[display](link)`, resulting in a result that's broken visually and, more importantly, doesn't actually link. This simple commit fixes that.

### DIFF
--- a/packages/docs/docs/components/path.mdx
+++ b/packages/docs/docs/components/path.mdx
@@ -37,7 +37,7 @@ The `Path` node allows us to draw and animate SVG paths.
 `Path` nodes are great for displaying precalculated SVG paths, but the format
 can be confusing. If you want to have more intuitive control of the path, and
 control each point within Motion Canvas, we recommend checking out the
-[`Spline`](spline) component.
+[`Spline`](/api/2d/components/Spline) component.
 
 :::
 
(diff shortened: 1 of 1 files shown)


packages/docs/docs/components/path.mdx
```diff
@@ -37,7 +37,7 @@ The `Path` node allows us to draw and animate SVG paths.
 `Path` nodes are great for displaying precalculated SVG paths, but the format
 can be confusing. If you want to have more intuitive control of the path, and
 control each point within Motion Canvas, we recommend checking out the
-[`Spline`][spline] component.
+[`Spline`](spline) component.
```
aarthificial, line 40: The syntax used previously was actually valid, it's called [reference-style](https://www.markdownguide.org/basic-syntax/#reference-style-links). However, the said reference was missing:
```md
[`Spline`][spline]

// This was missing from the bottom of the page
[spline]: /api/2d/components/Spline
```
Your new link still won't work, you'd need to specify the full link:
```md
[`Spline`](/api/2d/components/Spline)
```
or revert to the previous syntax and add the missing reference at the bottom.

aarthificial (changes requested):
Good catch! There's still an issue with the link though

packages/docs/docs/components/path.mdx
AbsoluteWisp: Apparently my new link still accidentally works since the two files (Path and Spline) just so happen to be in the same folder. I'll fix this in a second by just using the reference style properly

AbsoluteWisp:
Alright, this should be good now I think. Changed the link to be a full path-style one, since another place on the page used this style and I figured it's best to just be consistent.

aarthificial (approved):

aarthificial:
Awesome, thanks for the fix!

PR #1128, opened 2024-10-19
fix(core): always show both scenes when seeking mid-transition

r04423:
This PR updates seeking-related logic in `PlaybackManager` to ensure both scenes are visible when seek position is in the middle of transition. Previously, it could occur that only the `currentScene` will be rendered, especially visible when seeking backwards.

Before:

https://github.com/user-attachments/assets/e83742a1-d3e5-4e3a-a848-f7d0e1b16d5b

After:

https://github.com/user-attachments/assets/9c834528-a812-46d2-891e-305d38f0ad56


aarthificial:
Thanks for the fix!

PR #1135, opened 2024-11-25
fix: lower the globalAlpha of the previous context as you increase th…

CleverFlare:
…e current context

Previously, the fade transition caused the previous scene to suddenly vanish. This little tweak will cause the previous scene to gradually fade out while the current scene fades in.


aarthificial:
Can we make this into a parameter? The previous behavior was meant for fullscreen scenes where the next one would overlap the first one completely. This change uses a different type of alpha blending and makes the result look different so we should keep the default behavior the same as it used to be and let people opt in to the new behavior.

CleverFlare:
Sure thing. I'll make the necessary changes immediately.

Issue #1035, opened 2024-04-27
Allow matching substrings instead of patterns


pihedron:
# Description
Currently, the `Code.findAllRanges` method does not match an exact substring in a `Code` object. Instead, it unnecessarily creates a `RegExp` object. If the `findAllCodeRanges` function can take a `RegExp` object as an argument, why should it have to convert a `string` to `RegExp`?

# Solution
I could not find any existing function that matched a substring in the docs, so I am suggesting a quick change to remove some code from the `findAllCodeRanges` function in `packages/2d/src/lib/code/CodeRange.ts`.

```ts
// remove this block of code
if (typeof pattern === 'string') {
  pattern = new RegExp(pattern, 'g');
}
```

This means we can use `code().findFirstRange('n * (n + 1) / 2')` instead of typing out so many backslashes like this: `code().findFirstRange('n \\* \\(n \\+ 1\\) \\/ 2')`. It is probably better practice to use a `RegExp` constructor instead of implicitly converting a `string` into a `RegExp`.

aarthificial:
The current implementation of `findAllCodeRanges` uses `String.matchAll` to perform the search. This method expects a `RegExp` and passing a string to it will implicitly convert it to a `RegExp` just like we do. Removing the code you've mentioned will have no effect on the current behavior (aside from making the type checker angry). 

> It is probably better practice to use a RegExp constructor instead of implicitly converting a string into a RegExp.

The reason we convert strings to `RegExp`s is to easily reuse the current algorithm. The point was never to let people define actual `RegExp`s as strings and nowhere in the documentation do we suggest people do that.

Supporting strings without the need for escape sequences is a good idea but it will require more work than removing the mentioned lines of code. The current algorithm will have to be adapted to work with strings.

pihedron:
Yes, you are right. I forgot that `String.matchAll` converted `string` to `RegExp` anyway. Should I make a PR with a new function for finding a code range?

pihedron:
I have to admit motion-canvas is amazing though.

aarthificial:
@pihedron Apologies, I forgot to answer, feel free to take care of this issue if you're still interested, otherwise I can take care of it later this week.

PR #1153, opened 2025-02-03
feat(2d): modifies  to match exact strings when passed a plain string

henriqueinonhe:
DRAFT

An attempt at #1035.

### DIFF
--- a/packages/2d/src/lib/code/CodeRange.ts
+++ b/packages/2d/src/lib/code/CodeRange.ts
@@ -1,3 +1,5 @@
+import escapeRegExpString from 'escape-string-regexp';
+
 export type CodePoint = [number, number];
 
 function isCodePoint(value: unknown): value is CodePoint {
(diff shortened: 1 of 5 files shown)


aarthificial:
Thanks for the PR!
The logic looks good, not sure about the -3000 lines of code in `package.lock.json` though. Do you know why it may be?

packages/2d/src/lib/code/CodeRange.ts
```diff
@@ -168,7 +170,7 @@ export function inverseCodeRange(ranges: CodeRange[]): CodeRange[] {
  * Find all code ranges that match the given pattern.
  *
  * @param code - The code to search in.
- * @param pattern - Either a string or a regular expression to search for.
+ * @param pattern - Either a string or a regular expression to search for. When a string is passed, it looks for **exact matches**. When a RegExp object is passed, it will match against it.
```
aarthificial, line 173: Nit but can you apply this formatting to stay in the character limit:
```suggestion
 * @param pattern - Either a string or a regular expression to search for.
 *                  When a string is passed, it looks for **exact matches**.
 *                  When a RegExp object is passed, it will match against it.
```
Sadly I don't have a formatter configured for that yet.

aarthificial:
Thanks for the fix!
Like I mentioned in #1035, this was never meant to be a feature and so it is not documented anywhere. Our [examples in doc](https://motioncanvas.io/docs/code#code-ranges) use strings for exact string matching and regexes for regex.

PR #1156, opened 2025-02-06
Fix audio not playing when playhead is after audio's duration (not accounting for offset)

### LINKED ISSUE #1155: Audio stops playing in editor when playhead is after the audio's duration (disregarding offset)
**Describe the bug**

When previewing a scene in the editor/player, if the playhead (current time) is **after** the audio's total **duration** it stops playing the audio **even when the audio offset would make it such that it should be playing**.

**To Reproduce**

To reproduce:

1. Import some audio into a sample project
2. Create a scene whose duration exceeds the audio's duration
3. Offset the audio so that it "overshoots" the scene duration
4. Play

This is the code that's causing the issue

https://github.com/motion-canvas/motion-canvas/blob/13c9de85280cc1b893a178b9d6eecd8d639fd7bb/packages/core/src/media/AudioManager.ts#L60-L62

**Expected behavior**

Player should take the audio offset into consideration when deciding whether to play or pause the audio.

**Console errors**
If applicable, add a screenshot of the errors displayed in the console of your browser (F12) or your terminal.

**Package versions:**

- core: 3.17.2
- two: 3.17.2
- ui: 3.17.2

**Additional context**

When preparing to play, the player checks whether the audio should be playing by following this logic:

https://github.com/motion-canvas/motion-canvas/blob/13c9de85280cc1b893a178b9d6eecd8d639fd7bb/packages/core/src/app/Player.ts#L412-L413

One of the things it checks is whether the current time is in the audio's range.

When we look at `AudioManager.isInRange`'s implementation:

https://github.com/motion-canvas/motion-canvas/blob/main/packages/core/src/media/AudioManager.ts#L60-L62

we notice that it doesn't account for the audio offset, which causes the issue.

henriqueinonhe:
Fixes #1155 

When preparing to play, the player checks whether the audio should be playing by following this logic:

https://github.com/motion-canvas/motion-canvas/blob/13c9de85280cc1b893a178b9d6eecd8d639fd7bb/packages/core/src/app/Player.ts#L412-L413

One of the things it checks is whether the current time is in the audio's range.

When we look at `AudioManager.isInRange`'s implementation:

https://github.com/motion-canvas/motion-canvas/blob/main/packages/core/src/media/AudioManager.ts#L60-L62

we notice that it doesn't account for the audio offset, which causes the issue.

### DIFF
--- a/packages/core/src/media/AudioManager.ts
+++ b/packages/core/src/media/AudioManager.ts
@@ -58,7 +58,10 @@ export class AudioManager {
   }
 
   public isInRange(time: number) {
-    return time >= this.offset && time < this.audioElement.duration;
+    const audioRangeStart = this.offset;
+    const audioRangeEnd = this.audioElement.duration + this.offset;
+
+    return audioRangeStart <= time && time <= audioRangeEnd;
   }
 
   public toRelativeTime(time: number) {


aarthificial (approved):
Thanks so much! Great catch

PR #1158, opened 2025-02-09
feat: sound waveforms

aarthificial:
Renders the waveforms of individual sounds on top of the timeline:
image

### DIFF
--- a/packages/core/src/app/Player.ts
+++ b/packages/core/src/app/Player.ts
@@ -3,7 +3,7 @@ import {
   EventDispatcher,
   ValueDispatcher,
 } from '../events';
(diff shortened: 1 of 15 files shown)


aarthificial:
@Astavie could you take a look at this if you find the time?

I tweaked your implementation a bit, namely:
- I updated `toRelativeTime`/`toAbsoluteTime` so that the trim is applied before the playback rate. This is already the case in your ffmpeg logic and it makes the preview playback consistent.
- I updated how the pool is prepared to use the duration of the audio clip instead of the duration of a frame. This makes sure that audio is played even if you rewind back to the middle of it.

Astavie:
Haven't tested it yet, but looking at just the code the tweaks look alright

PR #1159, opened 2025-02-09
Validate resolution field (coerce values in UI so it remains within bounds)

### LINKED ISSUE #1149: Resoluition fields don't have validation
**Describe the bug**
Resolution fields inside video settings tab have no validation for zero and negative numbers.
I accidentally cleared the vertical field and was getting the following message in the console:
```
Failed to execute 'drawImage' on 'CanvasRenderingContext2D': The image argument is a canvas element with a width or height of 0.
```
I cleared the entire project before I accidentally noticed that the resolution field was empty.

**To Reproduce**
Enter video settings -> general, clear or enter zero or enter negative number inside any resolution field.

**Expected behavior**
On invalid number enter revert resolution field change or set minimum allowed resolution size

**Package versions:**
- core: 3.17.2
- two: 3.17.2
- ui: 3.17.2
- vitePlugin: 3.17.2

henriqueinonhe:
Fixes #1149 .

## Demo

motion-canvas

## Overview

Implements coercion for the resolution fields so that inputted values always remain within bounds (i.e. non-zero positive value).

I implemented a strategy which I think makes the most sense UX-wise, which is:

- When the inputted value is outside bounds, it gets clamped
- When the inputted value is not a number, it remains unaltered

To accomplish that, I:

- Implemented the possibility to pass bounds (lower and upper) to `Vector2MetaField` which constitutes its invariants and are checked as its constructor and mutating methods preconditions
- Made it so that `Vector2MetaFieldView` automatically reads these bounds and coerces inputted values

### DIFF
--- a/packages/core/src/app/ProjectMetadata.ts
+++ b/packages/core/src/app/ProjectMetadata.ts
@@ -18,7 +18,9 @@ function createProjectMetadata(project: Project) {
     shared: new ObjectMetaField('General', {
       background: new ColorMetaField('background', null),
       range: new RangeMetaField('range', [0, Infinity]),
-      size: new Vector2MetaField('resolution', new Vector2(1920, 1080)),
+      size: new Vector2MetaField('resolution', new Vector2(1920, 1080), {
+        min: [1, 1],
+      }),
       audioOffset: new NumberMetaField('audio offset', 0),
     }),
     preview: new ObjectMetaField('Preview', {
--- a/packages/core/src/app/ProjectMetadata.ts
+++ b/packages/core/src/app/ProjectMetadata.ts
@@ -18,9 +18,9 @@ function createProjectMetadata(project: Project) {
     shared: new ObjectMetaField('General', {
       background: new ColorMetaField('background', null),
       range: new RangeMetaField('range', [0, Infinity]),
-      size: new Vector2MetaField('resolution', new Vector2(1920, 1080), {
-        min: [1, 1],
-      }),
+      size: new Vector2MetaField('resolution', new Vector2(1920, 1080)).min([
+        1, 1,
+      ]),
       audioOffset: new NumberMetaField('audio offset', 0),
     }),
     preview: new ObjectMetaField('Preview', {
--- a/packages/core/src/meta/Vector2MetaField.ts
+++ b/packages/core/src/meta/Vector2MetaField.ts
@@ -1,33 +1,6 @@
+import {clamp} from '../tweening';
 import {PossibleVector2, Vector2} from '../types';
 import {MetaField} from './MetaField';
-
-export type Vector2MetaFieldConstructorParameters = {
-  /**
-   * A 2D tuple representing the minimum value
-   * (inclusive) for each element of the the vector.
-   *
-   * Defaults to [-Infinity, -Infinity].
-   *
-   * This constitutes an invariant of the vector
-   * and attempts to change its value will be
-   * validated against it, which will trigger
-   * an exception if validation fails.
-   */
-  min?: [number, number];
-
-  /**
-   * A 2D tuple representing the maximum value
-   * (inclusive) for each element of the the vector.
-   *
-   * Defaults to [Infinity, Infinity].
-   *
-   * This constitutes an invariant of the vector
-   * and attempts to change its value will be
-   * validated against it, which will trigger
-   * an exception if validation fails.
-   */
-  max?: [number, number];
-};
 
 /**
  * Represents a two-dimensional vector stored in a meta file.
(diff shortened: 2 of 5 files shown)


packages/core/src/app/ProjectMetadata.ts
```diff
@@ -18,7 +18,9 @@ function createProjectMetadata(project: Project) {
     shared: new ObjectMetaField('General', {
       background: new ColorMetaField('background', null),
       range: new RangeMetaField('range', [0, Infinity]),
-      size: new Vector2MetaField('resolution', new Vector2(1920, 1080)),
+      size: new Vector2MetaField('resolution', new Vector2(1920, 1080), {
+        min: [1, 1],
+      }),
```
aarthificial, line 23: Can we use a chaining API to make it consistent with how `MetaField.describe()` etc work?
```suggestion
      size: new Vector2MetaField('resolution', new Vector2(1920, 1080)).min([
        1, 1,
      ]),
```

packages/core/src/meta/Vector2MetaField.ts
```diff
@@ -1,17 +1,111 @@
...
+        `Vector2MetaField invariant violation!`,
+        `You're trying to construct/set this field to a value that is out of bounds!`,
+        `Value -> [${x}, ${y}]`,
+        `Min -> [${minX}, ${minY}]`,
+        `Max -> [${maxX}, ${maxY}]`,
+      ].join('\n');
+
+      throw new Error(message);
+    }
+  }
```
aarthificial, line 110: MetaFields should perform parsing (clamping, etc) instead of validation (throwing errors), similarly to how `RangeMetaField` works, for example. 

Often times, the same meta field is edited in multiple different places (a field in the settings, a gizmo in the editor, parsing JSON data from a meta file, etc.) if we put the responsibility of parsing on whoever invokes `set` we'll end up with the same logic duplicated in multiple places.

packages/ui/src/components/meta/MetaFieldView.tsx
```diff
@@ -39,6 +39,8 @@ export function MetaFieldView({field}: MetaFieldViewProps) {
   const Field: FiledView = TYPE_MAP.get(field.type) ?? UnknownMetaFieldView;
   const disabled = useSubscribableValue(field.onDisabled);
 
+  console.log(field);
+
```
aarthificial, line 43:
```suggestion
```

aarthificial (changes requested):
Thanks for the PR!
I left a few comments

packages/core/src/meta/Vector2MetaField.ts
```diff
@@ -7,11 +8,136 @@ import {MetaField} from './MetaField';
...
+
+  public override clone(): this {
+    const vectorClone = new Vector2(this.get().x, this.get().y);
+
+    // I'm not entirely sure what's going on here type-wise
+    // that it complains about the return type
+    return new Vector2MetaField(this.name, vectorClone)
+      .min(this.getMin())
+      .max(this.getMax()) as MetaField<PossibleVector2, Vector2> as any;
+  }
```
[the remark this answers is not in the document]
aarthificial: I'd write it as follows:
```ts
public override clone(): this {
  const clone = super.clone();
  return clone.min(this.getMin()).max(this.getMax()) as this;
}
```

The cast is necessary either way, however the problem with your implementation is that it will always instantiate the `Vector2MetaField` class, even when this method is called on an object of a class that **inherits** `Vector2MetaField`.

Using `super.clone()` makes sure that the correct class is instantiated despite inheritance.

packages/core/src/meta/Vector2MetaField.ts
```diff
@@ -7,11 +8,136 @@ import {MetaField} from './MetaField';
...
+        this.lowerBounds,
+        this.upperBounds,
+        this.get(),
+      ),
+    );
+
+    return this;
+  }
+
+  public max(upperBounds: [number, number]): Vector2MetaField {
```
aarthificial, line 52: Could we use `PossibleVector2` for bounds?

packages/core/src/meta/Vector2MetaField.ts
```diff
@@ -7,11 +8,136 @@ import {MetaField} from './MetaField';
...
+    const [maxX, maxY] = max;
+    const [originalX, originalY] = [vector.x, vector.y];
+
+    const coercedX = clamp(minX, maxX, originalX);
+    const coercedY = clamp(minY, maxY, originalY);
+
+    vector.x = coercedX;
+    vector.y = coercedY;
+
+    return vector;
```
aarthificial, line 137: How about we implement a `clamp` method on the `Vector2` class so that this becomes:
```tsx
return new Vector2(value).clamp(min, max);
```
?

packages/core/src/meta/Vector2MetaField.ts
```diff
@@ -7,11 +8,136 @@ import {MetaField} from './MetaField';
...
+  }
+
+  private lowerBounds: [number, number];
+
+  private upperBounds: [number, number];
```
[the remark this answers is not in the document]
aarthificial: Looking at the other classes again, I think we should rename the `min` and `max` functions to `setMin` and `setMax` since all other function follow a verb/noun pattern: `describe() / description`, `disable() / disabled`, `setRange / min,max`, etc.

These could then become `min` and `max`

Issue #1175, opened 2025-03-21
Project Selection Icon not Working For Nested Project


Urchmaney:
**Describe the bug**
In the user interface, the project selection icon does not link to the project selection page.

**To Reproduce**
Steps to reproduce the behavior or a code snippet containing the scene that causes the problem:
- Create a project that is nested. an example from my project `./src/africa_path/africa_path_project`
- Start motion canvas and navigate to the project. this will go to `http://localhost:9000/src/africa_path/africa_path_project`
- Now click the `project selection` icon. it navigates to `http://localhost:9000/src/` which is not a page.

**Expected behavior**
Expected to go to the `/` page where we can select a project.

**Console errors**

![Image](https://github.com/user-attachments/assets/e7f30200-ea5e-4873-8b5f-ca0d64f69465)

**Package versions:**
*Click on the version in the bottom right corner to copy the versions*

- ui: "^3.17.1"

PR #1176, opened 2025-03-21
fix(ui): project selection wrong path

Urchmaney:
#1175


BretHudson:
This was bothering me recently too, thanks for the PR!

Issue #1204, opened 2025-08-04
ffmpeg: Export does not contain all frames and frame inconsistencies.


prathje:
**Describe the bug**
The FFmpeg export finished early and did not contain all frames.

To me, it looks like the final 'end' message to close the ImageStream is transmitted before other frames (perhaps due to its smaller size).

Further, looking at the FFmpegExporterClient.ts implementation, I also fear that frames could be delivered out of order. This especially holds when the EXPORT_FRAME_LIMIT is reached: as long as the limit is not reached, frames should be transmitted in order, as the transmission is typically over the local network and the frames are the same size (though there is no guarantee). However, once a frame waits because of the EXPORT_RETRY_DELAY, a later frame might just be sent before an earlier frame, making out-of-order transmissions much more likely. Please correct me if I am wrong.

**To Reproduce**
Export any scene at a high resolution.
```ts
yield* waitFor(1.0);
```

**Expected behavior**
The ffmpeg export contains all frames.

**Console errors**
```
@motion-canvas_ui.js?v=609fba11:54971 An error occurred while exporting the video. 
{remarks: 'Method: end<br>Server error: Input stream error: stream.push() after EOF'}
```

**Package versions:**

- core: 3.18.0-alpha.0
- two: 3.18.0-alpha.0
- ui: 3.18.0-alpha.0
- vitePlugin: 3.18.0-alpha.0

**Additional context**
The issue disappears with lower resolution exports, so I assume it is related to the size of the frames compared to the size of the end request.

PR #1205, opened 2025-08-04
fix(ffmpeg): chain promises to keep frame order

prathje:
I encountered strange issues with the FFmpeg export: the export did not contain all frames. See #1204

To me, it looks like the final 'end' message to close the ImageStream is transmitted before other frames (perhaps due to its smaller size).
Further, looking at the FFmpegExporterClient.ts implementation, I also fear that frames could be delivered out of order. This especially holds when the EXPORT_FRAME_LIMIT is reached: as long as the limit is not reached, frames should be transmitted in order, as the transmission is typically over the local network and the frames are the same size (though there is no guarantee). However, once a frame waits because of the EXPORT_RETRY_DELAY, a later frame might just be sent before an earlier frame, making out-of-order transmissions much more likely. Please correct me if I am wrong.

To tackle these issues, I propose to (1) let the end message wait for all frames to be transmitted and (2) let frames wait for the previous frames to be transmitted.

I achieved both by keeping track of the promise for the current frame. A new frame replaces this promise while awaiting the previous promise. Finally, the end message is only sent after the current frame (which will then be the final frame) has been transmitted. Hence, frames are guaranteed to be in order, but at the same time, frames are no longer transmitted concurrently, resulting in a slowed-down export. A reliable concurrent export should include the frame indices, which the server can then use to reorder the frames.

### DIFF
--- a/packages/ffmpeg/client/FFmpegExporterClient.ts
+++ b/packages/ffmpeg/client/FFmpegExporterClient.ts
@@ -32,9 +32,6 @@ type FFmpegExporterOptions = ValueOf<
 
 type InvokeStrategy = 'ws' | 'octet-stream';
 
-const EXPORT_FRAME_LIMIT = 256;
-const EXPORT_RETRY_DELAY = 1000;
-
 /**
  * FFmpeg video exporter.
  *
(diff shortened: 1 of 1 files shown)


bufferhead-code:
just wanna say thank you for the PR, i also had problems with the export and your changes fixed it.

PR #1217, opened 2025-10-12
fix(2d): allow custom `HighlightStyle` with `all`-color. Rewrite cache to store color for every char

MaPePeR:
Fix the `LezerHighlighter`, so it also works when the `all`-color is set on the `HighlightStyle` (This is required to set the color for e.g. Python colons).
The tree-traversal seems to not be required to determine the bounds of token, because `highlightTree` already provides those in the `from` and `to` arguments to the `putStyle` function (at least I couldn't find or think of a reason, why the current code does this?).
The cache was replaced to use a `Uint8Array` to store a color-id for every character, which greatly simplifies color lookup in comparison to the `Map` that stores index ranges as strings.

<details><summary>Example Code</summary>

```js
import {HighlightStyle} from '@codemirror/language';
import {tags as t} from '@lezer/highlight';
import {parser} from '@lezer/python';
import {Code, LezerHighlighter, makeScene2D} from '@motion-canvas/2d';
import {waitFor} from '@motion-canvas/core';

  const style = HighlightStyle.define(
    [
      {tag: t.comment, color: 'blue'},
      {tag: t.keyword, color: 'yellow'},
      {tag: t.function(t.variableName), color: 'green'},
    ],
    {
      all: {color: 'gray'},
    },
  );

  const highlighter = new LezerHighlighter(parser, style);

  view.add(
    <Code
      highlighter={highlighter}
      code={`\
def f():
      print("Hello") 
      print("world") # this is a comment
`}
    />,
  );

  yield* waitFor(0.3);
});
```
</details>

Before:
grafik

After:
grafik


MaPePeR:
Breaks Tweening. As described in canvas-commons/canvas-commons#4

